PR #18, opened 2020-05-23
Add support for installing Pest into a Lumen application

michaeldyrynda:
This entails creating Laravel and Lumen-specific stubs, and ensuring that the appropriate stubs are copied as part of the pest install.

**The `stubs/Lumen/Pest.php` file needs to be updated before this can be merged**

As `uses(TestCase::class)` has no effect without the call to `->in()`, so tests won't actually work in a Lumen application because there is no namespace to load from.

Adding a call to `uses(TestCase::class)` in a test file itself *does* work, so it's possible to use Pest with Lumen, but will need the manual call in each test file at this stage.


owenvoke:
Can you not use `uses(TestCase::class)->in(__DIR__)`?

michaeldyrynda:
So you can!

nunomaduro:
Thanks @michaeldyrynda !

PR #29, opened 2020-05-25
Restore target for higher order method calls

### LINKED ISSUE #28: Idea: Higher Order Tests - handle null returns
When chaining multiple higher order methods it can happen, that one assertion does not return the target instance and simply returns null or nothing.

For example

```php
it('exposes the plugin class via composer.json')
    ->assertArrayHasKey('extra', $data)
    ->assertArrayHasKey('pest', $data['extra']);
```

does not work because `assertArrayHasKey` returns null. So the second assertion breaks.

The idea would be to keep the initial target and reuse it when a chain method does not return a value.

```php
    {
        $originalTarget = $target;

            $target = $message->call($target);

            if ($target === null) {
                $target = $originalTarget;
            }
        }
    }
```

What do you think?

fetzi:
if higher order method call does return null or nothing use the
original target for future calls

Resolves #28

### DIFF
--- a/src/Support/HigherOrderMessageCollection.php
+++ b/src/Support/HigherOrderMessageCollection.php
@@ -29,8 +29,14 @@ final class HigherOrderMessageCollection
      */
     public function chain(object $target): void
     {
+        $originalTarget = $target;
+
         foreach ($this->messages as $message) {
             $target = $message->call($target);
+
+            if ($target === null) {
+                $target = $originalTarget;
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/Support/HigherOrderMessageCollection.php
```diff
@@ -29,8 +29,14 @@ public function add(string $filename, int $line, string $methodName, array $argu
      */
     public function chain(object $target): void
     {
+        $originalTarget = $target;
+
         foreach ($this->messages as $message) {
             $target = $message->call($target);
+
+            if ($target === null) {
+                $target = $originalTarget;
```
nunomaduro, line 38: Maybe you can change this to use the last non null target?

fetzi: sure

nunomaduro:
@fetzi Today I am not sure about this. Because autocompletion wise, `assertTrue` returns void. Maybe in the future we can re-consider this.

Issue #49, opened 2020-05-30
--colors=never is not respected


maks-rafalko:
When I run

```bash
vendor/bin/pest --colors=never
```

Pest continues printing ascii chars that produces the following text:

```
  [37;41;1m FAIL [39;49;22m[39m PestExample\Test\PhpUnitSourceFileTest[39m
  [31;1m•[39;22m[39m [2mit can add two integers[22m[39m

  [2m---[22m

  [31;1m• PestExample\Test\PhpUnitSourceFileTest [39;22m> [31;1mit can add two integers[39;22m
  [41;1m Error [49;22m

[39;1m  Call to protected method PestExample\SourceFile::add() from context 'PestExample\Test\PhpUnitSourceFileTest'[39;22m

  at [32mtests/PhpUnitSourceFileTest.php[39m:[32m18[39m
     14▕     public function test_it_can_add_two_integers(): void
     15▕     {
     16▕         $source = new SourceFile();
     17▕ 
  ➜  18▕         $result = $source->add(3, 2);
     19▕ 
     20▕         $this->assertGreaterThan(0, $result);
     21▕     }
     22▕ }



  [37;1mTests:  [39;22m[31;1m1 failed[39;22m
  [37;1mTime:   [39;22m[39m0.03s[39m
```

For example, when I use `--colors=never` with PHPUnit, it correctly disables and prints just raw text:

```bash
PHPUnit 9.1.5 by Sebastian Bergmann and contributors.

.                                                                   1 / 1 (100%)

Time: 00:00.046, Memory: 6.00 MB
```

Related to #4 (In Infection PHP, we need a "raw" output without all those colours for terminal to parse the text);

Even without Infection, it's still the issue as some CI do not support such output as well

nunomaduro:
@maks-rafalko Maybe is better just change the printer to the default phpunit printer. what do you think?

maks-rafalko:
1. is it possible right now? can't find anything related
2. if it's possibl (or will be) - will I be able to use `--colors=never` with default PHPUnit printer?
3. I still think `colors=never` should be supported by Pest. Why? For example, if Pest is used for Jenkins, there is no support for ascii codes (from what I see in my daily job :) ). So I see `�[37;41;1m FAIL` insteaf of pretty raw `FAIL`

Jibbarth:
To launch Pest with default phpunit printer, I usually do : 

```bash
[...]

PR #51, opened 2020-05-31
Feature/support colors never

Jibbarth:
Related issue #49 

There is a still present escaped sequence ascii when running with `--colors=never`. I think it's because hardcoded here : https://github.com/nunomaduro/collision/blob/stable/src/Adapters/Phpunit/Style.php#L89


nunomaduro:
Thanks! Good job buddy!

PR #41, opened 2020-05-28
Some Windows fixes

Pierstoval:
Binary was using realpath, and targets were sanitized without taking the full path's drive name in account, so this was fixed.

I intend to wait for CI to make sure the test suite still works as expected.

I hope it'll fix #36 🤔

PR #61, opened 2020-06-02
Fixes autoloading, plugins and tests on Windows

dimitrioskarvounaris:
Based on https://github.com/pestphp/pest/pull/41 and all tests for `test:unit` pass under Windows and Linux.

It fixes issues, where otherwise some features wouldn't load or work at all under Windows.

### DIFF
--- a/src/PendingObjects/UsesCall.php
+++ b/src/PendingObjects/UsesCall.php
@@ -61,7 +61,7 @@ final class UsesCall
         $targets = array_map(function ($path): string {
             $startChar = DIRECTORY_SEPARATOR;
 
-            if ('\\' === DIRECTORY_SEPARATOR) {
+            if ('\\' === DIRECTORY_SEPARATOR || preg_match('~\A[A-Z]:(?![^/\\\\])~i', $path) > 0) {
                 $path = (string) preg_replace_callback('~^(?P<drive>[a-z]+:\\\)~i', function ($match): string {
                     return strtolower($match['drive']);
                 }, $path);
--- a/src/PendingObjects/UsesCall.php
+++ b/src/PendingObjects/UsesCall.php
@@ -83,7 +83,7 @@ final class UsesCall
                 throw new InvalidUsesPath($target);
             }
 
-            return realpath($target);
+            return (string) realpath($target);
         }, $targets);
     }
 
--- a/src/PendingObjects/UsesCall.php
+++ b/src/PendingObjects/UsesCall.php
@@ -59,7 +59,17 @@ final class UsesCall
     public function in(string ...$targets): void
     {
         $targets = array_map(function ($path): string {
-            return $path[0] === DIRECTORY_SEPARATOR
+            $startChar = DIRECTORY_SEPARATOR;
+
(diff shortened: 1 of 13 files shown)


nunomaduro:
Sounds a fantastic job. Do you know why tests are failing?

tests/Visual/SingleTestOrDirectory.php
```diff
@@ -3,63 +3,54 @@
...
 };
 
 test('allows to run a single test', function () use ($run) {
-    assertStringContainsString(<<<EOF
-   PASS  Tests\Fixtures\DirectoryWithTests\ExampleTest
-  ✓ it example
-
-  Tests:  1 passed
-EOF, $run('tests/Fixtures/DirectoryWithTests/ExampleTest.php'));
+    $output = $run('tests/Fixtures/DirectoryWithTests/ExampleTest.php');
```
nunomaduro, line 14: Can you revert this changes? I prefer in the way it was before.

If don't work on Windows, but create a file with the contents and do a `file_get_contents` on it.

dimitrioskarvounaris: Not sure how you mean with the file_get_contents()?
I were going to solve it differently first, by replacing the \r\n with \n, but then I understood, that there are cases, where anything that still contains linebreaks can break in more than one way and as such the line breaks should not be part of the comparison. Having a file, could still make it fail in some cases I think?

dimitrioskarvounaris:
> Do you know why tests are failing?

You mean, if I know, why the tests were failing? Yes, I debugged the code and figured the reason for each one. Some things are posted in Discord. 

Often - besides the linebreaks - it was that paths would not match, and as such, working on strings where part of path would be replaced or something added, would no longer happen or not happen correctly. Especially, when $rootPath was used. This would even lead to things as having the wrong output next to "PASS", traits not autoloading or plugins not even initializing with the "invalid path exception" when Plugin::uses() (as it sets `->in($rootPath)`) was used.

tests/Visual/SingleTestOrDirectory.php
dimitrioskarvounaris: I know, I like the other way it was also better, but this one works always.

nunomaduro: Can you revert all the changes in SingleTestOrDirectory.php ? Except the line 6.

src/PendingObjects/UsesCall.php
```diff
@@ -68,12 +78,12 @@ public function in(string ...$targets): void
         }, $targets);
 
         $this->targets = array_map(function ($target): string {
-            $realTarget = realpath($target);
-            if ($realTarget === false) {
-                throw new InvalidUsesPath($target);
+            $isValid = is_dir($target) || file_exists($target);
+            if (!$isValid) {
+                throw new InvalidUsesPath($target . "\n");
```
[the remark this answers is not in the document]
dimitrioskarvounaris: No idea. I think it wasn't from me... can be removed I guess

nunomaduro:
@dimitrioskarvounaris @Pierstoval Huge thanks! <3

Issue #45, opened 2020-05-29
Proposal: pest-plugin-init


nunomaduro:
After merging @fetzi's work on plugins. Would be cool to develop a `pest-plugin-init` that would work the following way: 

1. In applications without tests: `./vendor/bin/pest --init`
2. Boostraps the tests folder
3. Create a configuration file

owenvoke:
This would be nice. 👍 I'd be happy to take a look when the plugin PR is merged.

AlexMartinFR:
Nice!
Could we get rid of the dashes to provide a cleaner CLI experience `./vendor/bin/pest init` ?

Some more thought about CLI syntax here: [#10](https://github.com/pestphp/pest/issues/10#issuecomment-636305590)

nunomaduro:
Dashes are for options. So you are right

fetzi:
@owenvoke are you still on it?

owenvoke:
I was, but was struggling on a way to implement it with the current plugin API.

If it uses HandlesArguments, then it may be run after other plugins which is not desired as it basically just needs to be a separate command.

fetzi:
Yes, maybe this one shouldn't be a plugin and should be included in the core. @nunomaduro what do you think?

nunomaduro:
Should be a plugin that exists by default on the core. 😊

owenvoke:
Ok, I think we may need to modify the plugin API to allow this then. 👍

fetzi:
@owenvoke I think that `handleArguments` is the right place for that. It won't matter if another plugin gets executed beforehand.

owenvoke:
@fetzi, or afterwards? 🤔 Just seems a bit weird for `init` to also run tests. But if so, I'm happy to just do that.

nunomaduro:
@owenvoke Well, you can `exit(0)` when you are handling the arguments of the init right?

fetzi:
The watch plugin will also not continue to execute the tests directly. So I think `exit(0)` is good for the init plugin.

owenvoke:
Just to check, I assume that if that's at the end of the `handleArguments()` method, then the `addOutput()` wouldn't be called? Because it would just exit.

PR #65, opened 2020-06-03
Add basic container implementation

fetzi:
Relates to #45

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -85,6 +85,9 @@
             "providers": [
                 "Pest\\Laravel\\PestServiceProvider"
             ]
+        },
+        "branch-alias": {
+            "dev-master": "dev-feature/add-container"
         }
     }
 }
--- a/src/Support/Container.php
+++ b/src/Support/Container.php
@@ -1,0 +1,82 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pest\Support;
+
+use Pest\Exceptions\ShouldNotHappen;
+use ReflectionClass;
+use ReflectionParameter;
+
+/**
+ * @internal
+ */
+final class Container
+{
+    /**
+     * @var array<string, mixed>
+     */
+    private $instances = [];
+
+    /**
+     * Gets a dependency from the container.
+     *
+     * @return object
+     */
+    public function get(string $id)
+    {
+        if (array_key_exists($id, $this->instances)) {
+            return $this->instances[$id];
+        }
+
+        $this->instances[$id] = $this->build($id);
+
+        return $this->instances[$id];
+    }
+
+    /**
+     * Adds the given instance to the container.
+     *
+     * @param mixed $instance
+     */
+    public function add(string $id, $instance): void
+    {
+        $this->instances[$id] = $instance;
+    }
+
+    /**
+     * Tries to build the given instance.
+     */
+    private function build(string $id): object
+    {
+        /** @phpstan-ignore-next-line */
+        $reflectionClass = new ReflectionClass($id);
+
+        if ($reflectionClass->isInstantiable()) {
+            $constructor = $reflectionClass->getConstructor();
+
+            if ($constructor !== null) {
+                $params = array_map(
+                    function (ReflectionParameter $param) use ($id) {
+                        $candidate = null;
+
+                        if ($param->getType() !== null && $param->getType()->isBuiltin()) {
+                            $candidate = $param->getName();
+                        } elseif ($param->getClass() !== null) {
+                            $candidate = $param->getClass()->getName();
+                        } else {
+                            throw ShouldNotHappen::fromMessage(sprintf('The type of `$%s` in `%s` cannot be determined.', $id, $param->getName()));
+                        }
+
+                        return $this->get($candidate);
+                    },
+                    $constructor->getParameters()
+                );
+
+                return $reflectionClass->newInstanceArgs($params);
+            }
+        }
+
+        throw ShouldNotHappen::fromMessage(sprintf('A dependency with the name `%s` cannot be resolved.', $id));
+    }
+}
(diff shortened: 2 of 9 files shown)


fetzi:
@nunomaduro maybe you can have a look at the first draft of the container. I will then continue with the part inside pest-plugin

composer.json
```diff
@@ -22,6 +22,7 @@
         "pestphp/pest-plugin": "dev-master",
         "pestphp/pest-plugin-coverage": "dev-master",
         "phpunit/phpunit": "^9.1.4",
+        "psr/container": "^1.0",
```
nunomaduro, line 25: I am not even sure we should implement this interface. Realistically, people will **never** use this.

What do you think?

src/Support/Container.php
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * @return mixed
+     */
+    public function get($id)
+    {
+        if ($this->has($id)) {
+            return $this->instances[$id];
+        }
+
+        $this->instances[$id] = $this->build($id);
```
nunomaduro, line 42: No need for the `build` thing. If the instance do not exist let's throw a `ShouldNotHappen` in this cases.

src/Exceptions/ContainerTypeCannotBeDetermined.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Pest\Exceptions;
+
+use InvalidArgumentException;
+use Psr\Container\ContainerExceptionInterface;
+
+/**
+ * @internal
+ */
+final class ContainerTypeCannotBeDetermined extends InvalidArgumentException implements ContainerExceptionInterface
```
nunomaduro, line 13: Let's throw a ShouldNotHappen in this cases instead.

src/Exceptions/ContainerValueCannotBeResolved.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Pest\Exceptions;
+
+use InvalidArgumentException;
+use Psr\Container\NotFoundExceptionInterface;
+
+/**
+ * @internal
+ */
+final class ContainerValueCannotBeResolved extends InvalidArgumentException implements NotFoundExceptionInterface
```
nunomaduro, line 13: Let's throw a ShouldNotHappen in this cases instead.

tests/Unit/Support/Container.php
```diff
@@ -0,0 +1,76 @@
...
+
+it('adds an instance', function () {
+    $this->container->add(Container::class, new Container());
+    assertTrue($this->container->has(Container::class));
+})->group('container');
+
+it('gets an instance', function () {
+    $this->container->add(Container::class, $this->container);
+    assertSame($this->container, $this->container->get(Container::class));
+})->group('container');
```
nunomaduro, line 24: You can do `uses()->group('container')` on the being of the test file.

src/Support/Container.php
```diff
@@ -0,0 +1,94 @@
...
+{
+    /**
+     * @var array<string, mixed>
+     */
+    private $instances = [];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function has($id): bool
```
nunomaduro, line 26: No need for `has`.

src/Support/Container.php
```diff
@@ -0,0 +1,94 @@
...
+    {
+        return array_key_exists($id, $this->instances);
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @return mixed
+     */
+    public function get($id)
```
nunomaduro, line 36: `$id` is always a string.

src/Support/Container.php
```diff
@@ -0,0 +1,94 @@
...
+        return $this->instances[$id];
+    }
+
+    /**
+     * Adds the given instance to the container.
+     *
+     * @param string $id
+     * @param mixed  $instance
+     */
+    public function add($id, $instance): void
```
nunomaduro, line 53: lets type hint $id to string.

src/Support/Container.php
```diff
@@ -0,0 +1,94 @@
...
+    {
+        $this->instances[$id] = $instance;
+    }
+
+    /**
+     * Tries to build the given instance.
+     *
+     * @param mixed $id
+     */
+    private function build($id): object
```
nunomaduro, line 63: no need.

src/Support/Container.php
fetzi: I think we need the `build` method because we need it for the instance creation of the Plugins. Otherwise we need to push the container to the plugins via constructor.

owenvoke: Is pushing the container to plugins through the constructor an issue? 🤔 I feel like accessing `$this->container->get()` in a plugin class seems fine. Or have I misunderstood this? 🙂

nunomaduro: Injecting a container in classes is bad practice. Classes should declare their dependencies ( in your case the console output and the test suite ), and the container should resolve the dependencies while creating a new instance of that class.

owenvoke: Thanks for the explanation 👍

nunomaduro:
You can just rebase @fetzi ?

nunomaduro:
@fetzi Can you check why tests are not passing?

owenvoke:
Looks like it might be because it was merged rather than rebased? I think this [branch alias](https://github.com/pestphp/pest/pull/65/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R89) should be removed.

nunomaduro:
@fetzi Can you rebase again, and fix tests? I am sorry.

nunomaduro:
👏

PR #66, opened 2020-06-03
feat: adds pending higher order tests

nunomaduro:
Pending higher order tests is part of **Pest 0.2**. It basically allows to start a higher order test directly from the assertion. Here is an example using the Laravel plugin: https://github.com/pestphp/pest-plugin-laravel/pull/7.

```php
<?php

use Pest\Laravel\{get};

get('/')->assertSee('laravel');
```

Result:
```
   PASS  Tests\Http
  ✓ get '/' → assert see 'laravel'
```


src/Console/Command.php
```diff
@@ -88,6 +88,8 @@ protected function createRunner(): TestRunner
          */
         $this->arguments = AddsDefaults::to($this->arguments);
 
+        LoadStructure::in($this->testSuite->rootPath);
```
nunomaduro, line 91: This basically autoloads the helpers and pest file before the test suite itself.

PR #78, opened 2020-06-11
Optimise debug_backtrace calls

felixdorn:
With this PR, we prevent PHP from populating the backtrace with all the functions/methods arguments.

This PR can have little to BIG effects, depending on when `Backtrace::file()`  and `Backtrace::file()` are called, so chance are that at some point this can save a lot of memory, considering the fact that they are called *really* often, afaik.

I did some benchmarking and it saves between nothing and 6MB (with tests and logic) (using Blackfire). Ofc, it doesn't mean a lot because benchmarking will depends on a bunch of things.

Quoting [php.net](https://www.php.net/manual/en/function.debug-backtrace.php):
> DEBUG_BACKTRACE_IGNORE_ARGS : Whether or not to omit the "args" index, and thus all the function/method arguments, **to save memory**.

### DIFF
--- a/src/Support/Backtrace.php
+++ b/src/Support/Backtrace.php
@@ -25,7 +25,7 @@ final class Backtrace
     {
         $current = null;
 
-        foreach (debug_backtrace() as $trace) {
+        foreach (debug_backtrace(self::BACKTRACE_OPTIONS) as $trace) {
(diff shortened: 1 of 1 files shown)


nunomaduro:
Is this ready to be merged?

nunomaduro:
Thanks!

felixdorn:
Yes, it was ready! Thanks for this awesome framework!

PR #82, opened 2020-06-13
Fix instatiation of constructorless classes

fetzi:
This fix is needed for the pest-plugin fix.


nunomaduro:
@fetzi Tests are not passing buddy. can you take a look?

nunomaduro:
Merged here: https://github.com/pestphp/pest/commit/06d707fb4133b8223cb8e072eb790062d67fea64.

PR #100, opened 2020-06-19
Fix dataset name creation with objects

### LINKED ISSUE #98: Some tests do not traverse the whole dataset
Hey there,

First, thank you for the awesome work on Pest! 
This is the 1st time I use it on a project and it's a real delight.

However, I noticed that _sometimes_, for an unknown reason, the whole dataset is not entirely traversed when using the `with()` method.

Example [here](https://github.com/bpolaszek/mercure-php-hub/blob/623f8d5290406e599daf48688a4a65dcad7c97a7/tests/Unit/Security/TopicMatcher/Publish/TopicMatcherTest.php#L72-L79): 
`$combinations->asArray()` returns a dataset of `44` combinations (verified through [that first test](https://github.com/bpolaszek/mercure-php-hub/blob/623f8d5290406e599daf48688a4a65dcad7c97a7/tests/Unit/Security/TopicMatcher/Publish/TopicMatcherTest.php#L68-L70)), but only the `8th` first ones are run, as if the others simply didn't exist.

Here's the full output of the test:
```
   FAIL  Tests\Unit\Security\TopicMatcher\Publish\TopicMatcherTest
  ✓ it contains 44 combinations
  ✓ Topic can be updated only when authorized with ('/foo/bar', Lcobucci\JWT\Token Object (...), true, false)
  ✓ Topic can be updated only when authorized with ('/foo/bar', Lcobucci\JWT\Token Object (...), false, false)
  ✓ Topic can be updated only when authorized with ('/foo/bar', Lcobucci\JWT\Token Object (...), false, true)
  ✓ Topic can be updated only when authorized with ('/foo/bar', Lcobucci\JWT\Token Object (...), true, true)
  ✓ Topic can be updated only when authorized with ('/foo/baz', Lcobucci\JWT\Token Object (...), true, false)
  ✓ Topic can be updated only when authorized with ('/foo/baz', Lcobucci\JWT\Token Object (...), false, false)
  ✓ Topic can be updated only when authorized with ('/foo/baz', Lcobucci\JWT\Token Object (...), false, true)
  ✓ Topic can be updated only when authorized with ('/foo/baz', Lcobucci\JWT\Token Object (...), true, true)
  • All combinations have been tested

  ---

  • Tests\Unit\Security\TopicMatcher\Publish\TopicMatcherTest > All combinations have been tested
  Failed asserting that 8 matches expected 44.
```

Any ideas?

Thank you,
Ben

_PS: Running Pest 0.2 (used to fail also on 1.x) on PHP 7.4._

fetzi:
fixes #98

### DIFF
--- a/src/Datasets.php
+++ b/src/Datasets.php
@@ -91,7 +91,12 @@ final class Datasets
     private static function getDataSetDescription(array $data): string
     {
         $exporter = new Exporter();
+        $formatted = [];
 
-        return \sprintf(' with (%s)', $exporter->shortenedRecursiveExport($data));
+        foreach ($data as $value) {
+            $formatted[] = $exporter->export($value);
+        }
+
+        return \sprintf(' with (%s)', implode(', ', $formatted));
     }
 }
--- a/tests/Features/Datasets.php
+++ b/tests/Features/Datasets.php
@@ -3,6 +3,7 @@
 use Pest\Datasets;
 use Pest\Exceptions\DatasetAlreadyExist;
 use Pest\Exceptions\DatasetDoesNotExist;
+use Pest\Plugin;
 
 it('throws exception if dataset does not exist', function () {
     $this->expectException(DatasetDoesNotExist::class);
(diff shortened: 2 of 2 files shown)


owenvoke:
Does this still work recursively? 🤔

fetzi:
I think so but not from memory consumption side 😞

PR #109, opened 2020-06-21
pass calls to overloaded method if possible

### LINKED ISSUE #108: ReflectionException on overloaded method call
I've added a macro to `Illuminate\Testing\TestResponse` which I'm unable to call because pest throws a `ReflectionException`.

```
  1   vendor/pestphp/pest/src/Support/Reflection.php:30
      ReflectionClass::getMethod("assertMatchesHtmlSnapshot")

  2   vendor/pestphp/pest/src/Support/HigherOrderMessage.php:74
      Pest\Support\Reflection::call(Object(Illuminate\Testing\TestResponse), "assertMatchesHtmlSnapshot", [])

  3   vendor/pestphp/pest/src/Support/HigherOrderMessageCollection.php:33
      Pest\Support\HigherOrderMessage::call(Object(Illuminate\Testing\TestResponse))
```

I was able to fix it by adding a `try-catch` to the `Reflection::call()` method, check if `__call` exists and if so pass method call to overloaded `__call`.

```php
public static function call(object $object, string $method, array $args = [])
{
    $reflectionClass = new ReflectionClass($object);

    try {
        $reflectionMethod = $reflectionClass->getMethod($method);

        $reflectionMethod->setAccessible(true);

        return $reflectionMethod->invoke($object, ...$args);
    } catch(ReflectionException $ex) {
        if (method_exists($object, '__call')) {
            return $object->__call($method, $args);
        }

        throw $ex;
    }
}
```

Issue for https://twitter.com/devgummibeer/status/1274639315574894592

Gummibeer:
fixes #108

https://twitter.com/devgummibeer/status/1274639315574894592

### DIFF
--- a/src/Support/HigherOrderMessageCollection.php
+++ b/src/Support/HigherOrderMessageCollection.php
@@ -30,7 +30,7 @@ final class HigherOrderMessageCollection
     public function chain(object $target): void
     {
         foreach ($this->messages as $message) {
-            $target = $message->call($target);
+            $target = $message->call($target) ?? $target;
         }
     }
 
--- a/src/Support/HigherOrderMessageCollection.php
+++ b/src/Support/HigherOrderMessageCollection.php
@@ -30,7 +30,7 @@ final class HigherOrderMessageCollection
     public function chain(object $target): void
     {
         foreach ($this->messages as $message) {
-            $target = $message->call($target) ?? $target;
+            $target = $message->call($target);
         }
     }
 
--- a/tests/Features/Macro.php
+++ b/tests/Features/Macro.php
@@ -1,0 +1,12 @@
+<?php
+
+use Illuminate\Support\Traits\Macroable;
+use PHPUnit\Framework\TestCase;
+
+uses(Macroable::class);
+
+it('can call chained macro method')->macro('bar', function () {
+    assertInstanceOf(TestCase::class, $this);
+
+    return $this;
+})->bar();
--- a/tests/Features/Macro.php
+++ b/tests/Features/Macro.php
@@ -5,8 +5,16 @@ use PHPUnit\Framework\TestCase;
 
 uses(Macroable::class);
 
-it('can call chained macro method')->macro('bar', function () {
-    assertInstanceOf(TestCase::class, $this);
+beforeEach(function () {
+    $this->macro('bar', function () {
+        assertInstanceOf(TestCase::class, $this);
 
-    return $this;
-})->bar();
+        return $this;
+    });
+});
+
+it('can call chained macro method')->bar();
+
+it('will throw exception from call if no macro exists')
+    ->throws(BadMethodCallException::class)
+    ->foo();
--- a/tests/Features/Macro.php
+++ b/tests/Features/Macro.php
@@ -5,10 +5,8 @@ use PHPUnit\Framework\TestCase;
 
 uses(Macroable::class);
 
-beforeEach(function () {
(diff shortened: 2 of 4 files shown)


Gummibeer:
TestCase missing - will try to add one, that's why still draft.

src/Support/HigherOrderMessageCollection.php
```diff
@@ -30,7 +30,7 @@ public function add(string $filename, int $line, string $methodName, array $argu
     public function chain(object $target): void
     {
         foreach ($this->messages as $message) {
-            $target = $message->call($target);
```
Gummibeer, line 33: This could be breaking or not wanted behavior.
But for me it doesn't make sense to use `null` targets.
The better/more precise check would be a `is_object()` to prevent `TypeError` at all:
> Argument 1 passed to Pest\Support\HigherOrderMessage::call() must be an object, null given, called in src/Support/HigherOrderMessageCollection.php on line 33

src/Support/HigherOrderMessageCollection.php
```diff
@@ -30,7 +30,7 @@ public function add(string $filename, int $line, string $methodName, array $argu
...
+            $target = $message->call($target) ?? $target;
```
nunomaduro, line 33: Let's remove this for now.

tests/Features/Macro.php
```diff
@@ -0,0 +1,12 @@
...
+
+use Illuminate\Support\Traits\Macroable;
+use PHPUnit\Framework\TestCase;
+
+uses(Macroable::class);
+
+it('can call chained macro method')->macro('bar', function () {
+    assertInstanceOf(TestCase::class, $this);
+
+    return $this;
```
nunomaduro, line 11: You don't need this return.

src/Support/HigherOrderMessageCollection.php
Gummibeer: If I remove this the test won't pass anymore because `macro()` returns `null`.
It's a scenario only for the test - but I don't know how I should adjust the test to work without this? 🤔

nunomaduro:
Please re-ask me to be reviewer once you address my feedback.

src/Support/HigherOrderMessageCollection.php
Gummibeer: https://github.com/pestphp/pest/pull/109/commits/e135979f34bd6d9dcbffe5c10c4dbf4f9a2a7806

tests/Features/Macro.php
Gummibeer: https://github.com/pestphp/pest/pull/109/commits/bff97418edb5ca8b5493f24172780acbde2107b6

tests/Features/Macro.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+use Illuminate\Support\Traits\Macroable;
+use PHPUnit\Framework\TestCase;
+
+uses(Macroable::class);
+
+beforeEach(function () {
```
nunomaduro, line 8: `beforeEach()->macro(...`.

Gummibeer: Oh yeah, even better. You see I'm pretty new. 😄 
https://github.com/pestphp/pest/pull/109/commits/9a0cfaf33940af11471c89147c56735ad721d494

Gummibeer:
Thanks for your valuable feedback and patience! 🙂

PR #123, opened 2020-07-05
feat: add `expect` function

nunomaduro:
When we are writing tests, we often need to check that values meet certain conditions. And for that, at the time of this writing, we are using PHPUnit syntax using the global helpers: `assertEquals`, `assertTrue`, `assertFalse`, etc.

The problem with this is the number of global helpers that are injected in the global namespace: **One function per assertion**.

This Pull Request bootstraps the global helper `expect`, that would offer the same number of assertions of PHPUnit, but using a **single global function** and a **fluent syntax** like so:

```php
// old
assertTrue($variable);
// new
expect($variable)->toBe(true);
```

Fell free to help, and to add methods. All methods should be inspired by: **[jestjs.io/docs/en/expect](https://jestjs.io/docs/en/expect)**.


fetzi:
Should we add other methods directly to this PR, or should we wait for this to be merged?

owenvoke:
I guess wait for it to be merged @fetzi? Then we can discuss each new functionality separately? 🤔

nunomaduro:
@fetzi @owenvoke We can commit directly in this pull request.

One note tho, this package just released: https://github.com/verdephp/verde.

We can take ideas from it, or actually rease it behind the scenes.

ceceppa:
Hi, I'm the author of that package.

In case you want implement something similar in pest, I can help with that :)

nunomaduro:
@ceceppa Well, do you wanna lead this pull request? Actually the idea is copy the API we have on jestjs.io/docs/en/expect. But only the expectations, not mocking or mockey patching.

ceceppa:
Yep, happy to :).

In your example your using the PHP Unit Assert class.
Do you want me to carry on that, or you want something custom so that we could have error messages similar to the jest one?
I'm asking because jest prints a nice "diff" between the expected and receive values, and wondering if you want this or not. 
Also, in jest, you can pass a function and use his return value as expected one. Is also something you're looking for?

nunomaduro:
> Do you want me to carry on that, or you want something custom so that we could have error messages similar to the jest one?

Better carry with that at the moment. Using PHP Unit Assert class.

> Also, in jest, you can pass a function and use his return value as expected one. Is also something you're looking for?

Nop. Let's just handle assertions at the moment. No mocking or expect return values stuff.

nunomaduro:
@AlexMartinFR Wanna start tackle the docs about this? If don't have time ( or motivation ) I can do it.

owenvoke:
See my comment in the other PR for the fix for PHPStan. :+1: https://github.com/pestphp/pest/pull/125#pullrequestreview-447921943

AlexMartinFR:
I will have time next week to start something for these docs, if it's not too late :)

nunomaduro:
@ceceppa Can you confirm you have saw my message about missing methods? Let me know if you need me to do them. 👍

ceceppa:
Yes, I've seen it. I've added few more methods but didn't had time to push.

I'll update everything tomorrow

nunomaduro:
@ceceppa Great job on this! Thank you!

Let's wait for @AlexMartinFR to start the docs on this, fell free to actually do the docs yourself, in case  you wanna help.

ceceppa:
Hi @nunomaduro I'm away and so I'm not to help with the documentation.

PR #125, opened 2020-07-10
Feat/expect

ceceppa:
Adds more assertion methods for "expect".

@nunomaduro before I go too cray and add all the methods, check if this is what you're looking for :)


nunomaduro:
Make sure you run `composer update:snapshots` && `composer test` in your machine so tests can pass.

nunomaduro:
@ceceppa Did you found time to work on this?

ceceppa:
@nunomaduro yes, but this morning GitHub was down.
Can you please check the naming I've chosen for the directory* assertion? I wasn't sure on which name to go

nunomaduro:
@ceceppa Directory assertions assertions are not in the JEST API. Maybe we can add them later?

ceceppa:
I was implementing all the PHPUnit tests in "jest" style.

However, now I've implemented all the PHPUnit that have a jest equivalent and left all the file, JSON and XML related assertion.
Let me know what if is there anything else I should be adding.

Also, could you help me with the phpstan error? I couldn't figure out what does it mean...

src/Expectation.php
```diff
@@ -52,6 +52,347 @@ public function toBe($value): Expectation
...
+
+        return $this;
+    }
+
+    /**
+     * Assert that the value is an instance of $value.
+     *
+     * @param mixed $value
+     */
+    public function toBeInstanceOf($value): Expectation
```
owenvoke, line 259: Should `$value` not be type hinted as a string instead of mixed as that's all that PHPUnit supports for `assertInstanceOf()`? 🤔 It looks like this PHPStan error is related to the Psalm docblocks in PHPUnit.

https://github.com/sebastianbergmann/phpunit/blob/4272d6d7b51159e6de20e3276ef1d508e07ccdc1/src/Framework/Assert.php#L1317-L1339

You should be able to let PHPStan know that this is a `class-string<object>` type with:
```suggestion
     * @param class-string<object> $value
     */
    public function toBeInstanceOf(string $value): Expectation
```

nunomaduro:
Merging, and going to clean up the pull request on the main one.

nunomaduro:
@ceceppa Sorry, sounds very good the work so far. Fell free to keep adding the ones concerning the directory. 👍

nunomaduro:
@ceceppa Can you pull request against the original one again?

nunomaduro:
I think we should have assertions to all the methods on the assertions API, here are some missing ones:

`assertArrayHasKey`,  directory ones, ( checking some more ).

AlexMartinFR:
I will have time next week to start something for these docs, if it's not too late :)

nunomaduro:
@AlexMartinFR No problem. The idea is to add full documention about all the methods in the expectation API, just like in jest: https://jestjs.io/docs/en/expect#tobenull

AlexMartinFR:
Got it!

Issue #127, opened 2020-07-15
chore: missing tests against PHP 8


nunomaduro:
Currently, the test suite is testing against PHP 7.3, and PHP 7.4. To fix this issue, you should make sure our test suite is testing also against PHP 8. So we can be prepared for PHP 8 before it gets released. 👍🏻

To get started you should update https://github.com/pestphp/pest/blob/3eb0a95955750e4a83749d24462076e5aa1e9a33/.github/workflows/tests.yml#L12.

Finally, and if tests are failing, you should update the source code until the tests are passing.

nunomaduro:
@driesvints is also trying to test the Laravel Core against PHP 8: https://github.com/laravel/framework/compare/php-8-builds. His work may help us here.

GrahamCampbell:
Laravel 8 is coming along well now. A big blocker was Mockery, but I fixed up all that: https://github.com/mockery/mockery/pull/1072. It looks like you are facing a similar issue to mockery.

GrahamCampbell:
I think you can just delete these lines entirely. They are never hit:

<img width="1016" alt="image" src="https://user-images.githubusercontent.com/2829600/87592056-204bc800-c6e1-11ea-89b2-e3283de7c570.png">

GrahamCampbell:
Ish. I'll PR.

PR #128, opened 2020-07-15
feat: adds support to phpunit 9.3 and php 8.0

GrahamCampbell:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Fixed tickets | #127

This PR:

* Drops support for PHPUnit less than 9.3
* Adds support for PHP 8.0

Note that PHP 8.0.0-beta1 will be released on [Aug 06 2020](https://wiki.php.net/todo/php80) and PHPUnit 9.3.0 on [Aug 07 2020](https://github.com/sebastianbergmann/phpunit/milestone/38). I think this PR should be released simultaneously with PHPUnit 9.3.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,13 +17,12 @@
         }
     ],
     "require": {
-        "php": "^7.3",
+        "php": "^7.3 || ^8.0",
         "nunomaduro/collision": "^5.0.0-BETA4",
         "pestphp/pest-plugin": "^0.3",
         "pestphp/pest-plugin-coverage": "^0.3",
         "pestphp/pest-plugin-init": "^0.3",
-        "phpunit/phpunit": "^9.2.6",
-        "sebastian/environment": "^5.1.2"
+        "phpunit/phpunit": "9.3.2"
     },
     "autoload": {
         "psr-4": {
@@ -45,7 +44,7 @@
     "require-dev": {
         "illuminate/console": "^7.16.1",
         "illuminate/support": "^7.16.1",
-        "mockery/mockery": "^1.4.0",
+        "mockery/mockery": "^1.4.1",
         "pestphp/pest-dev-tools": "dev-master"
     },
     "minimum-stability": "dev",
--- a/src/Actions/ValidatesConfiguration.php
+++ b/src/Actions/ValidatesConfiguration.php
@@ -6,8 +6,7 @@ namespace Pest\Actions;
 
 use Pest\Exceptions\AttributeNotSupportedYet;
 use Pest\Exceptions\FileOrFolderNotFound;
-use PHPUnit\TextUI\Configuration\Configuration;
-use PHPUnit\TextUI\Configuration\Registry;
+use PHPUnit\TextUI\XmlConfiguration\Loader;
 
 /**
  * @internal
@@ -30,9 +29,7 @@ final class ValidatesConfiguration
             throw new FileOrFolderNotFound('phpunit.xml');
         }
 
-        $configuration = Registry::getInstance()
-            ->get($arguments[self::CONFIGURATION_KEY])
-            ->phpunit();
+        $configuration = (new Loader())->load($arguments[self::CONFIGURATION_KEY])->phpunit();
 
         if ($configuration->processIsolation()) {
             throw new AttributeNotSupportedYet('processIsolation', 'true');
--- a/src/Support/HigherOrderMessage.php
+++ b/src/Support/HigherOrderMessage.php
@@ -76,7 +76,7 @@ final class HigherOrderMessage
             Reflection::setPropertyValue($throwable, 'file', $this->filename);
             Reflection::setPropertyValue($throwable, 'line', $this->line);
 
-            if ($throwable->getMessage() === sprintf(self::UNDEFINED_METHOD, $this->methodName)) {
+            if ($throwable->getMessage() === self::getUndefinedMethodMessage($target, $this->methodName)) {
                 /** @var \ReflectionClass $reflection */
                 $reflection = new ReflectionClass($target);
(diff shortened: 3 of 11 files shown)


GrahamCampbell:
Looks like you are using PHPUnit code that was marked as internal, and was deleted on their 9.3 branch:

image

Also, it looks like there is a bug in composer preventing prefer-lowest from working. Kinda annoying this stops the other builds running.

GrahamCampbell:
The registry was deleted in https://github.com/sebastianbergmann/phpunit/commit/74fd29a34b6ab4804dd3f8777dd3acf0e59289de#diff-61ae1ab3f7e7fdca0d9366d1569777f6.

composer.json
```diff
@@ -17,7 +17,8 @@
         }
     ],
     "require": {
-        "php": "^7.3",
+        "php": "^7.3 || ^8.0",
+        "illuminate/support": "^6.18.25 || ^7.19.1 || ^8.0",
```
nunomaduro, line 21: Would be cool to copy/past the code used `Reflector::getParameterClassName` so we don't depend on this.

src/Actions/ValidatesConfiguration.php
```diff
@@ -30,12 +30,19 @@ public static function in($arguments): void
...
-            ->get($arguments[self::CONFIGURATION_KEY])
-            ->phpunit();
+        $configuration = self::getConfiguration($arguments[self::CONFIGURATION_KEY]);
 
         if ($configuration->processIsolation()) {
             throw new AttributeNotSupportedYet('processIsolation', 'true');
         }
     }
+
+    private static function getConfiguration($item)
```
nunomaduro, line 40: Let's make sure we add static types here.

composer.json
GrahamCampbell: Yeh, I was just thinking to do this. I think that'll fix the composer lowest issues.

src/Actions/ValidatesConfiguration.php
GrahamCampbell: 👍

GrahamCampbell: Actually, not possible to add a return type, since the type different on different phpunit versions. PHPStan will definitely not be happy. One thing we could do is drop support for PHPUnit less than 9.3, which would mean this PR would have to wait until August, when PHPUnit 9.3 is released. I think this is fine, because PHPUnit 9.2 fails with a parse error due to https://github.com/sebastianbergmann/phpunit/pull/4374 only being included in PHPUnit 9.3.

GrahamCampbell: I think dropping PHPUnit less than 9.3 will be the way to go for this PR, actually.

GrahamCampbell:
😆 Looks like is actually a bug in the PHPUnit code. It doesn't even type check. Better fix upstream first...

nunomaduro:
@GrahamCampbell Thank you for this! Will check this tomorrow. 💜

GrahamCampbell:
Cool. I think I can work through the core of the issues in the next 30 minutes or so. I think dropping PHPUnit less than 9.3 is gonna be the right approach. They also changed the signature of this method:

image

The current pest code calls passes an array of strings, and phpunit doesn't check the type, which causes issues down the line. That took some head scratching. 😆

GrahamCampbell:
Ok, got things working on my machine, apart from the visual test, which just needs to be re-generated since the execution order has changed, but is still perfectly valid.

GrahamCampbell:
At least Windows is passing so far. ;)

GrahamCampbell:
@nunomaduro This is ready for a first review now. :)

src/Support/HigherOrderMessage.php
```diff
@@ -87,4 +87,13 @@ public function call(object $target)
             throw $throwable;
         }
     }
+
+    private static function getUndefinedMethodMessage(object $target, string $methodName): string
```
nunomaduro, line 91: Self note: add tests about this.

tests/.snapshots/success.txt
```diff
@@ -167,5 +159,13 @@
    WARN  Tests\Visual\Success
   - visual snapshot of test suite on success
 
+   PASS  Tests\Features\Depends
```
nunomaduro, line 162: Self note: understand why this snapshot got different.

composer.json
```diff
@@ -17,13 +17,12 @@
...
     "require": {
-        "php": "^7.3",
+        "php": "^7.3 || ^8.0",
         "nunomaduro/collision": "^5.0.0-BETA4",
         "pestphp/pest-plugin": "^0.3",
         "pestphp/pest-plugin-coverage": "^0.3",
         "pestphp/pest-plugin-init": "^0.3",
-        "phpunit/phpunit": "^9.2.6",
-        "sebastian/environment": "^5.1.2"
+        "phpunit/phpunit": "9.3.2"
```
owenvoke, line 25: Just interested, is this intentionally a specific version rather than a range such as `^9.3`/`~9.3.2`? 🙂

GrahamCampbell: Yes, this is intentional, because we are using internal phpunit classes.

vincenzoraco:
Hi guys, this is just to remind you about this PR on a new tag unless you are already working on it. Thank you for your work!

GrahamCampbell:
We're aware of this PR. It will be released, probably at the start of September.

nunomaduro:
Going to merge this on master, but there is a few things left before tagging the `v0.3` version.

nunomaduro:
Of course, thanks @GrahamCampbell for your work on this. 👍

GrahamCampbell:
🥳

PR #134, opened 2020-07-19
Add name to description for named datasets

sshead:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | 

With named datasets, Pest does not currently include the dataset name in test output. This PR inserts the dataset name in the dataset description, so that the output of both the test result and failure messages mimic the PHPUnit format.

For example, given the test:

```
test('same pest style', function ($val) {
    $this->assertSame(1, $val);
})->with([
    'one' => [1],
    'two' => [2],
]);
```

the current output includes:
```
  ✓ same with (1)
  ⨯ same with (2)

  ---

  • Tests\Unit\Datasets > same with (2)
```

With this PR, the new output is:

```
  ✓ same with data set "one" (1)
  ⨯ same with data set "two" (2)

  ---

  • Tests\Unit\Datasets > same with data set "two" (2)
```

The PHPUnit failure output, for a test equivalent to the above, is:

```
1) DatasetsTest::testSame with data set "two" (2)
```

### DIFF
--- a/tests/Unit/Datasets.php
+++ b/tests/Unit/Datasets.php
@@ -8,6 +8,6 @@ it('show the names of named datasets in their description', function () {
         'two' => [[2]],
     ]));
 
-    $this->assertSame('test description with data set "one" (1)', $descriptions[0]);
-    $this->assertSame('test description with data set "two" (array(2))', $descriptions[1]);
+    expect($descriptions[0])->toBe('test description with data set "one" (1)');
+    expect($descriptions[1])->toBe('test description with data set "two" (array(2))');
 });
--- a/tests/Unit/Datasets.php
+++ b/tests/Unit/Datasets.php
@@ -1,0 +1,13 @@
+<?php
+
+use Pest\Datasets;
+
+it('show the names of named datasets in their description', function () {
+    $descriptions = array_keys(Datasets::resolve('test description', [
+        'one' => [1],
+        'two' => [[2]],
+    ]));
+
+    $this->assertSame('test description with data set "one" (1)', $descriptions[0]);
+    $this->assertSame('test description with data set "two" (array(2))', $descriptions[1]);
+});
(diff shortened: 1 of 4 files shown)


sshead:
Hmm... With the data, it can make for very long output lines - which you don't care about much if the test passed. But if a test fails, it's nice to see the data that failed, not just the name, in the context of the test suite output. And "only sometimes" would be a bigger refactor.

owenvoke:
Fair enough, that's fine. :+1: All good in my opinion.

nunomaduro (changes requested):
@sshead Can you rebase this pull request? Also, can you add a test that actually uses the feature and prints the output to the snapshot?

sshead:
Let me know if this wasn't what you were after, or if there's anything else I need to do. (I'm such a part-time "not my day job" dev that this is the first time I've ever force-pushed ... hope I didn't break the world 😅)

tests/Unit/Datasets.php
```diff
@@ -0,0 +1,13 @@
...
+
+use Pest\Datasets;
+
+it('show the names of named datasets in their description', function () {
+    $descriptions = array_keys(Datasets::resolve('test description', [
+        'one' => [1],
+        'two' => [[2]],
+    ]));
+
+    $this->assertSame('test description with data set "one" (1)', $descriptions[0]);
```
nunomaduro, line 11: Can you use the expectation API here?

nunomaduro (approved):
Sounds good to me, just left a comment.

tests/Unit/Datasets.php
sshead: Done.

nunomaduro (approved):
This is ready to merge and tag.

nunomaduro:
@sshead Congrats, great contribution.

PR #138, opened 2020-07-26
Polishing Expectation.php

AlexMartinFR:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Fixed tickets | -

- Streamlined comments.
- Removed potential confusion between `$value` & `$this->value`: now we use `$expected` & `$this->value`.

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -53,7 +53,7 @@ final class Expectation
      * value. Used on objects, it asserts that two
      * variables reference the same object.
      *
-     * @param mixed $value
+     * @param mixed $expected
      */
     public function toBe($expected): Expectation
     {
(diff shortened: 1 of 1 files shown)


AlexMartinFR:
While reviewing the Expectation API for the documentation, I spotted some potential improvements to make it more readable.

However, I have some PHPStan errors I can't seem to get rid of!
Any clue about that?

(At first I forgot to update the PHPDoc, but now it's fixed and the tests still fail)

owenvoke:
Yeah, I thought it was a cache issue but the PHPStan cache isn't cached in the action. 😕 I can't see any reason for it not to work. Looking at the log, it looks like it checked out the wrong commit. So going to re-run it.

AlexMartinFR:
Thanks for trying, I'll submit another PR with the proper file, it should be easier 👍

AlexMartinFR:
New PR: [https://github.com/pestphp/pest/pull/139](https://github.com/pestphp/pest/pull/139)

Issue #140, opened 2020-07-27
Support TeamCity format


olivernybroe:
For the phpstorm plugin to work correctly, pest has to create a usable teamcity output.

Right now we use the underlying phpunit teamcity printer, however this one relies on tests being in classes, which means that whenever a test start, I get the following data
```
##teamcity[testStarted name='it lives' locationHint='php_qn:///Users/olivernybroe/Code/dummy-project/vendor/pestphp/pest/src/Factories/TestCaseFactory.php(204) : eval()|'d code::\P\Tests\Unit\ExampleTest::it lives' flowId='83891']
````

I cannot use this location hint for anything, so we need our own printer for this format.

This can currently solve two issues for the phpstorm plugin
- https://github.com/pestphp/pest-intellij/issues/35
- https://github.com/pestphp/pest-intellij/issues/29

Possibly also solving:
- https://github.com/pestphp/pest-intellij/issues/32

[TeamCity docs](https://www.jetbrains.com/help/teamcity/build-script-interaction-with-teamcity.html#Reporting+Tests)

olivernybroe:
We now support teamcity format, closing this issue.

PR #141, opened 2020-07-28
feat(teamcity): Add basic teamcity output format

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #140 

This pull request adds team city output support to pest.

A future pr could be added, so the exception stack trace uses collision format.


nunomaduro:
Can you fix static analysis bugs buddy?

olivernybroe:
This should be ready now 👍

olivernybroe:
We probably could add some tests to check the output. However the biggest problem with the test is that we have to mimic the fake test case class we create 🙂

Might be worth it thougg as we prob. could use it for other tests also

olivernybroe:
I have no idea how to fix this rector error tbh 👍

@owenvoke you know how?

owenvoke:
Just investigating that issue, looks like it's due to the [`StaticCallOnNonStaticToInstanceCallRector`](https://github.com/rectorphp/rector/blob/1b059ee0c33df560233283cc95f24f47790317ca/rules/php70/src/Rector/StaticCall/StaticCallOnNonStaticToInstanceCallRector.php) rector.

It's caused by [line 82 of the TeamCity class](https://github.com/pestphp/pest/blob/c67411c64c475e828efe6932ff610247c3fed687/src/TeamCity.php#L82).

olivernybroe:
@owenvoke Sounds like an issue with `StaticCallOnNonStaticToInstanceCallRector` maybe? 

@TomasVotruba If you have a little time, could you help us on how to solve the rector error?

TomasVotruba:
Just flying by... Could you report and issue on Rector? With input/output/expected output.

If it a blocker, check the @rectorphp README to exclude single rule

olivernybroe:
@TomasVotruba Cool, I have created an issue https://github.com/rectorphp/rector/issues/3981 and will ignore the rule for now :)

olivernybroe:
I would really like to get this out as part of next version. 

The phpstorm plugin has a bunch of issues which can't be resolved before this is added.

owenvoke:
I thought this was already merged, I think it's fine. :+1:

owenvoke:
Probably my bad for missing it. 😬 I've now merged it. 🎉

olivernybroe:
Thanks!

PR #142, opened 2020-07-29
New expectations: toHaveKeys() & toContainAll()

AlexMartinFR:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | -

Adds new expectations:
- `toHaveKeys(array $keys)`
- `toContainAll(array $needles)`

Inspired by Freek in a recent livestream: https://youtu.be/KXUVnu-fTIM?t=3625

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -159,7 +159,7 @@ final class Expectation
     /**
      * Asserts that all of the provided $needles are elements of the value.
      *
-     * @param array{mixed} $needles
+     * @param array<mixed> $needles
      */
     public function toContainAll(array $needles): Expectation
     {
@@ -405,7 +405,7 @@ final class Expectation
     /**
      * Asserts that the value array has all of the provided $keys.
      *
-     * @param array{mixed} $keys
+     * @param array<mixed> $keys
      */
     public function toHaveKeys(array $keys): Expectation
     {
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -383,6 +383,18 @@ final class Expectation
     }
 
     /**
+     * Asserts that the value array has all of the provided $keys.
+     */
+    public function toHaveKeys(array $keys): Expectation
+    {
+        foreach ($keys as $key) {
+            Assert::assertArrayHasKey($key, $this->value);
+        }
+
+        return $this;
+    }
+
+    /**
      * Asserts that the value is a directory.
      */
     public function toBeDirectory(): Expectation
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -157,6 +157,26 @@ final class Expectation
     }
 
     /**
+     * Asserts that all of the provided $needles are elements of the value.
+     *
+     * @param array $needles
+     */
+    public function toContainAll(array $needles): Expectation
+    {
+        if (is_string($this->value)) {
+            foreach ($needles as $needle) {
+                Assert::assertStringContainsString($needle, $this->value);
+            }
+        } else {
+            foreach ($needles as $needle) {
+                Assert::assertContains($needle, $this->value);
+            }
+        }
+
+        return $this;
+    }
+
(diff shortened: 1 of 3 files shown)


src/Expectation.php
```diff
@@ -159,7 +159,7 @@ public function toContain($needle): Expectation
     /**
      * Asserts that all of the provided $needles are elements of the value.
      *
-     * @param array $needles
+     * @param array{mixed} $needles
```
canvural, line 162: This should be `array<mixed>`

`array{mixed}` describes the array shape. So it would mean in this array key `0` has `mixed` type value.

src/Expectation.php
```diff
@@ -404,6 +404,8 @@ public function toHaveKey(string $key): Expectation
 
     /**
      * Asserts that the value array has all of the provided $keys.
+     *
+     * @param array{mixed} $keys
```
canvural, line 408: Same as above. Should be array<mixed>

AlexMartinFR:
I'm not familiar with doc blocks best practices yet.

I'll commit this change 👍

src/Expectation.php
```diff
@@ -382,6 +402,20 @@ public function toHaveKey(string $key): Expectation
         return $this;
     }
 
+    /**
+     * Asserts that the value array has all of the provided $keys.
+     *
+     * @param array<mixed> $keys
```
nunomaduro, line 408: `@param array<int, int|string> $keys`.

AlexMartinFR: Of course!
Forgot about the tests, I'll get on it 😌

AlexMartinFR:
Writing the tests made me stumble upon a logic pitfall with the `not()` modifier.

When all values that shouldn't be here exist, the expected exception is thrown:

```php
test('not - fail when all strings exist')
    ->expect('Pest is awesome')
    ->not()->toContainAll(['Pest', 'awesome'])
    ->throws(ExpectationFailedException::class);
    // OK.
```

However, when only parts of the conditions are met, the exception is not thrown, and the test doesn't fail as it should:

```php
test('not - fail when at least one string exists')
    ->expect('Pest is awesome')
    ->not()->toContainAll(['Pest', 'clever'])
    ->throws(ExpectationFailedException::class);
    // Should fail, but is passing.
```

The same problem is encountered with `toHaveKeys`.

I have no clue how to circle around that, any idea?

nunomaduro:
@AlexMartinFR You code is correct:

```
test('not - fail when at least one string exists')
    ->expect('Pest is awesome')
    ->not()->toContainAll(['Pest', 'clever'])
    ->throws(ExpectationFailedException::class);
```

`Pest is awesome` does NOT contains all those keys. It contains some. 👍

nunomaduro:
@AlexMartinFR But maybe I expect that to contain none of the give values. Can you update the code?

AlexMartinFR:
You are right @nunomaduro, `not->toContainAll()` is working correctly, thanks to the word **all**.

I should have mentioned the real issue, `not->toHaveKeys()`:

```php
test('not - fails when at least one key exists')
    ->expect(['a' => 1, 'b', 'c' => 'world'])
    ->not()->toHaveKeys(['hello', 'a'])
    ->throws(ExpectationFailedException::class);

    // An exception should be thrown for 'a'.
    // However it's not the case.
```

I've been scratching my head around this one and can't find a way to make it work as I intend.

I could rename the expectation to `toHaveAllKeys()`, so that it matches the actual behaviour, but it doesn't seem ideal.

nunomaduro:
Can we keep this on old @AlexMartinFR? I will get back to it this month.

AlexMartinFR:
Sure!

nunomaduro:
Going to hold this for now, we maybe re-open this pull request later.

PR #146, opened 2020-08-02
add method / property override

### LINKED ISSUE #143: Overriding base testcase methods
[Linked tweet as requested ](https://twitter.com/enunomaduro/status/1288962431209869313)

As mentioned in the above conversation, there's a relatively common usecase for (abstract) base testcases giving extension points for the actual testcases. I ran into this when using Pest and [orchestral/testbench](https://github.com/orchestral/testbench#setup-environment) in conjunction, but it's also something I anticipate having trouble with in Laravel applications where I usually set up some generic API resource testing class.

I'd like to propose having the option to override protected methods/properties using the Pest syntax we've all grown to love, in a similar manner as the [uses option](https://pestphp.com/docs/underlying-test-case/). Although I'm not sure on what a good name would be, I _think_ it should be possible to implement using Reflection if nothing else.

nmokkenstorm:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | [#143](https://github.com/pestphp/pest/issues/143)

# Summary of the API:

A quick summary of this RFC's API.

extends can be used to extend a method or methods.
For extending a method supply the method name and the new value. You can also supply an array of method/value sets for overriding multiple methods or chain extends call.

```php
uses(MyTestCase::class)
    ->extends(
        'createBill',
        fn(string $name) => $this->billService->create($name)
    )
    ->extends([
        'containerProvides' => ['BillService', 'UserService', 'LoginService'],
        'initializeContainer' => function() {
            $container = parent::initializeContainer();
            $container->add(GoogleStorage::class, Mockery::spy(GoogleStorage::class));
        },
    ]);
```

For calling the parent instance method, simply use self::parent(). Calling static parent methods is currently not supported due to limitations within PHP itself.

with can be used to override a property.
For overriding a property, we can use the with method. It works the same way as extends except that it does not support closures, as statements in properties are forbidden.

```php
uses(MyTestCase::class)
    ->with('appName', 'MyAwesomePestApp')
    ->with([
        'useDatabase' => true,
        'databaseMode' => DatabaseMode::TRANSACTION,
    ]);
```

Fixes #143

### DIFF
--- a/src/Factories/TestCaseFactory.php
+++ b/src/Factories/TestCaseFactory.php
@@ -92,7 +92,7 @@ final class TestCaseFactory
     /**
      * Registered properties to override from base test class.
      *
-     * @var array <string, string|int|double|array>
+     * @var array <string, scalar|array|null>
      */
     public $properties = [];
 
--- a/src/Factories/TestCaseFactory.php
+++ b/src/Factories/TestCaseFactory.php
@@ -5,12 +5,15 @@ declare(strict_types=1);
 namespace Pest\Factories;
 
 use Closure;
+use ParseError;
 use Pest\Concerns;
 use Pest\Contracts\HasPrintableTestCaseName;
 use Pest\Datasets;
 use Pest\Exceptions\ShouldNotHappen;
+use Pest\Repositories\MethodProxyRepository;
 use Pest\Support\HigherOrderMessageCollection;
 use Pest\Support\NullClosure;
+use Pest\Support\Reflection;
 use Pest\TestSuite;
 use PHPUnit\Framework\TestCase;
 
@@ -66,7 +69,7 @@ final class TestCaseFactory
     /**
      * The FQN of the test case class.
      *
-     * @var string
+     * @var class-string
      */
     public $class = TestCase::class;
 
@@ -78,6 +81,20 @@ final class TestCaseFactory
     public $traits = [
         Concerns\TestCase::class,
     ];
+
+    /**
+     * Registered methods to override from base test class.
+     *
+     * @var array <string, \Closure>
+     */
+    public $methods = [];
+
+    /**
+     * Registered properties to override from base test class.
+     *
+     * @var array <string, string|int|double|array>
+     */
+    public $properties = [];
 
     /**
      * Holds the higher order messages
@@ -146,6 +163,7 @@ final class TestCaseFactory
 
         $createTest = function ($description, $data) use ($className, $test) {
             $testCase = new $className($test, $description, $data);
+
             $this->factoryProxies->proxy($testCase);
 
             return $testCase;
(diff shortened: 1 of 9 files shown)


nunomaduro:
@olivernybroe @owenvoke Can you give me your insights on this one. What do you think about the proposed API and the value of it?

nmokkenstorm:
The usecase is mostly for stuff like [testbench](https://github.com/orchestral/testbench), where the developer is supplied a testcase that they're expected to extend/override where needed.

For what it's worth, there's some PHP 7.4 & 8.0 stuff that's not supported right now, but I'd be happy to give that another shot if/when #197 gets merged.

olivernybroe:
I think the functionality is something we should have support for as the current approach would be to create a class that extends your testcase and do the overrides there.

I have a few questions about it.
- If I have a property and a method which are named the same, but returns different data, how do I specify that?
- What if I wanna return an anonymous function?
- Can I decorate the original method? e.g. call the parent method?


Also I don't see much value in this for other than extending existing methods, as pest has it's own way for adding helper methods. So I think the naming should be something like `extends`. 

I think it would also be nice if I could do the following for cases where the extends require a lot of lines.
```php
uses(MethodOverriddenTest::class)
    ->extends('assertOverride', function() {
        // many lines of php code
    })
    ->extends('foo', function() {
        // many lines of php code
    });
```

nmokkenstorm:
> * If I have a property and a method which are named the same, but returns different data, how do I specify that?

No can do, as right now we differentiate between the two based on the value of the array. 

> * What if I wanna return an anonymous function?

Same for the reason above. Both could possibly be remedied by having two different methods for both cases instead of using one version. Open for thoughts on this re: the API design @nunomaduro menioned

> * Can I decorate the original method? e.g. call the parent method?

We rebind the scope of the closure / the value of `$this`, so my naive thought would be that `parent::method()` should just work, but I'm going to test that.

> Also I don't see much value in this for other than extending existing methods, as pest has it's own way for adding helper methods. So I think the naming should be something like `extends`.

That is the one case it's meant for, as in the end we end up dynamically creating methods and proxying that.

> I think it would also be nice if I could do the following for cases where the extends require a lot of lines.
> 
> ```
> uses(MethodOverriddenTest::class)
>     ->extends('assertOverride', function() {
>         // many lines of php code
>     })
>     ->extends('foo', function() {
>         // many lines of php code
>     });
> ```

```
uses(FooTest::class)
    ->extends('longMethod', function () {
        //
    })
    ->extends([
        'firstShortMethod' => fn () => 'foo',
        'secondShortMethod' => fn () => 'bar',
    ])
    ->with([
        'propertyOne' => 'foo',
        'propertyTwo' => 'bar',
    ])
    ->with('propertyThree', 'baz')
```

Contrived and extreme example, but I think this solves all potential problems / edge cases, and is relatively easy to implement

nmokkenstorm:
> We rebind the scope of the closure / the value of `$this`, so my naive thought would be that `parent::method()` should just work, but I'm going to test that.

Doesn't work out of the box, am gonna check if I can solve that 😁

olivernybroe:
@nmokkenstorm Awesome, thanks for the detailed reply.

I think with these last changes we talked about I would say the API is ready 👍

## Summary of the API:
A quick summary of this RFC's API.

### `extends` can be used to extend a method or methods. 
For extending a method supply the method name and the new value. You can also supply an array of method/value sets for overriding multiple methods or chain `extends` call.

```php
with(MyTestCase::class)
    ->extends(
        'createBill',
        fn(string $name) => $this->billService->create($name)
    )
    ->extends([
        'containerProvides' => ['BillService', 'UserService', 'LoginService'],
        'initializeContainer' => function() {
            $container = $this::parent();
            $container->add(GoogleStorage::class, Mockery::spy(GoogleStorage::class));
        },
    ]);
```

For calling the parent method, simply do `$this::parent()`.

### `with` can be used to override a property.
For overriding a property, we can use the with method. It works the same way as `extends` except that it does not support closures, as statements in properties are forbidden.

```php
with(MyTestCase::class)
    ->with('appName', 'MyAwesomePestApp')
    ->with([
        'useDatabase' => true,
        'databaseMode' => DatabaseMode::TRANSACTION,
    ]);
```

nmokkenstorm:
> @nmokkenstorm Awesome, thanks for the detailed reply.

No problem, thanks a lot for taking the time to think with me 😄 

> ```php
> with(MyTestCase::class)->extends('initializeContainer', function() {
>             $container = $this::parent();
>             $container->add(GoogleStorage::class, Mockery::spy(GoogleStorage::class))
>         });
> ```
> 
> For calling the parent method, simply do `$this::parent()`.

Can't get this to to work, I briefly hoped I learned something new about accessing parent scopes, but no dice. For some reason static methods/constants do work, but accessing `self::parent` or `$this::parent()` does not.

olivernybroe:
@nmokkenstorm I think it is fine to just point that limitation out in the docs. 👍

nunomaduro:
Didn't followed the discussion, but why `with` instead of the normal `uses` method?

nmokkenstorm:
> Didn't followed the discussion, but why `with` instead of the normal `uses` method?

I _think_ @olivernybroe made a mistake, and I copied it without realising. It should be

```php
uses(TestCase::class)
    ->with([..])
    ->extends([...])
```

I have a fix for the `parent::` thing for the instance case, will try to get that in tonight or tomorrow. A known limitation will be `parent::` calls in static methods, which can be called out as such in the docs. I am thinking about raising a bug report for PHP itself to get that resolved though.

olivernybroe:
@nunomaduro Fixed my comment with the summary. :) it was a simple mistake, did it on my phone. 

@nmokkenstorm sounds good. Could you add a summary (maybe just copy they one I wrote) to the issue itself? Makes it easier for users to see what this is.

nmokkenstorm:
> @nmokkenstorm sounds good. Could you add a summary (maybe just copy they one I wrote) to the issue itself? Makes it easier for users to see what this is.

This issue, the php-src issue, or the docs? For the latter I'll be happy to send in a PR, which will probably be a modified version of your summary anyways, if that's okay with you.

olivernybroe:
@nmokkenstorm docs :) That's perfect, spend no to writing it, so would prefer that.

I think it makes sense to report it to php-src as a bug, however they might not say it's a bug as you can do it in another way. But worth a try!

olivernybroe (approved):
LGTM! 👍 

@nunomaduro or @owenvoke could you look at the implementation also?

nmokkenstorm:
Hey, I'll happily rebase this PR to fix the conflict in the snapshot tests, but I'd also love to have more feedback on this, @nunomaduro @owenvoke ?

nunomaduro:
For the moment I am going to close this pull request. Now, this don't mean your pull request won't be included in PEST soon - I am just too unsure about the API I want for this kind of thing.

Thank you for your work on this pull request, I promise you that I will take a second look on this pull request once we re-discuss an API for this.

PR #147, opened 2020-08-03
Rename phpunit.xml to phpunit.xml.dist

voyula:


GrahamCampbell:
This only makes sense if you also add `phpunit.xml` to the .gitignore file.

GrahamCampbell:
Well, if those files aren't updated, then this PR adds no value. In fact, it makes things worse because it means the file is shipped with the dist zips composer downloads, when it wasn't before.

PR #148, opened 2020-08-03
Rename .php_cs to .php_cs.dist

voyula:


GrahamCampbell:
This only makes sense if you also add `.php_cs` to the .gitignore file.

PR #151, opened 2020-08-03
Added additional rector sets

GrahamCampbell:
Let's see what this does... I've not run this on my machine. :trollface: EDIT: seems fine


GrahamCampbell:
I think the purpose of those sets is to make it use the features of those versions, so the intended use is to enable all php sets up to the 7.3?

GrahamCampbell:
There's also a php52 ruleset, but it doesn't do anything useful. ;)

nunomaduro:
Ok, thanks!

PR #155, opened 2020-08-04
Provide dist file support

voyula:


voyula:
Is this repo abandoned?

voyula:
Seems like repo abandoned.

owenvoke:
I've been on holiday, and other people can be busy. The repo isn't abandoned. :+1: I think this PR looked fine, but it's only been open for 4 days so it can be missed.

PR #169, opened 2020-08-24
Forward TestCase bad method calls to global functions

felixdorn:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

```php
// tests/Helpers.php

function assertSomething() {}
```

```php
// tests/SomeTest.php

it('works')->assertSomething();
```

Right now, any functions can be called like `substr` or `dd`, maybe we want to restrict that behavior to function asserting something. (if there is no new assertions, we still throw a BadMethodCallException) ?

This is still WIP but it works, without breaking any existing test.

### DIFF
--- a/src/Support/Reflection.php
+++ b/src/Support/Reflection.php
@@ -6,6 +6,7 @@ namespace Pest\Support;
 
 use Closure;
 use Pest\Exceptions\ShouldNotHappen;
+use Pest\TestSuite;
 use ReflectionClass;
 use ReflectionException;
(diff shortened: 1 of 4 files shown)


nunomaduro:
Can you rebase?

felixdorn:
I've managed to do it! That was more complicated that expected :smile:

owenvoke (approved):

nunomaduro:
Great! Can you update the docs about this?

PR #186, opened 2020-09-15
proof of concept custom expectations

adamthehutt:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

This is a proof of concept for custom expectations. The basic idea is that you'd optionally have an "Expectations.php" file in your root tests directory, which can define any number of expectation functions. Such functions must return true/false to indicate whether the expectation is met or not. Expectation functions would be passed the current value as the first parameter and any additional parameters afterward.

### DIFF
--- a/src/Actions/LoadStructure.php
+++ b/src/Actions/LoadStructure.php
@@ -23,6 +23,7 @@ final class LoadStructure
     private const STRUCTURE = [
         'Datasets.php',
         'Helpers.php',
+        'Expectations.php',
         'Pest.php',
         'Datasets',
     ];
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -494,4 +494,22 @@ final class Expectation
         /* @phpstan-ignore-next-line */
         return $this->{$name}();
     }
+
+    /**
+     * Dynamically calls custom expectation if it can be found
+     *
+     * @param string $method
+     * @param array $args
+     */
+    public function __call(string $method, array $args)
+    {
+        if (function_exists($method)) {
+            $result = $method($this->value, ...$args);
+            Assert::assertTrue($result);
+
+            return $this;
+        }
+
+        throw new \InvalidArgumentException("Could not find expectation for $method");
+    }
 }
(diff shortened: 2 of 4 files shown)


owenvoke:
Would it be better to namespace them maybe to reduce the functions in the global namespace? Seems like a neat idea. Was just thinking like `namespace Expectations` in the file, but not sure. :thinking:

adamthehutt:
@owenvoke Yeah, I had the same thought. The only reason I didn't do it was for PSR autoloading conventions. Personally, I would also prefer invokable classes to functions, but that seemed very un-pest-like. :)

olivernybroe:
I like this. Pretty much follows the same conventions as the `Helpers.php` file instead that we this time around are binding them to `expect`.

I actually think this is a pretty clean implementation 👍

olivernybroe:
Another simple solution could also to have a static method on Expectation? Using something like Laravel's Macroable?

So in my `Expectations.php` or just directly in `Helpers.php` I would do
```php
Expectation::extend('toBeAwesome', fn($value) => $value === 'awesome');
```
or with a helper, like when we write tests
```php
extend('toBeAwesome', fn($value) => $value === 'awesome');
```

owenvoke:
@olivernybroe, that's the way I was thinking would be nicer. `Expectation::macro('name', function () {});`

olivernybroe:
@owenvoke Yeah, I like it 👍  

Think it's pretty clear and it's a syntax familiar for a lot of users.

owenvoke:
@olivernybroe, what're your thoughts on this implementation? https://github.com/pestphp/pest/compare/master...owenvoke:feature%2Fexpectation-macro

nunomaduro:
@owenvoke So the macros would go to `Pest.php` file or?

owenvoke:
@nunomaduro, yes. Or if you wanted to make a plugin, you could just macro in your plugin's autoload file. 👍🏻

olivernybroe:
@owenvoke I like it. Think it's a good idea to just use Laravel's Macroable and it's pretty readable and easy to implement.

nunomaduro:
We already have this feature, thanks!

PR #190, opened 2020-09-17
feat(expectations): add toMatchConstraint

owenvoke:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #... 

This allows the use of [Constraint classes](https://github.com/sebastianbergmann/phpunit/tree/master/src/Framework/Constraint) for more complex validation.

```php
// From basic constraints
expect(true)->toMatchConstraint(new IsTrue());

// To more complex constraints
expect(true)->toMatchConstraint(
    $this->logicalAnd(
         $this->logicalNot(new IsFalse()),
         new IsType(IsType::TYPE_BOOL)
    )
);

// To custom constraints
expect('https://google.com')->toMatchConstraint(new IsValidUrlConstraint());

class IsValidUrlConstraint extends PHPUnit\Framework\Constraint
{
    public function toString()
    {
        return 'is a valid url';
    }

    protected function matches($other): bool
    {
        if (! is_string($other)) {
            return false;
        }

        return preg_match(
            Symfony\Component\Validator\Constraints\UrlValidator::PATTERN,
            $other
        ) > 0;
    }
}
```

Custom constraints should extend `PHPUnit\Framework\Constraint`, and provide a [`matches()`](https://github.com/sebastianbergmann/phpunit/blob/9830d0461c8c7b28e473c3a8b713847ae22eff41/src/Framework/Constraint/Constraint.php#L87) and [`toString()`](https://github.com/sebastianbergmann/phpunit/blob/9830d0461c8c7b28e473c3a8b713847ae22eff41/src/Framework/SelfDescribing.php#L20) method, and optionally override the [`evaluate()`](https://github.com/sebastianbergmann/phpunit/blob/9830d0461c8c7b28e473c3a8b713847ae22eff41/src/Framework/Constraint/Constraint.php#L42) method.

PR #191, opened 2020-09-18
feat(expectations): add toMatch

owenvoke:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #... 

I wasn't sure about the naming of this, whether it should be `toMatchRegEx`, `toMatchRegExp`, or `toMatchRegularExpression`. (The `toBe*` prefix doesn't really work with any of them).

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -506,6 +506,16 @@ final class Expectation
     }
 
     /**
+     * Asserts that the value matches a regular expression.
+     */
+    public function toMatchRegEx(string $expression): Expectation
+    {
+        Assert::assertMatchesRegularExpression($expression, $this->value);
+
+        return $this;
+    }
+
+    /**
      * Asserts that the value matches a constraint.
      */
     public function toMatchConstraint(Constraint $constraint): Expectation
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -508,7 +508,7 @@ final class Expectation
     /**
      * Asserts that the value matches a regular expression.
      */
-    public function toMatchRegEx(string $expression): Expectation
+    public function toMatch(string $expression): Expectation
     {
         Assert::assertMatchesRegularExpression($expression, $this->value);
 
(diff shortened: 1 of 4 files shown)


olivernybroe:
What if we just call it `toMatch` instead? 

Then let me do the following 
```php
expect('Hello World')->toMatch('/^hello wo.*$/i');
expect('Hello World')->toMatch('Hello World'); 
```

So a regex and a string is supported for matching.

owenvoke:
@olivernybroe, that makes sense too. 👍🏻 Although, I wonder if people might get confused between `toBe` and `toMatch`. 🤔 @nunomaduro, what're your thoughts on this?

olivernybroe:
@owenvoke Hmm true. Could we extend `toBe` to just support regex maybe?

nunomaduro:
So, to give context. Normally, we mimic the API of Jest. In this case, Jest don't have this feature. So let's take a moment to actually make sure we don't make a mistake here.

JavaScript uses `match` : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match.


I think we can use `toMatchRegex` for now, and change it later if we think people will be confortable with an `toMatch`.

olivernybroe:
@nunomaduro According to jest they use `toMatch` in their expect api
Which can take a string and a regex or am I misunderstanding something here? 👍 

https://jestjs.io/docs/en/expect#tomatchregexporstring

nunomaduro:
@owenvoke @olivernybroe If Jest uses `toMatch`, let's use `toMatch` too. 👍

owenvoke:
Pushed an update. 👍🏻

nunomaduro (approved):

nunomaduro:
Fell free to merge, tag a new version, update the docs, and make a tweet about it @owenvoke .

owenvoke:
@nunomaduro, I noticed you merged the docs for `toMatchConstraint`. Do you want me to merge #190 as well and include that in v0.3.6?

nunomaduro:
@owenvoke Yes.

PR #193, opened 2020-09-24
chore(composer): add support for PHPUnit 9.3.11

owenvoke:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Fixed tickets | #... 

I've also updated to use a version selector, instead of a long list of all the patches that are supported.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "pestphp/pest-plugin": "^0.3",
         "pestphp/pest-plugin-coverage": "^0.3",
         "pestphp/pest-plugin-init": "^0.3",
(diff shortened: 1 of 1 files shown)


nunomaduro:
If tests are passing, this kind of stuff you can merge and tag. No need for approval.

owenvoke:
I more just wanted to check that changing the version selector syntax was ok. Also got busy with work. 😬

lcdss:
Why allowing phpunit ^9.0.0 would be a problem? I didn't understand the necessary to change the supported versions one by one.

owenvoke:
@lcdss, it's because we use internal PHPUnit classes which aren't covered by the backwards compatibility guarantee. So they may be changed in a patch/minor version and Pest would break.

Issue #184, opened 2020-09-14
Adding skip() modifier on higher order test throws an exception.


DannyvdSluijs:
Adding the `skip()` modifier on higher order test throws an exception.

Given the following code:
```php
<?php

test('Runs another basic test')
    ->assertTrue(true)
    ->skip();
```

You get an error when running `./vendor/bin/pest`:
```bash

   FAIL  Tests\Unit\ExampleTest
  ⨯ Runs another basic test

  ---

  • Tests\Unit\ExampleTest > Runs another basic test
   TypeError 

  Argument 1 passed to Pest\Support\HigherOrderMessage::call() must be an object, null given, called in /Users/danny/Projects/Infi/PestPlayground/vendor/pestphp/pest/src/Support/HigherOrderMessageCollection.php on line 33



  Tests:  1 failed
  Time:   0.05s
```

For the sake of context adding the `only()` modifier to a higher order test works perfectly fine.

DannyvdSluijs:
After some debugging I've found that in `vendor/pestphp/pest/src/Support/HigherOrderMessageCollection.php` the above mentioned test results in two higher order messages the first for the method name `assertTrue` and the second method name is `markTestSkipped`.

On line 33 of the HigherOrderMessageCollection the target is overwritten with the return value. With the `assertTrue` method returning void the second message is invoked with a <NULL> for `$target` resulting in the error mentioned above. #196 is my attempt to resolve this bug.

PR #196, opened 2020-09-29
Revert to original target if new target is null

DannyvdSluijs:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #184 

This fixes higher order tests where an assert is followed by any other higher order message. This is due to the fact that the PHPUnit assert methods (located in `vendor/phpunit/phpunit/src/Framework/Assert.php`) return void.

### DIFF
--- a/src/Support/HigherOrderMessageCollection.php
+++ b/src/Support/HigherOrderMessageCollection.php
@@ -30,7 +30,7 @@ final class HigherOrderMessageCollection
     public function chain(object $target): void
     {
         foreach ($this->messages as $message) {
-            $target = $message->call($target);
+            $target = $message->call($target) ?? $target;
         }
     }
 
--- a/tests/.snapshots/success.txt
+++ b/tests/.snapshots/success.txt
@@ -278,6 +278,7 @@
 
    PASS  Tests\Features\HigherOrderTests
   ✓ it proxies calls to object
+  ✓ it is capable doing multiple assertions
 
    PASS  Tests\Features\It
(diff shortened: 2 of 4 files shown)


src/Support/HigherOrderMessageCollection.php
```diff
@@ -30,7 +30,7 @@ public function add(string $filename, int $line, string $methodName, array $argu
     public function chain(object $target): void
     {
         foreach ($this->messages as $message) {
-            $target = $message->call($target);
+            $target = $message->call($target) ?? $target;
```
nunomaduro, line 33: Can you add some tests? 

Also, `??` is equal `=== null` correct?

nunomaduro (changes requested):

src/Support/HigherOrderMessageCollection.php
owenvoke: Yeah, it will fallback to `$target` if the result of the left-hand value is null or is not set. 👍🏻

nunomaduro:
Please add some unit tests.

nunomaduro:
Can you make a pull request on the docs about this?

DannyvdSluijs:
Anything specific you would like to see added to the docs? Adding a higher order example to the the page on [skipping tests](https://pestphp.com/docs/skipping-tests/)?

nunomaduro:
Just propose what you personally would like to see in the docs there.

PR #197, opened 2020-10-02
Skip tests based on PHP version

felixdorn:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

It provides the ability to skip a test for a given PHP version.

Maybe we should also add a `onlyForVersion()` too.

### DIFF
--- a/src/PendingObjects/TestCall.php
+++ b/src/PendingObjects/TestCall.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Pest\PendingObjects;
 
 use Closure;
+use Pest\Expectation;
 use Pest\Factories\TestCaseFactory;
 use Pest\Support\Backtrace;
 use Pest\Support\NullClosure;
@@ -13,7 +14,7 @@ use PHPUnit\Framework\ExecutionOrderDependency;
 use SebastianBergmann\Exporter\Exporter;
 
 /**
- * @method \Pest\Expectation expect(mixed $value)
+ * @method Expectation expect(mixed $value)
  *
  * @internal
  */
@@ -78,7 +79,7 @@ final class TestCall
      * Runs the current test multiple times with
      * each item of the given `iterable`.
      *
-     * @param \Closure|iterable<int, mixed>|string $data
+     * @param Closure|iterable<int, mixed>|string $data
      */
     public function with($data): TestCall
     {
@@ -128,6 +129,17 @@ final class TestCall
     }
 
     /**
+     * Skips the test for a given version of PHP.
+     */
+    public function skipForVersion(string $version, string $operator = null, string $message = ''): TestCall
+    {
+        return $this->skip(
+            version_compare(PHP_VERSION, $version, $operator ?? '=='),
+            $message
+        );
+    }
+
+    /**
      * Skips the current test.
      *
      * @param Closure|bool|string $conditionOrMessage
@@ -155,6 +167,17 @@ final class TestCall
         }
 
         return $this;
+    }
+
+    /**
+     * Runs the test only for a given version of PHP.
+     */
+    public function onlyForVersion(string $version, string $message = ''): TestCall
+    {
+        return $this->skip(
+            version_compare(PHP_VERSION, $version, '=='),
+            $message
+        );
     }
 
     /**
--- a/tests/Features/Skip.php
+++ b/tests/Features/Skip.php
@@ -17,11 +17,15 @@ it('skips with message')
     ->assertTrue(false);
 
 it('skips with truthy closure condition')
-    ->skip(function () { return '1'; })
+    ->skip(function () {
+        return '1';
+    })
     ->assertTrue(false);
 
 it('do not skips with falsy closure condition')
-    ->skip(function () { return false; })
+    ->skip(function () {
+        return false;
+    })
     ->assertTrue(true);
 
 it('skips with condition and message')
@@ -31,3 +35,19 @@ it('skips with condition and message')
 it('skips when skip after assertion')
     ->assertTrue(true)
     ->skip();
+
+it('skips if the PHP version is the current one (with message)')
+    ->skipForVersion(PHP_VERSION, null, 'because of the current version')
+    ->assertTrue(true);
+
+it('skips if the PHP version is greater that 7 (should skip)')
+    ->skipForVersion('7', '>')
+    ->assertTrue(true);
+
+it('skips if the PHP version is less than 7 (should never skip)')
+    ->skipForVersion('7', '<')
+    ->assertTrue(true);
+
+it('runs only if the PHP version is the current one (should never skip)')
+    ->onlyForVersion(PHP_VERSION)
+    ->assertTrue(true);
--- a/tests/.snapshots/success.txt
+++ b/tests/.snapshots/success.txt
@@ -395,5 +395,5 @@
   ✓ depends with defined arguments
   ✓ depends run test only once
 
-  Tests:  10 skipped, 228 passed
+  Tests:  10 skipped, 232 passed
   
--- a/tests/.snapshots/success.txt
+++ b/tests/.snapshots/success.txt
@@ -304,6 +304,10 @@
   ✓ it do not skips with falsy closure condition
   - it skips with condition and message → skipped because foo
   - it skips when skip after assertion
+  - it skips if the PHP version is the current one (with message) → because of the current version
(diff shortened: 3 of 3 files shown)


olivernybroe:
Could you please elaborate on what this actually solves? 
Like what is the scenario that you have a need for this method?

src/PendingObjects/TestCall.php
```diff
@@ -140,7 +152,8 @@ public function skip($conditionOrMessage = true, string $message = ''): TestCall
 
         $condition = is_callable($condition)
             ? $condition
-            : function () use ($condition) { /* @phpstan-ignore-line */
+            : function () use ($condition) {
+                /* @phpstan-ignore-line */
```
owenvoke, line 156: Looks like this was accidentally reformatted.
```suggestion
            : function () use ($condition) { /* @phpstan-ignore-line */
```

src/PendingObjects/TestCall.php
```diff
@@ -157,6 +170,40 @@ public function skip($conditionOrMessage = true, string $message = ''): TestCall
         return $this;
     }
 
+    /**
+     * Compare two given PHP versions.
```
owenvoke, line 174: We should add typehints here. 👍🏻
```suggestion
     * Compare two given PHP versions.
     *
     * @param string|array $compared
     * @param string|array $version
```

I feel like it's probably better to use `version_compare()` in this method.

tests/Features/Skip.php
```diff
@@ -17,11 +17,15 @@
     ->assertTrue(false);
 
 it('skips with truthy closure condition')
-    ->skip(function () { return '1'; })
+    ->skip(function () {
+        return '1';
+    })
```
owenvoke, line 22: Looks like this was accidentally reformatted.
```suggestion
    ->skip(function () { return '1'; })
```

tests/Features/Skip.php
```diff
@@ -17,11 +17,15 @@
...
     ->assertTrue(false);
 
 it('do not skips with falsy closure condition')
-    ->skip(function () { return false; })
+    ->skip(function () {
+        return false;
+    })
```
owenvoke, line 28: Looks like this was accidentally reformatted.
```suggestion
    ->skip(function () { return false; })
```

owenvoke (commented):
Hey @felixdorn, I just thought I'd apply my review comments from Discord as an actual review so it's easier to keep track of them. 👍🏻

I'd also suggest a few extra tests. However, I feel like if `skipForVersion()` only supports a single version, we could probably just do:
```php
public function skipForVersion(string $version, string $operator = null, string $message = ''): TestCall
{
    return $this->skip(
        version_compare($version, PHP_VERSION, $operator ?? '=='),
        $message
    );
}
```

And exclude the `sameVersion()` method.

src/PendingObjects/TestCall.php
```diff
@@ -131,10 +131,10 @@ public function group(string ...$groups): TestCall
     /**
      * Skips the test for a given version of PHP.
      */
-    public function skipForVersion(string $version, string $message = ''): TestCall
+    public function skipForVersion(string $version, string $operator = null, string $message = ''): TestCall
     {
         return $this->skip(
-            $this->sameVersion($version),
+            version_compare(PHP_VERSION, $version, $operator ?? '==') === 0,
```
owenvoke, line 137: When using the 3rd parameter (operator), the `version_compare()` function will return a boolean. 👍🏻 So it doesn't need the `=== 0`. Same with the other one. 🙂

felixdorn: Changed that!

src/PendingObjects/TestCall.php
```diff
@@ -140,7 +152,7 @@ public function skip($conditionOrMessage = true, string $message = ''): TestCall
 
         $condition = is_callable($condition)
             ? $condition
-            : function () use ($condition) { /* @phpstan-ignore-line */
+            : function () use ($condition): bool {
```
owenvoke, line 155: @nunomaduro, I just wanted to check if there was a reason for not type hinting this before. 👍🏻

src/PendingObjects/TestCall.php
felixdorn: I've used `version_compare()`

src/PendingObjects/TestCall.php
felixdorn: Fixed!

src/PendingObjects/TestCall.php
felixdorn: With the type hint here it throws an error because it returns an integer for some reason.

tests/Features/Skip.php
```diff
@@ -31,3 +35,19 @@
 it('skips when skip after assertion')
     ->assertTrue(true)
     ->skip();
+
+it('skips if the PHP version is the current one (with message)')
+    ->skipForVersion(PHP_VERSION, 'because of the current version')
```
owenvoke, line 40: This is the broken test. 👍🏻 The 2nd character is the operator.
```suggestion
    ->skipForVersion(PHP_VERSION, null, 'because of the current version')
```

felixdorn:
@owenvoke I think my PR is ready, I've fixed the snapshots and rebased.

tests/Features/Skip.php
```diff
@@ -31,3 +35,19 @@
 it('skips when skip after assertion')
     ->assertTrue(true)
     ->skip();
+
+it('skips if the PHP version is the current one (with message)')
+    ->skipForVersion(PHP_VERSION, null, 'because of the current version')
+    ->assertTrue(true);
+
+it('skips if the PHP version greated that 7 (should skip)')
```
owenvoke, line 43: If you could fix this (very minor) typo, and rebase with the updated snapshots (as some other tests have been added) this seems ready to me. 👍🏻 Thanks for your work on this @felixdorn!
```suggestion
it('skips if the PHP version greater that 7 (should skip)')
```

owenvoke (approved):

nmokkenstorm:
> Hey @felixdorn, I just thought I'd apply my review comments from Discord as an actual review so it's easier to keep track of them. 👍🏻
> 
> I'd also suggest a few extra tests. However, I feel like if `skipForVersion()` only supports a single version, we could probably just do:
> 
> ```
> public function skipForVersion(string $version, string $operator = null, string $message = ''): TestCall
> {
>     return $this->skip(
>         version_compare($version, PHP_VERSION, $operator ?? '=='),
>         $message
>     );
> }
> ```
> 
> And exclude the `sameVersion()` method. I've created a diff for this if you want to have a look (with additional tests).
> https://gist.github.com/owenvoke/326c275befb8f2a28f365d93fe27f17f

Nitpick, I thought having `string $operator = null` in the function signature like that shouldn't work, surprised it does. Dut having `string $operator = '=='` is definitely correct and makes the actual `version_compare` cleaner because you don't need to check for `null`

owenvoke:
True, it doesn't really make sense without a nullable string type hint (but they aren't allowed in our PHPStan rules).

nunomaduro:
@felixdorn Can you resolve conflicts ?

tests/Features/Skip.php
```diff
@@ -31,3 +35,19 @@
 it('skips when skip after assertion')
     ->assertTrue(true)
     ->skip();
+
+it('skips if the PHP version is the current one (with message)')
+    ->skipForVersion(PHP_VERSION, null, 'because of the current version')
+    ->assertTrue(true);
+
+it('skips if the PHP version is greater that 7 (should skip)')
+    ->skipForVersion('7', '>')
```
nunomaduro, line 44: I think would be cool if we can use this API Like so:

```
->skipForVersion('7.0.1', 'reason'); // specific version
->skipForVersion('^7.0', 'reason'); // any version bigger than 7 but lower than 8, like composer.
```

nunomaduro (changes requested):

tests/Features/Skip.php
owenvoke: I had originally thought about that, but we'd probably want to pull in another package such as [`composer/semver`](https://github.com/composer/semver). 🤷🏻

Then it would be:
```php
Composer\Semver\Semver::satisfies($version, $constraint);
```

nunomaduro:
Going to hold on this pull request. I don't see the demand for this feature. Maybe in the future we can reconsider this.

PR #199, opened 2020-10-02
feat: adds support to PHPUnit 9.4

owenvoke:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Fixed tickets | #...

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "pestphp/pest-plugin": "^0.3",
         "pestphp/pest-plugin-coverage": "^0.3",
         "pestphp/pest-plugin-init": "^0.3",
-        "phpunit/phpunit": ">= 9.3.7 <= 9.3.11"
+        "phpunit/phpunit": ">= 9.3.7 <= 9.4.0"
     },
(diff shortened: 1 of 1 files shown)


nunomaduro:
There is no more patch releases on phpunit 9.3.x series?

owenvoke:
That's what I wasn't sure about. I was assuming there at least wouldn't be breaking changes to 9.3.x, but I guess there could be backports. Perhaps this should be `>= 9.3.7 <= 9.3.11 || 9.4.0` for now.

nunomaduro:
> Perhaps this should be >= 9.3.7 <= 9.3.11 || 9.4.0 for now.

Yes, do it. Then you can merge and release.

GrahamCampbell:
There will be no more 9.3.x releases of phpunit. I like the PR in it's current state.

owenvoke:
Thanks for the clarification @GrahamCampbell, if that's ok with you @nunomaduro, shall I merge and release as is?

nunomaduro:
Yes

Issue #205, opened 2020-10-13
Pest does not support phpunit 9.4.1


olivernybroe:
PhpUnit 9.4.1 has been released, and the contraints disallow us for installing pest with phpunit 9.4.1

PR #207, opened 2020-10-13
Update phpunit dependency constraint

fetzi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  no
| Fixed tickets | #205 

allow all phpunit versions greater or equal to 9.3.7

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "pestphp/pest-plugin": "^0.3",
         "pestphp/pest-plugin-coverage": "^0.3",
         "pestphp/pest-plugin-init": "^0.3",
-        "phpunit/phpunit": ">= 9.3.7 <= 9.4.0"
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -22,7 +22,7 @@
         "pestphp/pest-plugin": "^0.3",
         "pestphp/pest-plugin-coverage": "^0.3",
         "pestphp/pest-plugin-init": "^0.3",
-        "phpunit/phpunit": ">= 9.3.7 <= 9.4.0"
+        "phpunit/phpunit": ">= 9.3.7"
```
owenvoke, line 25:
```suggestion
Can you update it to use an upper limit so this can be merged? We are extending internal classes so can't just allow any version.
                "phpunit/phpunit": ">= 9.3.7 <= 9.4.1"
```

owenvoke (changes requested):

owenvoke (approved):

PR #209, opened 2020-10-15
Get root path based on composer autoloader

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | https://github.com/pestphp/pest-intellij/issues/73

So this is not necessarily something we want to support, so wanna get the rest of your guys feedback on it.

So basically the issue is that @NickSdot has a pretty strange folder structure, where the root of the project is actually not the root of php application.
The phpstorm plugin set's the CWD to be the project root, however Pest expects CWD to be the php application's root. This can prob. be solved in the plugin by changing the CWD to the php application root.

 The solution in this pr is created by @NickSdot.

It basically suggests a structure where we guess the php application root from the composer autoloader path instead. This solution should be viable as composer enforces a structure where the php application root will always be just outside the `vendor` folder.

The downside with this solution is you can't have a global composer pest installed and use that for running tests.

### DIFF
--- a/bin/pest
+++ b/bin/pest
@@ -19,16 +19,16 @@ use Symfony\Component\Console\Output\OutputInterface;
 
     if (file_exists($vendorPath)) {
         include_once $vendorPath;
-        $rootPath = dirname($vendorPath, 2);
+        $autoloadPath = $vendorPath;
     } else {
         include_once $localPath;
-        $rootPath = dirname($localPath, 5);
+        $autoloadPath = $localPath;
     }
 
     (new Provider())->register();
 
-    // fallback in case we couldn't find out the path before.
-    $rootPath = (isset($rootPath) && !empty($rootPath)) ? $rootPath : getcwd();
+    // get $rootPath based on $autoloadPath
+    $rootPath =  dirname($autoloadPath, 2);
 
     $testSuite = TestSuite::getInstance($rootPath);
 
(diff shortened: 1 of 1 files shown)


nunomaduro:
> The downside with this solution is you can't have a global composer pest installed and use that for running tests.

Nobody does that.

> The solution in this pr is created by @NickSdot.

Can you try to detail the potential impact on existing users of this solution?

NickSdot:
> Can you try to detail the potential impact on existing users of this solution?

I was thinking about this. In general there should be no impact on existing users, thought. 

In standard structures `getcwd();` must be the same as counting up the autoload.php path. Which is what we do now with my fix. 

In non standard structures there was no chance to find the `composer.json` yet – so this neither should have a negative impact. 

What is actually the reason to check for the `composer.json`? I couldn't find out this yet.

olivernybroe:
@NickSdot I think we use it to find all registered pest plugins. They are registered in composer.json

iksaku:
I would love for this to happen.

Just stumbled across this issue last night, configuring some tests for a [little advent project](https://github.com/iksaku/AdventOfCode/tree/173e7afef013a8dde9486719601274053fab2086).

When running from CLI, it works great, as you only need to `cd` in your root folder, but when working with PHPStorm, you can get the `vendor` directory just fine, but somehow it'll still miss the directory for `composer.json` file.

If you need any help testing, ping me up. Or if you could quickly guide me though the source code, I could try to make a PR for this.

olivernybroe:
@iksaku You are welcome to test it out 👍 If this solution also works for your problem, then it might help escalate merging it.

nunomaduro:
@olivernybroe Going to merge this, and let's keep this on master a few days before we release the v1.

PR #215, opened 2020-11-04
Improve exception output in teamcity printer

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no

Our teamcity printer didn't do the cleanup on the exception which phpunit does, so I added that in.

```php
it("fails", function () {
    throw new Exception("fails");
});
```

It still does not use collision, as I cannot figure out how to make that work. However this will at least help a lot of users with debugging their exceptions.

New result:
image
Old result:
image


nunomaduro:
Fell free to merge this @olivernybroe, and tag a new release with this feature.

PR #216, opened 2020-11-09
fix(depends): resolve issue with dependency names

### LINKED ISSUE #213: it and depends
When a test created with "it", the depends function doesn't work.

````php
it('can create a converter', function () {
    $factory = new ConverterFactory();
    $factory->register('md', MarkdownConverter::class);
    $converter = $factory->createConverter('md');
    expect($converter)->toBeInstanceOf(MarkdownConverter::class);
    return $converter;
});

it('can convert markdown', function ($converter) {
    $html = $converter->convert('**TEST**');
    expect($html)->toBe('<p><strong>TEST</strong></p>');
})->depends('it can create a converter');
````
This results in the following output:

````
WARN  Tests\Core\Infrastructure\Converter\ConverterFactoryTest
  ✓ it can create a converter
  ! it can convert markdown → This test depends on "P\Tests\Core\Infrastructure\Converter\ConverterFactoryTest::can create a converter" which does not exist.

  Tests:  1 warnings, 1 passed
  Time:   0.24s
````
Leaving "it" out of the testname doesn't work, but adding "it" twice to the testname works:

````php
it('can convert markdown', function ($converter) {
    $html = $converter->convert('**TEST**');
    expect($html)->toBe('<p><strong>TEST</strong></p>');
})->depends('it it can create a converter');
````
````
   PASS  Tests\Core\Infrastructure\Converter\ConverterFactoryTest
  ✓ it can create a converter
  ✓ it can convert markdown

  Tests:  2 passed
  Time:   0.29s
````

### PR DESCRIPTION
**Pull request by owenvoke**
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | fixes #213

This resolves #213, although it needs some thorough testing.

The issue is that [`ExecutionOrderDependency::createFromDependsAnnotation()`](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/ExecutionOrderDependency.php#L46) splits the provided test name by spaces in order to determine the ["clone option"](https://phpunit.readthedocs.io/en/9.3/writing-tests-for-phpunit.html#writing-tests-for-phpunit-test-dependencies). We currently use this in the [`TestCall::depends()`](https://github.com/pestphp/pest/blob/d85432933cfd916f691a564ccd10dab82d008dd6/src/PendingObjects/TestCall.php#L98) method, so when depending on a test with spaces in the name, you need to prefix it with another word.

For example:

```php
// Resulting testname is `it asserts true is true`
it('asserts true is true')->assertTrue(true);

// This does not work, despite the test name being correct
it('does not work')->assertTrue(true)->depends('it asserts true is true');

// This works, however it uses `it it` in the test name (clone option would be "it")
it('works... but needs a double "it"')->assertTrue(true)->depends('it it asserts true is true');

// This also works, and the "clone option" is set to `bah`
it('works... but is the wrong name')->assertTrue(true)->depends('bah it asserts true is true');
```

See https://github.com/pestphp/pest/issues/213#issuecomment-723937696 for more background details.

This does mean that we drop support for clone options, although... I'm not really sure what they are and whether Pest users are using them anyway...

### DIFF
--- a/src/PendingObjects/TestCall.php
+++ b/src/PendingObjects/TestCall.php
@@ -95,7 +95,11 @@ final class TestCall
         $className = $this->testCaseFactory->getClassName();
 
         $tests = array_map(function (string $test) use ($className): ExecutionOrderDependency {
-            return ExecutionOrderDependency::createFromDependsAnnotation($className, $test);
+            if (strpos($test, '::') === false) {
+                $test = "{$className}::{$test}";
+            }
+
+            return new ExecutionOrderDependency($test, null, '');
         }, $tests);
 
         $this->testCaseFactory
--- a/tests/.snapshots/success.txt
+++ b/tests/.snapshots/success.txt
@@ -392,10 +392,12 @@
    PASS  Tests\Features\Depends
   ✓ first
   ✓ second
+  ✓ it asserts true is true
   ✓ depends
   ✓ depends with ...params
   ✓ depends with defined arguments
   ✓ depends run test only once
+  ✓ depends works with the correct test name
 
-  Tests:  7 skipped, 233 passed
+  Tests:  7 skipped, 235 passed
   
--- a/tests/Features/Depends.php
+++ b/tests/Features/Depends.php
@@ -32,3 +32,7 @@ test('depends with defined arguments', function (string $first, string $second) 
 test('depends run test only once', function () use (&$runCounter) {
     expect($runCounter)->toBe(2);
 })->depends('first', 'second');
+
+// Regression tests. See https://github.com/pestphp/pest/pull/216
+it('asserts true is true')->assertTrue(true);
+test('depends works with the correct test name')->assertTrue(true)->depends('it asserts true is true');


owenvoke:
Requesting a review from Graham as it was added in https://github.com/pestphp/pest/commit/3cfadee2bb960457d4c1e4e9cd51faff309808ff 👍🏻 Hope you don't mind.

olivernybroe:
@owenvoke Overall I think this makes sense, however I am also not sure what the clone option is, so don't really know what the implications will be 😟

nunomaduro:
This need some tests... @owenvoke

owenvoke:
Hard to do tests but I guess it can be 3 tests like in the example?

nunomaduro:
I trust you to find a good test for this. If you think is not worth, fell free to merge.

owenvoke:
Well, I've done some tests, although they cause Pest to exit with a warning:
image

Which would fail CI. 🤔 Even if you skip the test, it throws the warning before. I guess I could just add the two passing tests, as they fail on the current version of Pest, and if they fail in future then we know there's a regression.

---

On a side note, I think this wasn't spotted before because the tests in the `depends()` file use single-word test names (i.e. `first` and `second`).

nunomaduro (approved):

nunomaduro:
@owenvoke This should be merged?

owenvoke:
@nunomaduro, apologies been away. Yes, it should. 👍🏻

nunomaduro:
@owenvoke Well I get this got tagged. Can you add a entry to the change log?

owenvoke:
Sure. 👍🏻 Just a heads up, you forgot to update the version number in `src/Pest.php`.

PR #217, opened 2020-11-10
feat: add Pest options to help output

owenvoke:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #... 

This adds an additional help section to the output of `pest --help`.

Preview screenshot of help command output


tests/.snapshots/success.txt
```diff
@@ -348,6 +348,9 @@
   ✓ it throws exception when `process isolation` is true
   ✓ it do not throws exception when `process isolation` is false
 
+   PASS  Tests\Unit\Console\Help
+  ✓ it outputs the help information when --help is used
```
nunomaduro, line 352: We may need an integration test on this. Do you mind of adding it?

nunomaduro (changes requested):

tests/.snapshots/success.txt
owenvoke: Sure. :+1: So just call `pest --help` and check it's in the output?

owenvoke: Is that ok? Or should they be in the same file? 🤔

tests/Visual/Help.php
```diff
@@ -0,0 +1,13 @@
...
+test('visual snapshot of help command output', function () {
+    $output = function () {
+        $process = (new Symfony\Component\Process\Process(['php', 'bin/pest', '--help', '--colors=never']));
+
+        $process->run();
+
+        return $process->getOutput();
+    };
+
+    expect($output())->toContain('Pest Options:');
```
nunomaduro, line 12: I would prefer this line to contain a snapshot. So we can see the entire thing. Sorry.

nunomaduro (changes requested):

tests/Visual/Help.php
owenvoke: No problem 👍🏻 Updated.

nunomaduro (approved):

PR #228, opened 2020-11-26
feat: add support for macroing Expectation

owenvoke:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #... 

This is an alternative to #186, it adds support for using the `::macro()` method on the `Expectation` class.

This allows the ability for users or plugins to add new custom expectations.

For example:

```php
use Pest\Expectation;

Expectation::macro('toBeResource', function () {
    Assert::assertIsResource($this->value);

    return $this;
});

// Write test
it('asserts resource is resource', function () {
    expect($resource)->toBeResource();
});
```

Note that any assertions can be used, although you must return `$this` when assertions are finished.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
     ],
     "require": {
         "php": "^7.3 || ^8.0",
+        "illuminate/macroable": "^8.16",
         "nunomaduro/collision": "^5.0",
         "pestphp/pest-plugin": "^0.3",
         "pestphp/pest-plugin-coverage": "^0.3",
--- a/composer.json
+++ b/composer.json
@@ -42,8 +42,8 @@
         ]
     },
     "require-dev": {
-        "illuminate/console": "^7.16.1",
-        "illuminate/support": "^7.16.1",
+        "illuminate/console": "^8.0",
+        "illuminate/support": "^8.0",
         "mockery/mockery": "^1.4.1",
         "pestphp/pest-dev-tools": "dev-master"
     },
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Pest;
 
+use Illuminate\Support\Traits\Macroable;
 use PHPUnit\Framework\Assert;
 use PHPUnit\Framework\Constraint\Constraint;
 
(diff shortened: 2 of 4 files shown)


composer.json
```diff
@@ -18,6 +18,7 @@
     ],
     "require": {
         "php": "^7.3 || ^8.0",
+        "illuminate/macroable": "^8.16",
```
nunomaduro, line 21: Can we copy this code? So we don't depend from it?

nunomaduro (changes requested):

nunomaduro:
@owenvoke @olivernybroe How about?

```php
expectation('toBeResource', function () {
    Assert::assertIsResource($this->value);

    return $this;
});

// Write test
it('asserts resource is resource', function () {
    expect($resource)->toBeResource();
});
```

composer.json
owenvoke: Sure, it's just a single trait. 👍🏻 If we are doing it inline, it should be fine to always return `$this`, and then we can simplify to:
```php
Expectation::macro('toBeResource', function () {
    Assert::assertIsResource($this->value);
});
```

olivernybroe:
@nunomaduro To be honest, I like both 👍 

But prob. best to only have one way of doing it right?

owenvoke:
Personally I prefer `::macro()`, mainly because it's the same as other Laravel macros, and it doesn't introduce another global function. I don't mind either way though.

nunomaduro:
How is works in JEST:
Screenshot 2020-11-28 at 19 54 36

owenvoke:
@nunomaduro, are you suggesting an `expect()->extend('name', function () {});` method?

nunomaduro:
```php
expect()->extend('name', function () {});
```

Actually I like this...

olivernybroe:
Well I in general feel the word extend is better than macro.

I think I would personally use `Expectation:: extend`

hackel:
Probably too late here, but one big advantage of using Macroable is mixins.  A Pest plugin can provide a single class which gets mixed in to Expectation to add all of the additional methods it has in one shot.

owenvoke:
That's a good point @hackel, I guess we could add mixin support in future. 🤷🏻 I'm a big fan of mixins. 👍🏻

PR #231, opened 2020-11-27
feat: adds key/value context when toMatchObject or toMatchArray fails

nunomaduro:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

This pull request context to the user about what is the key/values that fails while using `toMatchObject` or `toMatchArray`.

Screenshot 2020-11-27 at 21 53 56

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -6,6 +6,7 @@ namespace Pest;
 
 use PHPUnit\Framework\Assert;
 use PHPUnit\Framework\Constraint\Constraint;
+use SebastianBergmann\Exporter\Exporter;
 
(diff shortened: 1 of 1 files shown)


olivernybroe:
Could you add a test for it also? :)

nunomaduro:
We already have a couple of tests on this. And to be honest, that would be re-testing base phpunit "Assert" messages. I am not sure.

HENRIQUELOBO:
image
“NAME” field informed in the error message is the correct one, what is wrong is the “AGE”, did this pull fix it?

PR #232, opened 2020-11-29
feat: makes expect extendable

nunomaduro:
This pull request proposes yet another alternative to (#186) and (#228) adding the following API for add methods to `expect()`:

```php
expect()->extend('toBeWithinRange', function ($min, $max) {
    return $this->toBeGreaterThanOrEqual($min)->toBeLessThanOrEqual($max);
});

test('numeric ranges', function () {
    expect(100)->toBeWithinRange(90, 110);
});
```

### DIFF
--- a/src/Concerns/Extendable.php
+++ b/src/Concerns/Extendable.php
@@ -1,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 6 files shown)


owenvoke (approved):
Brilliant!

PR #234, opened 2020-12-04
Update PHPUnit dependency version

soilSpoon:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Fixed tickets | #...


soilSpoon:
@nunomaduro Is there any reason why the phpunit dependency version is not specified as `^9.3.7`?

owenvoke:
Because we use internal classes from PHPUnit which might break in a minor or patch release.

PR #256, opened 2021-01-14
Add shorthand functions `xtest` and `xit` for skipping tests

thled:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

Adding shorthand for skipping tests by adding an `x` prefix to the test functions, e.g. `xtest()` and `xit()`.

Example:

```php
xit('has home', function () {
// ..
});
```

is the same as

```php
it('has home', function () {
// ..
})->skip();
```

This is equivalent to the aliases in Jest (see https://jestjs.io/docs/en/api#testskipname-fn).

Would love to see this feature merged, because I often find myself skipping a newly written test to do a small refactoring before I can solve the new test.

### DIFF
--- a/src/Functions.php
+++ b/src/Functions.php
@@ -72,6 +72,16 @@ function test(string $description = null, Closure $closure = null)
 }
 
 /**
+ * Shorthand function to skip the current test.
+ *
+ * @return TestCall|TestCase|mixed
+ */
+function xtest(string $description, Closure $closure = null): TestCall
+{
+    return test($description, $closure)->skip();
+}
+
+/**
  * Adds the given closure as a test. The first argument
  * is the test description; the second argument is
  * a closure that contains the test expectations.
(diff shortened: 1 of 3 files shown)


owenvoke:
Interesting way of shortening it.

In a way, I feel like this should be a plugin with namespaced functions, just because there are already a few global functions. 🤷🏻 But up to Nuno.

leonstafford:
Works great like this in https://github.com/ShellSpec/ShellSpec, along with "f" for focus on one test and conditional "Skip when condition" and other nice bits, mostly inspired from other test frameworks.

nunomaduro:
I don't know. I feel that would be better to put this on plugin maybe?

octoper:
I agree with @nunomaduro this will be better as a plugin.

nunomaduro:
That been said. We have two options:

1. An official plugin, that will be place for this globals.
2. A community plugin, in your own organisation.

thled:
Thank you for your response. I will look into making this functionality available as a plugin.

thled:
I created a plugin for the shorthand functions: https://github.com/thled/pest-plugin-shorthands

owenvoke:
@thled, nice work!

nunomaduro:
@thled Can you tweet about it?

PR #263, opened 2021-01-24
feat: phpunit 10 support

nunomaduro:


nunomaduro:
@lukeraymonddowning @owenvoke Any of you would like to spend time on this one? Won't wanna lie, this may take some time to investigate...

owenvoke:
I might have another go at it, should also be easier to manage now that those core plugins have been merged. 👍🏻

composer.json
```diff
@@ -17,10 +17,10 @@
         }
     ],
     "require": {
-        "php": "^7.3 || ^8.0",
+        "php": "^7.4 || ^8.0",
```
owenvoke, line 20: Looks like PHPUnit 10 is going to only support PHP 8 and above, so I guess we'll want to mirror that. 👍🏻 (Same on the workflows).
```suggestion
        "php": "^8.0",
```

nunomaduro: Nice! Now we have reason for this. Ahah.

Issue #268, opened 2021-02-13
Custom suffixes with "multiple" file extensions result in invalid classnames


jordanbrauer:
I was just upgrading from PHPUnit 6 to PHPUnit 9 + Pest (love this library!), and noticed that the PHPUnit `suffix` configuration option causes some issue with the [`TestCaseFactory`](https://github.com/pestphp/pest/blob/master/src/Factories/TestCaseFactory.php#L182-L205) if it contains "multiple" file extensions, like so,

```xml
<testsuites>
    <testsuite name="all">
        <directory suffix=".spec.php">tests/</directory>
    </testsuite>
</testsuites>
```

Which when executing the suite or a single test yields the following error,

![image](https://user-images.githubusercontent.com/18744334/107843131-25afe480-6d8e-11eb-8e31-e38a616ce289.png)

Taking a quick peak at what's going on, we can see that the PHP code passed to `eval` contains said syntax error in the classname.

![image](https://user-images.githubusercontent.com/18744334/107842569-e5e6fe00-6d89-11eb-8bf7-27f8813da881.png)

---

I made [a quick PR](https://github.com/pestphp/pest/pull/269) here as a proof of concept and am pretty sure there is a more robust solution, but figured I'd raise the issue before getting too into it and adding a test. If this seems like too much of an edge case, I don't mind closing this and just sticking with `Test.php` suffix.

PR #269, opened 2021-02-13
fix: allow multiple file extensions in test suffix (prevent class & file name syntax errors)

jordanbrauer:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| Fixed tickets | #268 

Allow custom test suffixes with "multiple" file extensions, such as

```
Example.spec.php
```

### Changes

The goal of this MR in addition to the 1 very small thing raised in #268, is to make the TestCaseFactory a little more bullet-proof when it comes to how many different ways a file could be named, and provide a fallback attempt or an error explaining why the test failed to run/parse.

#### Tests

Listed below are the changes and additions made to the test code, backing the [source changes](#Source).

- add tests to cover various file names ("affixes") that could/would cause the `TestCaseFactory` to throw fatal errors;
  - `@#$%^&()-_=+.php` – valid file name; invalid PHP class name;
  - `A Test With Spaces.php` – valid filename; invalid PHP class name;
  - `AdditionalFileExtension.spec.php` – original cause of PR (#268)
  - `kebab-case-spec.php` – valid filename; invalid PHP class name;
  - `ManyExtensions.class.test.php` – potentially many extra extensions;
  - `snake_case_spec.php` – valid filename; invalid PHP class name;
  - `Test 'Case' With Quotes.php` – valid filename; invalid PHP classname & breaks `private static \$__filename` in the form of "code injection";
- regenerate the `Visual\Success` snapshot, containing new test output;
  - also a small addition to the snapshot generation, removing additional random data;

#### Source

Listed below are the changes and additions made to the source code for this PR.

**Allowing Multiple File Extensions:**

- fix/update the final `$relativePath` replacement regexp to `/[^A-Za-z0-9\\\\]/`, removing dots;
  - it also appears that the regexp was "invalid" due to a missing escape, so we essentially replace `.` with `\` 😁;

image

**Preventing Other Syntax Errors:**

- **In the event that a class name can't be determined** for whatever reason, generate some random one so that the tests can still run;
  - The class name for these tests is `InvalidTestName%s` where `%s` is a 16-byte long alpha-numeric string, ensuring we do not collide with any other class names;
  - To facilitate this, I've added a `random` method to the `Str` support class, taking some simple logic from an earlier version of laravel (noted in the doc block);
- Escape the filename to ensure that ones with quotes in them do not break the syntax for `private static $__filename`;
  - We also have to retroactively remove the escaped backslash for Windows paths;
- `try...catch` the `eval` process in `TestCaseFactory` and inform the user about the `ParseError` from their input;

image


nunomaduro:
This looks cool, don't forget to add some tests.

owenvoke:
Just interested, how does PHPUnit expect classes with this sort of filename? I guess if you had 2 files (`Example.php` and `Example.spec.php`), would you name them both `Example`? 🤔

jordanbrauer:
> would you name them both Example? I guess not as **that would cause a conflict.** 

Correct. AFAIK PHPUnit just handles it the same way `Test`, or any other suffix is handled. Assumably, after PHPUnit drops the file extension from the test file, it will normalize the rest of the name as `PascalCase`, thus eliminating a whole class of issues such as this one. But that is just a guess, I haven't looked at their code since PHPUnit 6 😅 .. 

But from my understanding, here is it how it works,

| Filename | Classname |
|:---------|:------------|
| ExampleTest.php | ExampleTest |
| Example.spec.php | ExampleSpec |
| Example.class.spec.php | ExampleClassSpec |

jordanbrauer:
@nunomaduro, @owenvoke – the branch is good for review now 👍 . I've updated the PR description with an explanation of my changes for ya'.

nunomaduro:
@owenvoke Can you take over here?

owenvoke (approved):

nunomaduro:
Thank you, @jordanbrauer!

Issue #255, opened 2021-01-14
Reusable beforeEach and beforeAll


albertpratomo:
I want to make a Trait that does something in the `beforeAll` of the test.

Is it possible to use the `beforeAll()` function in a trait? And then call `uses(Trait::class)` in the test case? I was having problem with that..

Also I tried to use PHPUnit TestCase methods like `setUp` and `setUpBeforeClass`, but it says that it conflicts with Pest's TestCaseFactory..

jordanbrauer:
@albertpratomo not sure if you have found a workaround for you use case yet, but I have been sharing closures through static constructors, like so (note that this is just sample/pseudo-code and not specific to any framework)

```php
namespace YourApp\Test;

final class Setup
{
    public static function logout(): Closure
    {
        return fn (): void => auth()->logout();
    }

    public static function rollback(): Closure
    {
        return fn (): void => DB::rollback();
    }

    public static function txn(): Closure
    {
        return fn (): void => DB::beginTransaction();
    }
}
```

and used like so

```php
beforeEach(Setup::txn());
afterEach(Setup::rollback());
```

I haven't really needed to in my own project, _yet_, but if you were to need more than one to run, you could easily compose them, and just pass the final closure to `beforeEach`, `beforeAll`, etc.

You could probably even, as an invokable object just autoload the setups you need instead of one with static methods, like so, while maintaining composability since at the core you're just dealing in closures (i.e., callables).

```php
namespace YourApp\Test\Setup;

final class MyFeatureSetup
{
    public function __invoke(): void
    {
        // check & call all the stuff you need to here;
    }
}
```

and then use it like;

```php
beforeEach(new MyFeatureSetup());
```

---

Hope this helps!

nunomaduro:
@jordanbrauer Would be cool to actually have something to allow to configure global beforeBeachs...

```
uses()->beforeEach(fn () => ...)->in('folder');
```

jordanbrauer:
Woah, that would be cool! I guess you would just configure/call that in the `Pest.php` file?

nunomaduro:
Correct. Let me know if you are up to work on it.

jordanbrauer:
> Correct. Let me know if you are up to work on it.

Sure thing, I'll take a crack at it! I'll get started ASAP , likely next weekend.

jordanbrauer:
Worked started in #282. Started with a little PoC with `beforeEach`.

nunomaduro:
Released.

PR #282, opened 2021-03-28
Reusable hooks

jordanbrauer:
| Q             | A
| ------------- | ---
| Bug fix?      | `no`
| New feature?  | `yes`
| Fixed tickets | #255 

- add global/shared hooks that are configurable in `Pest.php` similar to `groups` through the `uses()` function;
    - `beforeEach(Closure $hook)`
    - `beforeAll(Closure $hook)`
    - `afterEach(Closure $hook)`
    - `afterAll(Closure $hook)`
- add new helper function to `ChainableClosure` to create static instances;
    - `fromStatic(Closure $closure, Closure $next): Closure`
- load test structure in `run` before we calling the parent's `run` method, instead of in `createRunner`
- add tests for each new uses hook;

### DIFF
--- a/src/Console/Command.php
+++ b/src/Console/Command.php
@@ -90,8 +90,6 @@ final class Command extends BaseCommand
          */
         $this->arguments = AddsDefaults::to($this->arguments);
 
-        LoadStructure::in($this->testSuite->rootPath);
-
         $testRunner = new TestRunner($this->arguments['loader']);
         $testSuite  = $this->arguments['test'];
 
@@ -127,6 +125,8 @@ final class Command extends BaseCommand
      */
     public function run(array $argv, bool $exit = true): int
     {
+        LoadStructure::in($this->testSuite->rootPath);
+
         $result = parent::run($argv, false);
 
         /*
--- a/src/PendingObjects/UsesCall.php
+++ b/src/PendingObjects/UsesCall.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Pest\PendingObjects;
 
+use Closure;
 use Pest\Exceptions\InvalidUsesPath;
 use Pest\TestSuite;
 
@@ -12,6 +13,13 @@ use Pest\TestSuite;
  */
 final class UsesCall
 {
+    /**
+     * Contains a global before each hook closure to be executed.
+     *
+     * @var Closure
+     */
+    private $beforeEach;
+
     /**
      * Holds the class and traits.
      *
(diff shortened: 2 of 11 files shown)


nunomaduro:
Nice! Looks very promising. Keep the good work.

tests/Pest.php
```diff
@@ -1,3 +1,6 @@
 <?php
 
 uses()->group('integration')->in('Visual');
+uses()->beforeEach(function (): void {
```
nunomaduro, line 4: Since uses can also be used on the test files itself, it may be important to add a test for that too.

tests/Hooks/BeforeEachTest.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+beforeEach(function (): void {
```
nunomaduro, line 3: I think you can remove the `void` types from tests.

tests/Pest.php
jordanbrauer: What do you think is the best behaviour in this case? We could have the test's `use` call overwrite the global one in `Pest.php`, **_or;_** compose them and have the global execute first, then the local? Perhaps something else?

jordanbrauer: Here is an example, where I dump the order of execution and I the global + local `uses` get composed

image

image

nunomaduro: Compose them. First runs the global, after that runs the local one in the file.

nunomaduro: Yeah - your example is perfect.

src/Console/Command.php
```diff
@@ -127,6 +125,8 @@ protected function createRunner(): TestRunner
      */
     public function run(array $argv, bool $exit = true): int
     {
+        LoadStructure::in($this->testSuite->rootPath);
+
```
jordanbrauer, line 129: This change was necessary in order to ensure that the `Pest.php` file gets loaded before the actual test file itself (see screenshot + explanation below).

Screen Shot 2021-04-07 at 9 36 17 AM

In the first invocation, I am running with a specific directory of tests, which loads `Pest.php` _before_ the test file itself. In the second invocation, I am using the Composer `test:unit` script which does not specify a directory. When this happens, the `parent::run($argv, false)` call on line 130 will load the test file first, causing the test's `uses()` calls to execute before those in the `Pest.php` file.

src/PendingObjects/UsesCall.php
```diff
@@ -12,6 +13,20 @@
...
+     * Array indices here matter. They are mapped as follows:
+     *
+     * - `0` => `beforeAll`
+     * - `1` => `beforeEach`
+     * - `2` => `afterEach`
+     * - `3` => `afterAll`
+     *
+     * @var array<int, Closure>
+     */
+    private $hooks = [];
```
jordanbrauer, line 28: Thoughts on constants to reference these index values? I saw that plain indices are used in the `TestRepository` for the `TestRepository::use`  method, so I just stuck with that for now.

jordanbrauer (commented):
@nunomaduro this should be ready for some review whenever you have some time!

nunomaduro:
@jordanbrauer Can you pull requests docs please?

jordanbrauer:
> @jordanbrauer Can you pull requests docs please?

Sure thing, @nunomaduro ! I'll whip those up tonight – any preference of which section they could/should be added to?

nunomaduro:
You decide @jordanbrauer buddy.

Issue #194, opened 2020-09-29
Tests outside of `tests` directory won't load/run


faustbrian:
On larger projects with Laravel I keep tests in the same directory as the source but this seems to cause issues with Pest, PHPUnit works fine.

```php
<?php

use Illuminate\Foundation\Testing\RefreshDatabase;
use Tests\TestCase;

uses(TestCase::class, RefreshDatabase::class)->in(realpath(__DIR__.'/../app'));
```

This results in the following.

```
  Undefined index: file

      +2 vendor frames
  3   [internal]:0
      Composer\Autoload\ClassLoader::loadClass("App\Console\Commands\SomeCommand")

  4   [internal]:0
      spl_autoload_call("App\Console\Commands\SomeCommand")
```

Should this just work out of the box or is there something I am overlooking that needs to be adjusted?

owenvoke:
Yeah, Pest currently expects a `tests` directory and this isn't configurable at the moment. 🤔

nunomaduro:
Yeah - no plans for having this.

faustbrian:
Any specific reason for not making this configurable? It's not uncommon to keep the source and tests in the same directory when working on large projects.

nunomaduro:
I think the vast majority of Pest users do put their tests in the `tests` folder.

georgeboot:
@nunomaduro I maintain a mono repo (both php and javascript packages in same repo) and have my php tests in `/php-tests`.

Would you accept a PR for adding this as a configurable option?

nunomaduro:
Yes.

lcdss:
I'm using laravel-modules in a project and there are a test directory for each module created, so would be great if pest could run test from multiple directories as phpunit do.

faustbrian:
https://github.com/pestphp/pest/pull/283

binaryfire:
Hi. I’m also using laravel-modules (which has a separate tests folder in each module).

So just to clarify- there’s no way of using Pest in this scenario?

lcdss:
@binaryfire Unfortunately no. I switched projects and haven't worked with Pest/PHP anymore.

shijunti19:
```
packages\a031\comm
 "repositories": [
        {
            "type": "path",
            "url": "./packages/a031/comm"
        }
    ]
packages\a031\comm\tests
Now I want to execute in the root directory \Vendor bin pest -- test directory packages a031 comm tests
```
How to achieve it

chrispage1:
[...]

PR #283, opened 2021-03-29
make test directory configurable

faustbrian:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #194

> This is an implementation to make the test directory configurable as requested in https://github.com/pestphp/pest/issues/194.

It does so by introducing a new `--test-directory` argument. If this flag is set you can call `Pest\testDirectory` which will return the path value and optionally concatenate it with a filename. Calling `testDirectory('directory/File.php')` will return `tests/directory/File.php` by default.

```php
// When using an invalid directory
❯ bin/pest --test-directory=blarbs

   Pest\Exceptions\InvalidUsesPath

  The path `/Users/brian/Work/fork/blarbs` is not valid.
```

### DIFF
--- a/src/Pest.php
+++ b/src/Pest.php
@@ -11,5 +11,5 @@ function version(): string
 
 function testDirectory(string $file = ''): string
 {
-    return TestSuite::getInstance()->testPath . $file;
+    return TestSuite::getInstance()->testPath . '/' . $file;
 }
--- a/src/Pest.php
+++ b/src/Pest.php
@@ -8,3 +8,8 @@ function version(): string
 {
     return '1.0.4';
 }
+
+function testDirectory(string $file = ''): string
+{
+    return TestSuite::getInstance()->testPath . $file;
+}
--- a/bin/pest
+++ b/bin/pest
@@ -28,11 +28,12 @@ use Symfony\Component\Console\Output\OutputInterface;
     (new Provider())->register();
 
     // get $rootPath based on $autoloadPath
-    $rootPath =  dirname($autoloadPath, 2);
+    $rootPath = dirname($autoloadPath, 2);
+    $argv = new ArgvInput();
 
-    $testSuite = TestSuite::getInstance($rootPath);
+    $testSuite = TestSuite::getInstance($rootPath, $argv->getParameterOption('--test-directory', 'tests'));
 
-    $isDecorated = (new ArgvInput())->getParameterOption('--colors', 'always') !== 'never';
+    $isDecorated = $argv->getParameterOption('--colors', 'always') !== 'never';
     $output = new ConsoleOutput(ConsoleOutput::VERBOSITY_NORMAL, $isDecorated);
 
     $container = Container::getInstance();
@@ -40,6 +41,15 @@ use Symfony\Component\Console\Output\OutputInterface;
     $container->add(OutputInterface::class, $output);
 
     ValidatesEnvironment::in($testSuite);
+
+    // lets remove any arguments that PHPUnit does not understand
+    if ($argv->hasParameterOption('--test-directory')) {
+        foreach ($_SERVER['argv'] as $key => $value) {
+            if (strpos($value, '--test-directory') !== false) {
+                unset($_SERVER['argv'][$key]);
+            }
+        }
+    }
 
     exit($container->get(Command::class)->run($_SERVER['argv']));
 })();
--- a/src/Actions/LoadStructure.php
+++ b/src/Actions/LoadStructure.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Pest\Actions;
 
 use Pest\Support\Str;
+use function Pest\testDirectory;
 use PHPUnit\TextUI\Configuration\Configuration;
 use PHPUnit\Util\FileLoader;
 use RecursiveDirectoryIterator;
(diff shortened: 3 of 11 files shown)


nunomaduro (commented):
This looks good to me. Just wondering if this could be done at Plugin level? https://github.com/pestphp/pest-plugin-coverage/blob/50260c8671377a5082a06b9e228a984c13ed199c/src/Plugin.php#L76

faustbrian:
I tried that initially but there was one place that would result in issues because the custom path would be available too late, don't remember right now which. I'll take another look tomorrow at doing it that way.

faustbrian:
Just had a quick look and I think it was this line https://github.com/pestphp/pest/blob/master/src/Plugin.php#L25 that is causing issues with doing it inside of plugins because plugins require this method but this method already needs to know about the path before any plugins do so it'll fail at that point.

nunomaduro:
@faustbrian This is ready to be merged?

faustbrian:
Should be if you have no further requests for changes.

src/Pest.php
```diff
@@ -8,3 +8,8 @@ function version(): string
 {
     return '1.0.4';
 }
+
+function testDirectory(string $file = ''): string
+{
+    return $_ENV['PEST_TEST_DIRECTORY'] . $file;
```
nunomaduro, line 14: Replace this by `$_SERVER`.

nunomaduro (changes requested):
@olivernybroe Plugin / IDE wise, any changes required?

src/Pest.php
faustbrian: Just this occurrence or all `$_ENV` usages in this PR?

nunomaduro: Can you actually use the `TestSuite::getInstance()->setTestDirectory()` instead of using ENVs?

faustbrian: Let me test but if I remember right it wasn't possible because the test path had to be set before the instance was created.

faustbrian: Yeah that doesn't work because https://github.com/pestphp/pest/blob/master/src/TestSuite.php#L98-L100 ends up executing some things that already need access to the test path.

faustbrian: An alternative could be this change.

```php
$testSuite = TestSuite::getInstance($rootPath, (new ArgvInput())->getParameterOption('--test-directory', 'tests'));

public static function getInstance(string $rootPath = null, string $testPath = null): TestSuite
{
    if (is_string($rootPath)) {
        self::$instance = new TestSuite($rootPath);

        if (is_string($testPath)) {
            self::$instance->testPath = $testPath;
        }

        foreach (Plugin::$callables as $callable) {
            $callable();
        }

        return self::$instance;
    }

    if (self::$instance === null) {
        throw new InvalidPestCommand();
    }

    return self::$instance;
}
```

faustbrian: Alternative https://github.com/faustbrian/pest/commit/c8a3e7e1f4747f838292b75e44be8ac33892f2ff but results in an odd issue I need to check if this is preferred.

```
❯ bin/pest --test-directory=tests
PHPUnit 9.5.3 by Sebastian Bergmann and contributors.

Unknown option "--test-directory"
```

olivernybroe:
Hmm, by a quick look I don't think it will have an impact on the plugin. 
As far as I remember, we don't depend on the naming of the directory, but if it is marked as a test folder. 

The only thing is that this is of course not configurable via the plugin, so if we want the user to be able to configure this, we need to add a new feature. (if they don't use the env parameter mode)

faustbrian:
Pushed the changes from https://github.com/pestphp/pest/pull/283#discussion_r625011768 but will need to check tomorrow what causes the `Unknown option "--test-directory"` issue once it tries to bootstrap plugins after it has registered the path. Most likely PHPUnit not liking the parameter and it somehow being passed down to it.

faustbrian:
@nunomaduro one issue with this implementation which might be worth considering is that there are a few places that create a `TestSuite` instance outside of the CLI so they don't have access to the `--test-directory` flag while an environment variable would be accessible to them.

nunomaduro:
@faustbrian That's fine. Can you make a pull request to docs?

faustbrian:
Will submit a PR for documentation tomorrow. Just pushed a fix for the `Unknown option` parameter since PHPUnit can't handle that so we have to remove it before passing `argv` to it.

faustbrian:
Updated the PR description and submitted the docs PR https://github.com/pestphp/docs/pull/41/files

owenvoke (approved):

src/Pest.php
```diff
@@ -8,3 +8,8 @@ function version(): string
 {
     return '1.0.4';
 }
+
+function testDirectory(string $file = ''): string
+{
+    return TestSuite::getInstance()->testPath . $file;
```
nunomaduro, line 14: It is not missing a DIRECTORY_SEPARATOR between those two?

nunomaduro (changes requested):

src/Pest.php
owenvoke: Looks like it, should be able to just use `/` as that is platform agnostic and cleaner.

faustbrian: Don't remember running into any issues with this but will add it.

owenvoke: It should have caused an issue with `testDirectory('Datasets/%s.php')`, as if you used `--test-directory=test` it would return `testDatasets/%s.php` 🤔

nunomaduro (approved):
@owenvoke Can you test this pull request locally?

owenvoke:
This seems to be working well for me, I was having issues originally (but they were due to symlinking when requiring the package), and then further issues (but that was as my `phpunit.xml` testsuites weren't up to date).

PR #288, opened 2021-04-10
Ignore the absence of the tests folder

### LINKED ISSUE #194: Tests outside of `tests` directory won't load/run

gregorip02:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | Fix #194 

This PR makes it less strict to check for the existence of a `tests` folder in the main directory of the project. It can be useful for projects that prefer to test in some other weird folder. For example.

```sh
# A structure of a scripting project only

├── app
│   │── fibonacci.php
│   └── fibonacci_test.php
├── composer.json
├── composer.lock
└── phpunit.xml
```

```xml
<!-- phpunit.xml -->

<testsuite name="Test Suite">
    <directory suffix="_test.php">./app</directory>
</testsuite>
```

Note that my tests are simple files with a suffix `_test.php` in a folder other than `tests`, so with the current version of pest, if I run `vendor/bin/pest` it will fail with an exception of type `InvalidUsesPath`.

Also contributing to #283


nunomaduro:
Lets wait to have  #283 merged first.

nunomaduro:
@gregorip02 Is this ready?

gregorip02:
> @gregorip02 Is this ready?

Yep, it's ready.

PR #289, opened 2021-04-10
feat: adds built-in mocking

nunomaduro:
**Note: Fell free to ask me if you would like to keep working on this pull request. This is just a POC, that I've done in 10 minutes. I may have no time to work on this in the future.**

This pull request adds a POC about what could eventually be built-in mocks in Pest. At this time, the code doesn't do much. But the goal is to provide the following API:

```php
it('can mock methods', function () {
    $mock = mock(Http::class)->expect(
        get: 'foo',
    );

    expect($mock->get())->toBe('foo');
});;
```

Of course, this is multiple questions here: like how to provide and method expectation without return values, how to define the number of times a method expectation should happen, etc.

### DIFF
--- a/src/Exceptions/MissingDependency.php
+++ b/src/Exceptions/MissingDependency.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pest\Exceptions;
+
+use InvalidArgumentException;
+use NunoMaduro\Collision\Contracts\RenderlessEditor;
+use NunoMaduro\Collision\Contracts\RenderlessTrace;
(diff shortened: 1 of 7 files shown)


nunomaduro:
@owenvoke Any way we can avoid this test on the CI for PHP versions that are not PHP 8?

owenvoke:
Hmm, I'm not sure. 🤔 I think we'd either have to make it a separate test suite (and then just run that suite on PHP 8), or... Not sure.

I think the issue is that the file is being parsed, so we can't even use `->skip(version_compare('8', PHP_VERSION, '>')` or something.

Issue #238, opened 2020-12-22
Is it necessary to display the dataset it self?


Ilyes512:
So I going to use this as an example:

```php
<?php

declare(strict_types=1);

it('uses data from a dataset', function ($name) {
    expect($name)->toBeString();
})->with([
    'set1' => ['name' => 'hello'],
    'set2' => ['name' => 'bye'],
    'set3' => ['name' => ['oeps' => 'a array']],
]);
```

When you run this `./vendor/bin/pest` without any arguments... I get the following output:

```
> pest 'tests/FoobarTest.php'

   FAIL  Tests\FoobarTest
  ✓ it uses data from a dataset with data set "set1" ('hello')
  ✓ it uses data from a dataset with data set "set2" ('bye')
  ⨯ it uses data from a dataset with data set "set3" (array('a array'))
```

For simple datasets this might not be a problem, but with a larger set it gets rather messy really quick ie

`✓ it accepts eloquent models with data set "with_keyname_id" (array(1, 'a', 'b'), 'id', 'int', true, array('1', array('a', 'b')))`

I also notice that the message is malformed when setting `testdox="true"` in phpunit.xml:

```
PHPUnit 9.5.0 by Sebastian Bergmann and contributors.

Foobar (P\Tests\Foobar)
 ✔ It uses data from a dataset with data set "set 1" ('hello') with 
 ✔ It uses data from a dataset with data set "set 2" ('bye') with 
 ```
 
 Notice that it say `with` twice.

**Comment by nunomaduro** [comment]
You can do something like this:

```php
foreach($dataset as $data) {
    test('...', ...)
};
```

Ilyes512:
Hmm, I was hoping for another way of "fixing" this. I found the line responsible for outputting it: https://github.com/pestphp/pest/blob/master/src/Datasets.php#L116 But no easy way of overwriting it... do I really need to fork it :/

Anyway, thanks for you time.

nunomaduro:
You can propose something about this. 👍🏻

Ilyes512:
Well the easiest would be to update `getDataSetDescription` to something like:

```php
   /**
     * @param int|string        $key
     * @param array<int, mixed> $data
     */
    private static function getDataSetDescription($key, array $data): string
    {
        $nameInsert = is_string($key) ? \sprintf('data set "%s" ', $key) : '';

        return \sprintf(' with %s', $nameInsert);
    }
    ```

[...]

PR #302, opened 2021-05-23
hides dataset values if they have a name

fabio-ivona:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #238 

**This PR implements the solution proposed by @Ilyes512, all the credits goes to him**

As described in #238 , for complex datasets, for the sake of readability it should be better to display the dataset name over the actual values, in order to avoid some unreadable log:

image

The issue was closed even if after a while the solution was accepted as fine by @owenvoke (https://github.com/pestphp/pest/issues/238#issuecomment-750400639) and marked as an enhancement.

I think the closed status could have caused the enhancement to be "forgotten", so I think it could be a nice idea to bring it back up.

@Ilyes512 solution is simple: showing the dataset name if it is present, otherwise showing the actual dataset values.

this is done by changing the `Dataset::getDataSetDescription()` static method from:

```php
private static function getDataSetDescription($key, array $data): string    {        
    $exporter = new Exporter();

    $nameInsert = is_string($key) ? \sprintf('data set "%s" ', $key) : '';

    return \sprintf(' with %s(%s)', $nameInsert, $exporter-&gt;shortenedRecursiveExport($data));    
}
```

to

```php
private static function getDataSetDescription($key, array $data): string
{
    $exporter = new Exporter();

    if(is_int($key)){
        return \sprintf(' with (%s)', $exporter->shortenedRecursiveExport($data));
    }

    return \sprintf(' with data set "%s', $key);
}
```

hope to have done an helping action by bringing this proposal back to life. If not, please forgive me

### DIFF
--- a/src/Datasets.php
+++ b/src/Datasets.php
@@ -111,8 +111,10 @@ final class Datasets
     {
         $exporter = new Exporter();
 
-        $nameInsert = is_string($key) ? \sprintf('data set "%s" ', $key) : '';
+        if(is_int($key)){
+            return \sprintf(' with (%s)', $exporter->shortenedRecursiveExport($data));
+        }
 
-        return \sprintf(' with %s(%s)', $nameInsert, $exporter->shortenedRecursiveExport($data));
+        return \sprintf(' with data set "%s"', $key);
     }
 }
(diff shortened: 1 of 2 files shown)


nunomaduro:
Thank you.

PR #305, opened 2021-06-02
Ensure that generated tests name has the Test suffix

cerbero90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

This PR ensures that tests generated via `php artisan pest:test` follow the standard nomenclature and have the `Test` suffix in their name.

### DIFF
--- a/src/Laravel/Commands/PestTestCommand.php
+++ b/src/Laravel/Commands/PestTestCommand.php
@@ -39,7 +39,7 @@ final class PestTestCommand extends Command
         TestSuite::getInstance(base_path(), $this->option('test-directory'));
 
         /** @var string $name */
-        $name = $this->argument('name');
+        $name = Str::finish($this->argument('name'), 'Test');
 
         $type = ((bool) $this->option('unit')) ? 'Unit' : (((bool) $this->option('dusk')) ? 'Browser' : 'Feature');
 
--- a/src/Support/Str.php
+++ b/src/Support/Str.php
@@ -48,4 +48,14 @@ final class Str
 
         return substr($target, -$length) === $search;
     }
+
+    /**
+     * Ensures that `$target` ends with a single instance of the given `$cap`.
+     */
+    public static function finish($target, $cap)
+    {
+        $quoted = preg_quote($cap, '/');
+
+        return preg_replace('/(?:'.$quoted.')+$/u', '', $target).$cap;
+    }
 }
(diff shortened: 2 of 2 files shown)


src/Support/Str.php
```diff
@@ -48,4 +48,14 @@ public static function endsWith(string $target, string $search): bool
 
         return substr($target, -$length) === $search;
     }
+
+    /**
+     * Ensures that `$target` ends with a single instance of the given `$cap`.
+     */
+    public static function finish($target, $cap)
```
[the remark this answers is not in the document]
cerbero90: Good point, I missed that :)

src/Laravel/Commands/PestTestCommand.php
```diff
@@ -39,7 +39,7 @@ public function handle(): void
         TestSuite::getInstance(base_path(), $this->option('test-directory'));
 
         /** @var string $name */
-        $name = $this->argument('name');
+        $name = Str::finish($this->argument('name'), 'Test');
```
nunomaduro, line 42: Laravel has this feature too? On the `make:test` command?

nunomaduro (changes requested):

src/Laravel/Commands/PestTestCommand.php
cerbero90: Looks like Laravel doesn't have this feature either. It might be useful to have it as the default PHPUnit configuration doesn't run tests that don't have the `Test.php` suffix. When generating a new test, a developer may expect it to be loaded automatically and could be puzzling if it doesn't

owenvoke: Doesn't look like it actually. 🤔
https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Console/TestMakeCommand.php#L65

nunomaduro: I actually remove that "Test.php" suffix in my apps.

owenvoke: 😮 🤯

cerbero90: @nunomaduro mmm I see, it's probably better to simply respect the name that developers give to their tests as this PR wouldn't cover all the use-cases 👍

nunomaduro: Thank you for the suggestion tho.

Issue #307, opened 2021-06-02
Can not load plugins when project path contains an `@`


titouanmathis:
The explode made in [src/Repositories/TestRepository.php#L53](https://github.com/pestphp/pest/blob/master/src/Repositories/TestRepository.php#L53) based on the `@` character prevent usage of plugins when there already is an `@` in the folder absolute path.

For example, I am getting the following error when using the snapshot plugin in a project located in `/Users/<username>/Dev/@namespace/project`:

```
  Call to undefined method PHPUnit\Framework\TestCase::assertMatchesSnapshot()

  at vendor/spatie/pest-plugin-snapshots/src/Functions.php:14
     10▕ Plugin::uses(MatchesSnapshots::class);
     11▕ 
     12▕ function assertMatchesSnapshot($actual, Driver $driver = null): void
     13▕ {
  ➜  14▕     test()->assertMatchesSnapshot(...func_get_args());
     15▕ }
     16▕ 
     17▕ function assertMatchesFileHashSnapshot(string $filePath): void
     18▕ {
```

I was able to quickly fix this by changing the `@` use as a separator in the test keys to a more specific value `@@@@`. I can open a PR with my fix if needed.

PR #308, opened 2021-06-02
Fix/test key separator

titouanmathis:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #307

This PR introduces a more specific separator for the tests keys to avoid matching a potential folder name. The separator is changed from `@` to `@@@@` as it has more chance to not match anything in the test file absolute path. Maybe this value could be more complex?

### DIFF
--- a/src/Repositories/TestRepository.php
+++ b/src/Repositories/TestRepository.php
@@ -20,6 +20,11 @@ use PHPUnit\Framework\TestCase;
 final class TestRepository
 {
     /**
+     * @var string
+     */
+    private const SEPARATOR = '>>>';
+
+    /**
      * @var array<string, TestCaseFactory>
      */
     private $state = [];
@@ -50,7 +55,7 @@ final class TestRepository
             [$classOrTraits, $groups, $hooks] = $uses;
 
             $setClassName = function (TestCaseFactory $testCase, string $key) use ($path, $classOrTraits, $groups, $startsWith, $hooks): void {
-                [$filename] = explode('@', $key);
+                [$filename] = explode(self::SEPARATOR, $key);
 
                 if ((!is_dir($path) && $filename === $path) || (is_dir($path) && $startsWith($filename, $path))) {
                     foreach ($classOrTraits as $class) { /** @var string $class */
@@ -131,10 +136,10 @@ final class TestRepository
             throw ShouldNotHappen::fromMessage('Trying to create a test without description.');
         }
 
-        if (array_key_exists(sprintf('%s@%s', $test->filename, $test->description), $this->state)) {
+        if (array_key_exists(sprintf('%s%s%s', $test->filename, self::SEPARATOR, $test->description), $this->state)) {
             throw new TestAlreadyExist($test->filename, $test->description);
         }
 
-        $this->state[sprintf('%s@%s', $test->filename, $test->description)] = $test;
+        $this->state[sprintf('%s%s%s', $test->filename, self::SEPARATOR, $test->description)] = $test;
     }
 }
--- a/tests/.snapshots/success.txt
+++ b/tests/.snapshots/success.txt
@@ -122,6 +122,10 @@
    PASS  Tests\PHPUnit\CustomAffixes\AdditionalFileExtensionspec
   ✓ it runs file names like `AdditionalFileExtension.spec.php`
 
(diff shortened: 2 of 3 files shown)


owenvoke:
Could probably do something like `@@pest@@` as the separator. Although, it's unlikely `@@@@` would be taken anyway. Interesting use of `@` in your directories, not seen that before. 👍🏻

lukeraymonddowning:
I agree with @owenvoke that if we're making it specific, we might as well make it *really* specific.

lukeraymonddowning:
@nunomaduro @owenvoke 

This raises another interesting point.

Screenshot 2021-06-03 at 09 22 07

Note that the runner doesn't include the @ symbol. This makes sense, because it is removed in the `TestCaseFactory` on line 188 so that it can be a valid namespace. Do we include an 'escaped' replacement value for certain valid path symbols, such as '@' to 'AtSymbol'?

lukeraymonddowning:
For the separator, how about using `>>>`? `>` is a reserved symbol on Windows computers, so it is pretty much guaranteed that there will not be a directory including that symbol, but it also still conveys the idea of relationship between the test case and the individual test.

titouanmathis:
`>>>` should work and its connotation of the relationship is definitely a plus. I tested a line break as a separator, it seems to work as well, but I am not sure of its compatibility. 

Let me know what you prefer, I will update my PR.

tests/PHPUnit/CustomAffixes/FolderWithMultiple@@@@/ExampleTest.php
```diff
@@ -0,0 +1,17 @@
...
+{
+    public function assertTrueIsTrue()
+    {
+        $this->assertTrue(true);
+    }
+}
+
+uses(MyFailingCustomClassTest::class);
+
+test('custom traits can be used', function () {
```
lukeraymonddowning, line 15: This TestCase is no longer needed and will actually fail completely after the change to `>>>`

lukeraymonddowning (changes requested):

tests/PHPUnit/CustomAffixes/FolderWithMultiple@@@@/ExampleTest.php
titouanmathis: I just updated the tests to include `>>>` in the folder name, as it seems to work on macOS. I think that keeping the test can help people having the same issue when they have some strange folder names as I had.

lukeraymonddowning: Out of interest, what would happen if you attempted to pull this repo on a Windows machine?

owenvoke:
Using `>>>` sounds fine to me. 👍🏻

lukeraymonddowning:
@owenvoke I don't think it needs that, we just don't need the test class with >>> in the title. No developer can ever use > in a directory if they ever plan on supporting windows platforms, which is why I suggested it. That makes the test case unnecessary. Hope that makes some sort of sense.

owenvoke:
@lukeraymonddowning I guess that could be removed, same with the one ending in `>`, just wasn't sure if we needed at least one test with the separator in. 🤔 Probably not, as you said.

titouanmathis:
I removed the tests, this PR will keep the history of the bug in case anyone has some `>>>` in their folder path.

lukeraymonddowning:
Lol, sorry @titouanmathis

It would make sense to me to keep the test, but have the `@` symbol in the directory, which was the original issue. *But* there is no need to have a test case with the `>` symbol in a directory because if they have that in their project, Pest is going to be the least of their problems.

titouanmathis:
Haha, my bad, I will re-add the first test 👍

lukeraymonddowning (approved):

nunomaduro:
@lukeraymonddowning Tests are not passing. But once tests are passing, feel free to merge and to release.

A release requires changes on the changelog + https://github.com/pestphp/pest/commit/9133b88d6550023207fef892bef5f38dc07e6b30.

titouanmathis:
I added the output of the 2 new tests to the `success.txt` snapshot, the integration tests should pass now.

PR #312, opened 2021-06-09
feat: add support for using Finder objects for `uses()`

owenvoke:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #... 

This came up [on Discord](https://discord.com/channels/705126275232038944/710642595705126923/851535915016519740) where it was asked whether it was possible to provide something such as globs (e.g. `**/Feature`) to the `uses()->in()` call. This will basically throw a `ShouldNotHappen` exception if the `Finder` class doesn't exist, and the usual `uses()->in()` call can be used as normal.

For example, here's how you could specify all `Feature` subdirectories to use a specific testcase class.

```php
use Symfony\Component\Finder\Finder;

uses(Tests\TestCase::class)->inFinder(
    Finder::create()
        ->name('Feature')
        ->directories()
        ->in(__DIR__)
);
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,8 @@
         "illuminate/console": "^8.47.0",
         "illuminate/support": "^8.47.0",
         "laravel/dusk": "^6.15.0",
-        "pestphp/pest-dev-tools": "dev-master"
+        "pestphp/pest-dev-tools": "dev-master",
+        "symfony/finder": "^5.2"
     },
     "minimum-stability": "dev",
     "prefer-stable": true,
@@ -84,5 +85,8 @@
                 "Pest\\Laravel\\PestServiceProvider"
             ]
         }
+    },
+    "suggest": {
+        "symfony/finder": "If using `uses()->inFinder($finder)` for class or trait usage declaration. (^5.2)"
     }
 }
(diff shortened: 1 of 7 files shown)


tombenevides:
You can just set the dir name (in this case, "__DIR__") and the Finder class will keep looking recursively.

```PHP
use Symfony\Component\Finder\Finder;

uses(Tests\TestCase::class)->inFinder(
    Finder::create()
        ->name('Feature')
        ->directories()
        ->in(__DIR__)
);
```

Just to be clear, the parameter is a Finder instance for not add Symfony/Finder to Pest dependencies?

owenvoke:
Yeah, the parameter is a `Symfony\Component\Finder\Finder` instance, but I wasn't sure whether to add it as a dependency for now. It's just suggested if people want to use it.

nunomaduro:
Can’t we make the “in” method accept a string or a instance of Finder?

nunomaduro:
Also, can you add a test for this? Maybe add a new folder that extends a specific test case.

nunomaduro:
cc @owenvoke

owenvoke:
Apologies, I'd lost track of this. I'd thought about adding it directly to the `in()` call. I guess I can do that. 👍🏻

nunomaduro:
@owenvoke What's the current state of this pull request?

nunomaduro:
No plans for this, thanks!

Issue #316, opened 2021-06-14
Re-merge `pest-plugin-coverage`, `pest-plugin-expectations`, `pest-plugin-init`


nunomaduro:
I don't think we've made a good decision on splitting `pest-plugin-coverage`, `pest-plugin-expectations`, `pest-plugin-init` to its own package. The goal of this issue, is merge the code again in this repository.

owenvoke:
I kind of think `expectations` is quite nice as a separate plugin. That means you can use it with PHPUnit as well if you wanted. The others I think would be good to merge. 👍🏻

I'm happy to help merge all 3 though.

nunomaduro:
@owenvoke I think is quite nice - but we don't need it personally.

owenvoke:
Cool, that's fine. 👍🏻 I've opened a PR for the init plugin part of this issue: https://github.com/pestphp/pest/pull/317

nunomaduro:
Feel free to merge everything if you are confident on those pull requests... We will run a test when everything is ready.

Ilyes512:
Wow... really don't like this move. I really liked `pestphp/pest-plugin-expectations` as a separate package. I guess I will just add `pestphp/pest` instead and only use the expectations as I do now.

(Above is just a personal "preference" :) just needed to get it out.)

PR #317, opened 2021-06-15
feat: move init command out of external plugin

owenvoke:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #... 

This moves the Pest Init plugin into Pest core.

Part of https://github.com/pestphp/pest/issues/316.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,7 +8,7 @@
 .gitattributes           export-ignore
 .gitignore               export-ignore
 phpstan.neon             export-ignore
-phpunit.xml              export-ignore
+/phpunit.xml             export-ignore
 CHANGELOG.md             export-ignore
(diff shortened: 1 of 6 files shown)


.gitattributes
```diff
@@ -8,7 +8,7 @@
 .gitattributes           export-ignore
 .gitignore               export-ignore
 phpstan.neon             export-ignore
-phpunit.xml              export-ignore
+/phpunit.xml             export-ignore
```
owenvoke, line 11: This is necessary otherwise the `phpunit.xml` stub is ignored on export. I guess we could call the stub `phpunit.xml.stub`, but that seems a bit weird when it's already in a `stubs` directory.

nunomaduro: No problem.

Issue #249, opened 2021-01-05
Clean error message when missing dataset


olivernybroe:
If you write a test and that test was supposed to have a dataset, meaning you have added arguments to the function, but haven't added the dataset yet, you get an error message looking like this
```
ArgumentCountError : Too few arguments to function P\Tests\Feature\UpgraderTest::{closure}(), 0 passed and exactly 2 expected
```
And the message is even worse if you have an `expectException` call on 
```
Failed asserting that exception of type "ArgumentCountError" matches expected exception "App\Exceptions\UpgradeInProgressException". Message was: "Too few arguments to function P\Tests\Feature\UpgraderTest::{closure}(), 0 passed and exactly 2 expected" at
```

To clean this up, I suggest we check if a test function has arguments, if it has, then a dataset is required. If there is no dataset, then show a user friendly message about a dataset is missing or something like that.

jordanbrauer:
@nunomaduro Mind if I pick this one up?

nunomaduro:
@jordanbrauer  sure.

nunomaduro:
@jordanbrauer Did you made any progress?

jordanbrauer:
@nunomaduro Oh my goodness! Yes, I've got some stuff stashed away but got side tracked with some home-related things. Will make some commits and send a PR in tonight!

jordanbrauer:
@olivernybroe and @nunomaduro , the PR branch should be good for review now, over at #321!

jordanbrauer:
@nunomaduro @olivernybroe I suppose we can close this, now that #321 was merged?

PR #321, opened 2021-06-16
Add user-friendly exception message for missing test inputs

jordanbrauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #249 

### Overview

- add a new exception type, `DatasetMissing`;
- add the new `bool $dependent` property to the test factory to indicate `depends` has been called;
    - defaults to `false`;
    - we set it to `true` after adding the `addDependencies` factory proxy message from the `TestCall::depends` method;
- make check for datasets within the `TestRepository::set` method;
    - we throw the new exception with user-friendly message when all the following are true;
        - the test has no dependencies;
        - the test has no datasets;
        - the test closure has an argument list greater than 0 (checked via reflection);
- add new `Reflection::getFunctionArguments` helper method to return a map of argument names to their types;
    - used to double for exception message formatting as well as argument list length checks;
    - handles PHP 8 union types as well;

### Examples

image

image

image

### DIFF
--- a/src/PendingObjects/TestCall.php
+++ b/src/PendingObjects/TestCall.php
@@ -97,6 +97,8 @@ final class TestCall
             ->factoryProxies
             ->add(Backtrace::file(), Backtrace::line(), 'addDependencies', [$tests]);
 
+        $this->testCaseFactory->dependent = true;
+
         return $this;
     }
 
--- a/src/PendingObjects/TestCall.php
+++ b/src/PendingObjects/TestCall.php
@@ -102,8 +102,6 @@ final class TestCall
             ->factoryProxies
             ->add(Backtrace::file(), Backtrace::line(), 'addDependencies', [$tests]);
 
-        $this->testCaseFactory->dependent = true;
-
         return $this;
     }
 
--- a/src/Repositories/TestRepository.php
+++ b/src/Repositories/TestRepository.php
@@ -5,11 +5,13 @@ declare(strict_types=1);
 namespace Pest\Repositories;
 
 use Closure;
+use Pest\Exceptions\DatasetMissing;
 use Pest\Exceptions\ShouldNotHappen;
 use Pest\Exceptions\TestAlreadyExist;
 use Pest\Exceptions\TestCaseAlreadyInUse;
 use Pest\Exceptions\TestCaseClassOrTraitNotFound;
 use Pest\Factories\TestCaseFactory;
+use Pest\Support\Reflection;
 use Pest\Support\Str;
 use Pest\TestSuite;
 use PHPUnit\Framework\TestCase;
@@ -138,6 +140,14 @@ final class TestRepository
 
         if (array_key_exists(sprintf('%s%s%s', $test->filename, self::SEPARATOR, $test->description), $this->state)) {
             throw new TestAlreadyExist($test->filename, $test->description);
+        }
+
+        if ($test->dependent === false && count($test->datasets) === 0) {
+            $arguments = Reflection::getFunctionArguments($test->test);
+
+            if (count($arguments) > 0) {
+                throw new DatasetMissing($test->filename, $test->description, $arguments);
+            }
         }
 
         $this->state[sprintf('%s%s%s', $test->filename, self::SEPARATOR, $test->description)] = $test;
--- a/src/Repositories/TestRepository.php
+++ b/src/Repositories/TestRepository.php
@@ -142,7 +142,7 @@ final class TestRepository
             throw new TestAlreadyExist($test->filename, $test->description);
         }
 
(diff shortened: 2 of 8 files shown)


tests/Unit/TestSuite.php
```diff
@@ -11,3 +12,11 @@
...
     $testSuite->tests->set(new \Pest\Factories\TestCaseFactory(__FILE__, 'foo', $test));
 });
+
+it('alerts users about tests with arguments but no input', function () {
+    $testSuite = new TestSuite(getcwd(), 'tests');
+    $test = function (int $arg) {};
+    $this->expectException(DatasetMissing::class);
+    $this->expectExceptionMessage(sprintf("A test with the description '%s' has %d argument(s) ([%s]) and no dataset(s) provided in %s", 'foo', 1, 'int $arg', __FILE__));
+    $testSuite->tests->set(new \Pest\Factories\TestCaseFactory(__FILE__, 'foo', $test));
+});
```
owenvoke, line 22: Is it worth using the `->throws()` syntax here, or will that not work for some reason? 🤔

```suggestion
    $testSuite->tests->set(new \Pest\Factories\TestCaseFactory(__FILE__, 'foo', $test));
})->throws(
    DatasetMissing::class,
    sprintf("A test with the description '%s' has %d argument(s) ([%s]) and no dataset(s) provided in %s", 'foo', 1, 'int $arg', __FILE__)
);
```

jordanbrauer: Good idea. I copied the above test down. I gave it the same treatment – seems to work just fine. 👍

src/PendingObjects/TestCall.php
```diff
@@ -97,6 +97,8 @@ public function depends(string ...$tests): TestCall
             ->factoryProxies
             ->add(Backtrace::file(), Backtrace::line(), 'addDependencies', [$tests]);
 
+        $this->testCaseFactory->dependent = true;
```
nunomaduro, line 100: Can't you count dependencies internally?

src/Repositories/TestRepository.php
```diff
@@ -140,6 +142,14 @@ public function set(TestCaseFactory $test): void
             throw new TestAlreadyExist($test->filename, $test->description);
         }
 
+        if ($test->dependent === false && count($test->datasets) === 0) {
```
nunomaduro, line 145: Can't this just be `count($test->datasets) === 0`?

src/PendingObjects/TestCall.php
jordanbrauer: I tried doing that, but was not able to find where I can gain access to the `[$tests]` lists, the same we we have `$test->datasets`. Perhaps I missed something though. Where did you have mind?

src/Repositories/TestRepository.php
jordanbrauer: Unfortunately not, because tests with dependencies can/will have function arguments, but no dataset so they will fail the test. I.e., we test for datasets & dependencies; if neither exist, then you have (presumably) not supplied any arguments to the test.

lukeraymonddowning: Thoughts on moving this `if` logic into the `TestCaseFactory`? Something like:

```php
if ($test->doesNotSupportArguments())
```

jordanbrauer: Yeah, I like that idea. I will refactor into this method

src/PendingObjects/TestCall.php
jordanbrauer: After some more looking, it seems that in order to count them locally, we would need to loop through the proxy message list, and check for `addDependencies` entries. If you don't think this will be an issue for performance (we make the check for every test and would have to loop over the whole message collection) we can try it, adding a method to the `HigherOrderMessageCollection` to get a count/bool?

src/Repositories/TestRepository.php
jordanbrauer: Updated @lukeraymonddowning, @nunomaduro :)

lukeraymonddowning (approved):
Nice work @jordanbrauer

src/PendingObjects/TestCall.php
nunomaduro: > adding a method to the HigherOrderMessageCollection to get a count/bool?

Looks good to me.

jordanbrauer: Nice, I'll add and update later today.

jordanbrauer: @nunomaduro Swapped out the prop set/check with the aforementioned `HigherOrderMessageCollection` method. Should be good for a review again whenever ya' get some minutes :)

PR #322, opened 2021-06-16
Reimplements the new `toHaveArray` features.

lukeraymonddowning:
This is a copy of https://github.com/pestphp/pest-plugin-expectations/pull/15 after plugins have been moved into Pest core.

One change I made was to relocate the `Arr` class to `Support\Arr` so that it can be used by other Pest features in the future if desired.


lukeraymonddowning:
Documentation added: https://github.com/pestphp/docs/pull/63

PR #325, opened 2021-06-18
Method json() to parse JSON strings

dansysanalyst:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #... 

Allows to parse a JSON string to array and access it when using Higher Order expectations.

```php

//Propose
expect('{"name":"Nuno"}')->json()->name->toBe('Nuno');

//Instead of 
expect(json_decode('{"name":"Nuno"}'), true)->name->toBe('Nuno');
```

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -61,7 +61,7 @@ final class Expectation
     }
 
     /**
-     * Parses Json String to Array.
+     * Creates a new expectation with the decoded JSON value.
      */
     public function json(): Expectation
     {
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -65,10 +65,7 @@ final class Expectation
      */
     public function json(): Expectation
     {
-        Assert::assertIsString($this->value);
-        Assert::assertJson($this->value);
-
-        return new self(json_decode($this->value, true));
+        return $this->toBeJson()->and(json_decode($this->value, true));
     }
 
     /**
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -61,6 +61,17 @@ final class Expectation
     }
 
     /**
+     * Parses Json String to Array.
+     */
+    public function json(): Expectation
+    {
+        Assert::assertIsString($this->value);
+        Assert::assertJson($this->value);
+
+        return new self(json_decode($this->value, true));
+    }
+
+    /**
      * Dump the expectation value and end the script.
      *
      * @param mixed $arguments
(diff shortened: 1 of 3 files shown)


src/Expectation.php
```diff
@@ -60,6 +60,17 @@ public function and($value): Expectation
         return new self($value);
     }
 
+    /**
+     * Parses Json String to Array.
+     */
+    public function json(): Expectation
+    {
+        Assert::assertIsString($this->value);
```
lukeraymonddowning, line 68: Could this be refactored to `return $this->toBeJson()->and(json_decode($this->value, true));`?

src/Expectation.php
```diff
@@ -60,6 +60,17 @@ public function and($value): Expectation
         return new self($value);
     }
 
+    /**
+     * Parses Json String to Array.
```
nunomaduro, line 64: ` Return the decoded JSON value.`

src/Expectation.php
nunomaduro, line 68: `$this->toBeString();`

src/Expectation.php
```diff
@@ -60,6 +60,17 @@ public function and($value): Expectation
...
+     */
+    public function json(): Expectation
+    {
+        Assert::assertIsString($this->value);
+        Assert::assertJson($this->value);
```
nunomaduro, line 69: `$this->toBeJson();`

src/Expectation.php
nunomaduro: ahah.

lukeraymonddowning: Seen as `toBeJson()` already does the string check, we don't need this check.

lukeraymonddowning:
Also, looks like you might need to run `composer update:snapshots`

owenvoke:
This is really nice. Good work!

owenvoke (approved):

nunomaduro:
This can be tagged, if someone have a free time todo it.

owenvoke:
@lukeraymonddowning, don't forget to follow the release guide when tagging. 👍🏻

https://github.com/pestphp/pest/blob/master/RELEASE.md

Need to update the changelog, and probably best to follow the commit message template for consistency. 👍🏻

lukeraymonddowning:
Thanks @owenvoke, I'll go through next time. Appreciate you pointing it out.

PR #330, opened 2021-06-24
Allows `and` in Higher Order Expectations.

lukeraymonddowning:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes

This is kind of a bug fix, but I prefer to think of it as a new feature 😆 

Previously, calling the `and` method in a Higher Order expectation would fail because the original Higher Order expectation was still the target. This PR now allows for the following:

```php
it('can start a new higher order expectation using the and syntax', function () {
    expect(new HasMethodsAndProperties())
        ->toBeInstanceOf(HasMethodsAndProperties::class)
        ->meta->toBeArray
        ->and(['foo' => 'bar'])
        ->toBeArray()
        ->foo->toEqual('bar');
});
```

### DIFF
--- a/src/HigherOrderExpectation.php
+++ b/src/HigherOrderExpectation.php
@@ -64,6 +64,16 @@ final class HigherOrderExpectation
     }
 
     /**
+     * Creates a new expectation.
+     *
+     * @param mixed $value
+     */
+    public function and($value): HigherOrderExpectation
+    {
+        return new self($this->expect($value), $value);
+    }
+
+    /**
      * Dynamically calls methods on the class with the given arguments.
      *
      * @param array<int, mixed> $arguments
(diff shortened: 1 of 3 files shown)


src/HigherOrderExpectation.php
```diff
@@ -63,6 +63,16 @@ public function not(): HigherOrderExpectation
...
     }
 
+    /**
+     * Creates a new expectation.
+     *
+     * @param mixed $value
+     */
+    public function and($value): HigherOrderExpectation
+    {
+        return new self($this->expect($value), $value);
```
nunomaduro, line 73: I would expect this to be just `return new Expectation($value);`. Can you explain me your reasoning?

nunomaduro (changes requested):

src/HigherOrderExpectation.php
lukeraymonddowning: I just looooove overcomplicating things. Good catch. Updated. Also allows us to be more specific with Generics.

PR #331, opened 2021-06-24
Adds a new `tap` method to Higher Order tests

lukeraymonddowning:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

Let's face it, Higher Order tests are the bomb. I love them. But occasionally, I'll need to nip into the test case for something or call a method that can't be called immediately (like the `route` method in Laravel).

This PR adds a new `tap` method, which provides this functionality:

```php
it('can tap into the test')
    ->expect('foo')->toBeString()->toBe('foo')
    ->tap(function () { expect($this)->toBeInstanceOf(TestCase::class); })
    ->and('hello world')->toBeString();
```

If you return from the `tap` method, you can actually springboard off of that for the rest of the higher order test:

```php
it('can return from a tap')
    ->tap(fn () => expect(route('welcome')))
    ->toStartWith('https://');
```

Excited to hear your thoughts!

### DIFF
--- a/src/PendingObjects/TestCall.php
+++ b/src/PendingObjects/TestCall.php
@@ -11,7 +11,8 @@ use Pest\Support\NullClosure;
 use Pest\TestSuite;
 use SebastianBergmann\Exporter\Exporter;
 
-/**
+/*
+ * @mixin \Pest\Support\HigherOrderCallables
  * @method \Pest\Expectations\Expectation expect(mixed $value)
  *
  * @internal
--- a/src/PendingObjects/TestCall.php
+++ b/src/PendingObjects/TestCall.php
@@ -7,15 +7,15 @@ namespace Pest\PendingObjects;
 use Closure;
 use Pest\Factories\TestCaseFactory;
 use Pest\Support\Backtrace;
+use Pest\Support\HigherOrderCallables;
 use Pest\Support\NullClosure;
 use Pest\TestSuite;
 use SebastianBergmann\Exporter\Exporter;
 
-/*
- * @mixin \Pest\Support\HigherOrderCallables
- * @method \Pest\Expectations\Expectation expect(mixed $value)
- *
+/**
  * @internal
+ *
+ * @mixin HigherOrderCallables
  */
 final class TestCall
 {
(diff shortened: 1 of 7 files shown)


nunomaduro:
Can you give me real usages of this `tap` thing?

lukeraymonddowning:
@nunomaduro of course! One of my main ones is in Laravel. If you want to call a facade, you'd have to write a full PhpUnit test case method, which is super overkill. Instead, you could just do this:

```php
it("has the correct default config")
    ->tap(fn() => expectConfig::get('my-package.cuz')))
    ->toBeString()->toEqual("some config value");
```

Basically, it's a little shortcut for those times where writing a method in the TestCase would be overkill, but what you're trying to access would throw an exception at runtime. Hope that makes some sort of sense?

nunomaduro:
It makes sense, yet "tap" doesn't mean that. Usually the object returned by the "tap" gets ignored.

lukeraymonddowning:
The other method name I thought might work is `defer`?

"Defer the value of this callable until it can be resolved in the test case?"

lukeraymonddowning:
Another option is `with`

```php
it("has the correct default config")
    ->with(fn() => expect(Config::get('my-package.some-option')))
    ->toBeString()->toEqual("some config value");
```

cerbero90:
Thanks @lukeraymonddowning for this great feature ❤️ What do you guys think of `bind()`, `scope()`, `scoped()` or `proxy()`?

nunomaduro:
@lukeraymonddowning @cerbero90 I think the `expect` method, when called from high order tests, should resolve the given callable.

```php
it("has the correct default config")
    ->expect(fn () => Config::get('my-package.some-option'))
    ->toBeString()->toEqual("some config value");
```

lukeraymonddowning:
@nunomaduro that's a good idea. Perhaps we could have have the `tap` method return the test case then? Which would then truly be a `tap`? That would then solve both issues.

nunomaduro:
Ok.

lukeraymonddowning:
@nunomaduro @cerbero90 

Updated with the new functionality. Passing a callable to a Higher Order Test's `expect` or `and` method will now resolve the callable as the expectation rather than the callable itself. The `tap` method has been updated to always return the test case.

PR #338, opened 2021-07-08
Closures passed to the `skip` method are now bound to the test case

lukeraymonddowning:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

Hey y'all,

something that came up in the stream yesterday was this:

```php
public function requiresPostgreSQL(): void
{
    if (DB::getDriverName() !== 'pgsql') {
        $this->markTestSkipped('This test requires a PostgreSQL database connection');
    }
}
```

Obviously, this condition requires the test case setup method to already have been run, so you wouldn't be able to write it in Pest's `skip` method. This PR enables this functionality by moving the conditional logic into the HigherOrderMessage itself, and binding the conditional closure to the test case.

```php
it('works when using PostgreSQL', function() {
    expect(true)->toBeTrue();
})->skip(fn() => DB::getDriverName() !== 'pgsql', 'This test requires a PostgreSQL database connection');
```

This feature actually opens up the potential for other bound callable too, because now when adding to a HigherOrderMessage, you can simply use the newly built `when` method and pass a conditional; it will always be bound to the test case.

Kind Regards,
Luke

### DIFF
--- a/src/Support/HigherOrderMessage.php
+++ b/src/Support/HigherOrderMessage.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Pest\Support;
 
+use Closure;
+use const PHP_MAJOR_VERSION;
 use ReflectionClass;
 use Throwable;
 
@@ -51,6 +53,13 @@ final class HigherOrderMessage
     public $arguments;
 
     /**
+     * An optional condition that will determine if the message will be executed.
+     *
+     * @var callable|null
+     */
+    public $condition = null;
+
+    /**
      * Creates a new higher order message.
      *
      * @param array<int, mixed> $arguments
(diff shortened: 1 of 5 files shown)


src/Support/HigherOrderMessage.php
```diff
@@ -50,6 +52,13 @@ final class HigherOrderMessage
      */
     public $arguments;
 
+    /**
+     * An optional condition that will determine if the message will be executed.
+     *
+     * @var callable|null
```
nunomaduro, line 58: Can you type the format of the callable?

src/Support/HigherOrderMessage.php
```diff
@@ -88,9 +102,16 @@ public function call(object $target)
         }
     }
 
+    public function when(callable $condition): self
```
nunomaduro, line 105: Can you type the format of the callable?

nunomaduro (changes requested):

src/Support/HigherOrderMessage.php
lukeraymonddowning: Doh! Was meant to come back to this but slipped my mind. Updated 👍

src/Support/HigherOrderMessage.php
lukeraymonddowning: Updated 👍

nunomaduro (approved):

PR #339, opened 2021-07-08
Allows you to just specify an exception message when calling `throws`.

lukeraymonddowning:
This PR comes from a discussion with @nunomaduro 

It allows for a user to omit the exception type from a `throws` call and just specify an exception message instead:

```php
it('can just define the message', function () {
    throw new Exception('Something bad happened');
})->throws('Something bad happened');
```

### DIFF
--- a/src/PendingObjects/TestCall.php
+++ b/src/PendingObjects/TestCall.php
@@ -61,9 +61,13 @@ final class TestCall
      */
     public function throws(string $exceptionClass, string $exceptionMessage = null): TestCall
     {
-        $this->testCaseFactory
-            ->proxies
-            ->add(Backtrace::file(), Backtrace::line(), 'expectException', [$exceptionClass]);
+        if (class_exists($exceptionClass)) {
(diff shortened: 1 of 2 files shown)


src/PendingObjects/TestCall.php
```diff
@@ -61,9 +61,13 @@ public function __construct(TestSuite $testSuite, string $filename, string $desc
      */
     public function throws(string $exceptionClass, string $exceptionMessage = null): TestCall
```
nunomaduro, line 62: @lukeraymonddowning Can you rename `$exceptionClass` to just `$exception`?

nunomaduro (changes requested):

src/PendingObjects/TestCall.php
lukeraymonddowning: Updated

nunomaduro:
Can you pull request this stuff to the docs, merge, release, and tweet?

PR #340, opened 2021-07-09
Better tap type hinting

lukeraymonddowning:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no

Previously, calling `tap` in a higher order test would provide no IDE autocomplete suggestions afterwards. This PR resolves that. It also enforces a `TestCase` being passed to the `HigherOrderTapProxy`, because the logic dictates that it would always be a `TestCase` by the time a proxy is resolved.


olivernybroe:
<3

olivernybroe (approved):
Nice catch 🪝

PR #348, opened 2021-07-16
Fix the unit test stub

jdanino:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no

The stub for creating a new unit test through the pest:test command was creating a not working example test.

### DIFF
--- a/stubs/Unit.php
+++ b/stubs/Unit.php
@@ -1,5 +1,5 @@
 <?php
 
 test('{name}', function () {
-    $this->assertTrue(true);
+    expect(true)->toBeTrue();
 });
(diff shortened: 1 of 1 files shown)


stubs/Unit.php
```diff
@@ -1,5 +1,5 @@
 <?php
 
 test('{name}', function () {
-    assertTrue(true);
+    $this->assertTrue(true);
```
owenvoke, line 4: To be honest, I think this should use the expectation API and be:
```suggestion
    expect(true)->toBeTrue();
```

owenvoke (approved):

jdanino (commented):
👍

PR #350, opened 2021-07-16
Teamcity styling

lukeraymonddowning:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

This PR aims to add Pest styling to the PhpStorm `TeamCity` Logger. Here are some output examples:

Screenshot 2021-07-16 at 17 53 09

Screenshot 2021-07-16 at 17 53 50

Screenshot 2021-07-16 at 17 54 09

Screenshot 2021-07-16 at 17 55 19

Screenshot 2021-07-16 at 17 55 57

### DIFF
--- a/src/Logging/TeamCity.php
+++ b/src/Logging/TeamCity.php
@@ -4,8 +4,6 @@ declare(strict_types=1);
 
 namespace Pest\Logging;
 
-use ReflectionClass;
-use ReflectionProperty;
 use function getmypid;
 use Pest\Concerns\Testable;
 use function Pest\version;
@@ -17,6 +15,7 @@ use PHPUnit\Framework\TestSuite;
 use PHPUnit\Framework\Warning;
 use PHPUnit\Runner\PhptTestCase;
 use PHPUnit\TextUI\DefaultResultPrinter;
+use ReflectionProperty;
 use function round;
 use function str_replace;
 use function strlen;
@@ -309,6 +308,10 @@ final class TeamCity extends DefaultResultPrinter
 
     private function removePestFromStackTrace(Throwable $e): void
     {
+        if (!property_exists($e, 'serializableTrace')) {
+            return;
+        }
+
         $property = new ReflectionProperty($e, 'serializableTrace');
         $property->setAccessible(true);
         $trace = $property->getValue($e);
--- a/src/Logging/TeamCity.php
+++ b/src/Logging/TeamCity.php
@@ -249,9 +249,7 @@ final class TeamCity extends DefaultResultPrinter
     public function endTest(Test $test, float $time): void
     {
         if (!TeamCity::isPestTest($test)) {
-            $this->phpunitTeamCity->endTest($test, $time);
-
-            return;
+            parent::endTest($test, $time);
         }
 
         $this->printEvent('testFinished', [
--- a/src/Logging/TeamCity.php
+++ b/src/Logging/TeamCity.php
@@ -17,6 +17,7 @@ use PHPUnit\Runner\PhptTestCase;
 use PHPUnit\TextUI\DefaultResultPrinter;
 use function round;
 use function str_replace;
+use function strlen;
 use Throwable;
 
 final class TeamCity extends DefaultResultPrinter
@@ -45,11 +46,14 @@ final class TeamCity extends DefaultResultPrinter
      */
     protected $outputStack = [];
 
-    public function __construct(bool $verbose, string $colors)
+    /**
+     * @param resource|string|null $out
+     */
+    public function __construct($out, bool $verbose, string $colors)
     {
-        parent::__construct(null, $verbose, $colors, false, 80, false);
+        parent::__construct($out, $verbose, $colors, false, 80, false);
         $this->phpunitTeamCity = new \PHPUnit\Util\Log\TeamCity(
-            null,
+            $out,
             $verbose,
             $colors,
             false,
@@ -76,7 +80,7 @@ final class TeamCity extends DefaultResultPrinter
     /** @phpstan-ignore-next-line */
     public function startTestSuite(TestSuite $suite): void
     {
-        if (str_starts_with($suite->getName(), 'P\\')) {
+        if (static::isPestTestSuite($suite)) {
             $this->writeWithColor('fg-white, bold', '  ' . substr_replace($suite->getName(), '', 0, 2) . ' ');
         }
 
@@ -116,7 +120,7 @@ final class TeamCity extends DefaultResultPrinter
     {
         $suiteName = $suite->getName();
 
-        if (str_starts_with($suiteName, 'P\\')) {
+        if (static::isPestTestSuite($suite)) {
             $this->writeNewLine();
         }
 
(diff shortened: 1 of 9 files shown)


lukeraymonddowning:
Multiple test cases look like this when run

Screenshot 2021-07-16 at 17 43 23

In a similar style to Pest's console runner, I opted to defer error and failure messages until the end of the output, as you can see below.

Screenshot 2021-07-16 at 17 43 57

olivernybroe:
Looks really great! 

It's not possible for us to share the output with the normal console so we don't have to maintain the output in two places?

src/Logging/TeamCity.php
```diff
@@ -71,7 +71,7 @@ public function printResult(TestResult $result): void
     /** @phpstan-ignore-next-line */
     public function startTestSuite(TestSuite $suite): void
     {
-        if (!str_contains($suite->getName(), '.php')) {
+        if (str_starts_with($suite->getName(), 'P\\')) {
```
nunomaduro, line 79: str_starts_with is PHP 8 only?

nunomaduro (changes requested):

src/Logging/TeamCity.php
lukeraymonddowning: Lol. My bad. Will update.

lukeraymonddowning:
> Looks really great! 
> 
> It's not possible for us to share the output with the normal console so we don't have to maintain the output in two places? 

Hadn't really thought about that but will look at the possibility of unifying the two 👍

src/Logging/TeamCity.php
olivernybroe: @lukeraymonddowning would it be possible to add a test that runs the teamcity logger? :) 

It's kinda bad that our CI did not catch this.

lukeraymonddowning: This actually will succeed on <8.0 because its using a polyfill from symfony. Don't like that we're relying on that though so will alter anyway.

lukeraymonddowning:
@olivernybroe we now defer to Collision's adapter for output. However, there seems to be a little issue with test failures showing the error output:

Screenshot 2021-07-19 at 13 44 16

I imagine this is to do with test listeners and the output helpers. Any ideas?

src/Logging/TeamCity.php
```diff
@@ -27,37 +30,53 @@ final class TeamCity extends DefaultResultPrinter
...
+    private $collisionPrinter = null;
 
-    public function __construct(bool $verbose, string $colors)
+    /**
+     * @param resource|string|null $out
+     */
+    public function __construct($out, bool $verbose, string $colors)
     {
-        parent::__construct(null, $verbose, $colors, false, 80, false);
-        $this->phpunitTeamCity = new \PHPUnit\Util\Log\TeamCity(
```
olivernybroe, line 42: I might be a bit blind, but what replaced the part which did this? 🤔

tests/Visual/JUnit.php
```diff
@@ -0,0 +1,29 @@
...
+use PHPUnit\Framework\AssertionFailedError;
+use PHPUnit\Framework\TestSuite;
+use PHPUnit\Framework\Warning;
+
+beforeEach(function () {
+    file_put_contents(__DIR__ . '/junit.html', '');
+});
+
+it('is can successfully call all public methods', function () {
+    $junit = new JUnit(__DIR__ . '/junit.html');
```
olivernybroe, line 13: Really nice with the tests. 

Should we also test the output? We can do that by having the junit.html file saved and checking if the result is identical with the generated one 🤔

olivernybroe (commented):
Nice! 

I think the code looks great and really nice cleanup. I haven't tested it out locally yet, hopefully I get time to do that soon or someone else does it 👍

src/Logging/TeamCity.php
lukeraymonddowning: Basically nothing. We're firing all of the events that team city were firing manually, and now that we don't share their output at all, it became redundant. Can you see an issue with it?

tests/Visual/JUnit.php
lukeraymonddowning: @olivernybroe I've added a test in `SingleTestOrDirectory` that runs using JUnit logging and asserts it matches the given snapshot.

src/Logging/TeamCity.php
lukeraymonddowning: I've gone through and manually implemented some of the events. We have to do it manually so as not to give PHPStorm too much information (because then it writes to the console) but just enough to get it to display the correct icon. If we decorate the base `TeamCity` logger it writes out too much to the console.

lukeraymonddowning:
@nunomaduro @olivernybroe I'm pretty happy with this PR now. Could do with somebody who has PhpStorm seeing if I've missed anything. The only quirk is on skipped tests:

Screenshot 2021-07-23 at 00 10 57

In order to get the skip indicators in PhpStorm, it has to output the "Test ignored." text. I can get it to go to the bottom using `ob_start` and `ob_flush` around the event printer, but I can't get the text to disappear completely. Thoughts?

lukeraymonddowning:
Follow-up: We can pass a space for the message on skipped tests, which is cleaner, and combined with `ob_start` results in just a bunch of new lines at the bottom of the console output:

Screenshot 2021-07-23 at 00 18 49

olivernybroe:
Hey @lukeraymonddowning, there is unfortunately something wrong 😢 

When I have a failing test, it does not report the error message of it as part of the specific test case.
image
image
I have to go to all output to be able to see why it failed. However before you could see this on each specific test, which is really helpful when having multiple failing tests.

lukeraymonddowning:
Thanks @olivernybroe! That's super helpful. Do you know how we could support the full teamcity events without having it print to the console?

olivernybroe:
@lukeraymonddowning not completely no 🙁

We might be able to do a wrapper around the out object given to collision, so we can get the output from it and wrap it in a team city format? 🤔

nunomaduro:
@lukeraymonddowning @olivernybroe Once this is ready, fell free to merge and release.

src/Logging/TeamCity.php
```diff
@@ -89,11 +153,52 @@ public function startTestSuite(TestSuite $suite): void
...
+    {
+        return strncmp($suite->getName(), 'P\\', strlen('P\\')) === 0;
+    }
+
+    /**
+     * @param array<string, string|int> $params
+     */
+    private function printEvent(string $eventName, array $params = []): void
+    {
+        $this->write("\n##teamcity[{$eventName}");
```
olivernybroe, line 171:
```suggestion
        $this->write("##teamcity[{$eventName}");
```

Actually we should remove this first `\n`, we always append it, so having it both places is not needed. 

I also think we need to do this because phpstorm has the method called `serviceMessageHasNewLinePrefix` which is scheduled for removal in 2021.3.

lukeraymonddowning:
@olivernybroe I've just gone and refactored all of the work here to clean it up nicely. Ready for another review whenever you get chance.

olivernybroe (approved):
I am done testing this out now and it works really well! Amazingly done ❤️

PR #354, opened 2021-07-21
Higher order test expect escape and skip fix

lukeraymonddowning:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no

Hi guys,

## Higher Order Tests quirk fix

So something that bothered me the other day was the fact that in higher order tests, when using expect chains, I got stuck in the chain and couldn't back out again. This was because expect thought we were working with higher order expectations and so tried to call the method/property on that, but no fallback was in place for the use-case where you actually want to move onto another method call.

```php
// This would previously fail, because `getNumAssertions` isn't available to the string 'foo'
it('can call test methods after an expect chain')
    ->expect('foo')
    ->toBeString()->toBe('foo')
    ->getNumAssertions();
```

This PR fixes that issue by catching the error exception and trying again on the test object. It allows you to chain global functions and test case methods onto the end of an exception chain in higher order tests.

## Skip fix

Something I noticed whilst working on this PR was that the skip methods weren't working correctly, specifically this:

```php
it('do not skips')
    ->skip(false, 'here')
    ->assertTrue(true);
```

The test suite was exiting with an unknown error. After some digging, it seems to be related to wrapping the passed value in a callable, which doesn't seem to retain its original state when fired in the proxy calls. I've also moved the skip method from a `chain` to a `proxy`, which fixes this test which was passing before but only because it had been written incorrectly:

```php
it('skips when skip after assertion')
    ->assertTrue(false)
    ->skip();
```

Let me know what you think 👍

Kind Regards,
Luke

### DIFF
--- a/src/PendingObjects/TestCall.php
+++ b/src/PendingObjects/TestCall.php
@@ -8,7 +8,6 @@ use Closure;
 use Pest\Factories\TestCaseFactory;
 use Pest\Support\Backtrace;
 use Pest\Support\HigherOrderCallables;
-use Pest\Support\NullClosure;
 use Pest\TestSuite;
 use SebastianBergmann\Exporter\Exporter;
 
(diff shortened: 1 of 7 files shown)


nunomaduro:
@lukeraymonddowning I am unable to understand this pull request. If you are very confident, and can merge and release.

Issue #351, opened 2021-07-16
'beforeAll' hook on Pest.php is executed more than once for a single file


fnzr:
Hi,
I'm trying to register a global hook that will be executed once before all tests, but when registering it on the Pest.php file, it is called once per test instead.
Reproducible from this repository: https://github.com/fnzr/pest-sample-temp
What I've done:
1) `composer require pestphp/pest --dev --with-all-dependencies`
2) `./vendor/bin/pest --init`
3) Created the sample tests:
```php 
//Pest.php
uses()->beforeAll(function() {
    echo 'Global: beforeAll' . PHP_EOL;
})->in('subdir');
```
```php
//subdir/MyTest.php
beforeAll(function(){
    echo 'Subdir: beforeAll'  . PHP_EOL;
});

test('first', function() {
    $this->assertTrue(true);
});

test('second', function() {
    $this->assertTrue(true);
});
```
3) `./vendor/bin/pest`

Result:
```
Global: beforeAll
Global: beforeAll
Subdir: beforeAll

   PASS  Tests\subdir\MyTest
  ✓ first
  ✓ second

  Tests:  2 passed
  Time:   0.05s
```
Adding a third test,  "Global: beforeAll" is shown 3 times. Note that "Subdir: beforeAll" shows up only once.
I expected "Global: beforeAll" to appear only once. Did I misunderstand something?

Edit: ah, and I'm on PHP 8.0.8

jordanbrauer:
Interesting. Apologies if this has been a blocker for you! 

I'll take a look at this and see if we can't get it reproduced with a test case to start on a fix. 🙂

fnzr:
I've been debugging it a bit, and all repeated calls to the global setup on Pest.php are called before any test runs.

It seems that for each test in the file, the global beforeAll hook is being stacked on the Testable::$beforeAll variable.  When Testable::setUpBeforeClass executes the $beforeAll closure on line 198, the entire Testable::$beforeAll stack is executed (the exact function call to the global hook happens on the closure returned by ChainableClosure::fromStatic, line 32). 

Basically, the problem happens because each test, when it's being parsed, registers the global hook on a shared variable (Testable::$beforeAll) as if it were a new hook. The solution would be...well, not do that.  Sorry, I dont understand the library enough to propose a fix.

jordanbrauer:
@fnzr no worries, thank you for the information! Should be useful when whipping a fix up. Sounds like we need to do some sort of reset in this case.

[...]

PR #357, opened 2021-07-22
Fix static hooks

jordanbrauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #351 

### Issue

The issue is essentially, exactly, as described by @fnzr in the related ticket (great job and good find!).

> Basically, the problem happens because each test, when it's being parsed, registers the global hook on a shared variable (Testable::$beforeAll) as if it were a new hook. The solution would be...well, not do that. [...]

More specifically it is a result of the static nature of these hooks, which is why we do not see this behaviour on the `beforeEach` and `afterEach` variants. This would have been a huge oversight on my part when initially implementing this feature in #282 , and could pose a potential performance issue for some folks, should they be doing any heavy work or network activity within.

### Solution

Funnily enough, it's a one-liner ... 😑 ... 🤣 . Nearly all changes are test related.

#### Source

- ensure that we re-initialize `static ?Closure $beforeAll` and `static ?Closure $afterAll` at constructor-time for each test case, within the `Testable` concern trait;

#### Tests

Due to how these hooks work (e.g., can be defined in multiple files, variable amount of tests using them, etc.), it's a little tricky to ensure they are only run once per file. The solution to this is outlined below.

- add a `calls` property to the `global $globalHook` object;
    - within here we keep a running total of hook calls, starting at 0;
- within each hook (we are testing), assert & increment it's call count;
    - call count is always incremented by 1, except in the tests themselves;
    - when asserting the call count, depending on the test we need to calculate and apply a delta/offset to our expected value;
        - see [`BeforeAllTest.php`](https://github.com/jordanbrauer/pest/blob/fix-static-hooks/tests/Hooks/BeforeAllTest.php#L6-L8) and [`AfterAllTest.php`](https://github.com/jordanbrauer/pest/blob/fix-static-hooks/tests/Hooks/AfterAllTest.php#L4-L8) header comments for full details

### DIFF
--- a/src/Concerns/Testable.php
+++ b/src/Concerns/Testable.php
@@ -73,7 +73,8 @@ trait Testable
     {
         $this->__test        = $test;
         $this->__description = $description;
-        self::$beforeAll     = self::$afterAll = null;
+        self::$beforeAll     = null;
+        self::$afterAll      = null;
 
         parent::__construct('__test', $data);
     }
--- a/src/Concerns/Testable.php
+++ b/src/Concerns/Testable.php
@@ -73,6 +73,7 @@ trait Testable
     {
         $this->__test        = $test;
         $this->__description = $description;
+        self::$beforeAll     = self::$afterAll = null;
 
         parent::__construct('__test', $data);
     }
--- a/tests/Hooks/AfterAllTest.php
+++ b/tests/Hooks/AfterAllTest.php
@@ -2,26 +2,50 @@
 
 global $globalHook;
 
+// NOTE: this test does not have a $globalHook->calls offset since it is first
+// in the directory and thus will always run before the others. See also the
+// BeforeAllTest.php for details.
+
 uses()->afterAll(function () use ($globalHook) {
     expect($globalHook)
         ->toHaveProperty('afterAll')
         ->and($globalHook->afterAll)
-        ->toBe(0);
+        ->toBe(0)
+        ->and($globalHook->calls)
+        ->afterAll
+        ->toBe(1);
 
     $globalHook->afterAll = 1;
+    $globalHook->calls->afterAll++;
 });
 
 afterAll(function () use ($globalHook) {
     expect($globalHook)
         ->toHaveProperty('afterAll')
         ->and($globalHook->afterAll)
-        ->toBe(1);
+        ->toBe(1)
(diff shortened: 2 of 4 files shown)


jordanbrauer:
CC @nunomaduro @owenvoke @lukeraymonddowning 

Requesting a review for this bug fix when you find some time 🙂

src/Concerns/Testable.php
```diff
@@ -73,6 +73,7 @@ public function __construct(Closure $test, string $description, array $data)
     {
         $this->__test        = $test;
         $this->__description = $description;
+        self::$beforeAll     = self::$afterAll = null;
```
lukeraymonddowning, line 76: I know this is ripe coming from the guy who promotes higher order expectations inside higher order tests but should we split this over two lines for clarity?

lukeraymonddowning:
@jordanbrauer just for my own understanding, if we add another test to the AfterEachHook test down the line, we'll need to update the BeforeEachHook test offset, yes?

src/Concerns/Testable.php
jordanbrauer: Probably, yeah. It is kind of easy to miss, if not careful. Two lines it is!

jordanbrauer:
> if we add another test to the AfterEachHook test down the line, we'll need to update the BeforeEachHook test offset, yes?

@lukeraymonddowning , not quite, and thank goodness haha. The offset would only need to be updated if another test file was added to the `tests/Hooks/` directory.

For a better visual of why offset is `2`,

```
tests/Hooks
├── AfterAllTest.php           <-- no tests run before this one, so no offset
├── AfterEachTest.php          <-- 2 is our offset for the hook calls count
├── BeforeAllTest.php          <-- two tests run before this one
└── BeforeEachTest.php
```

This kind of sucks though still, because if we ever have a feature to randomize the test order (or if we currently do), it would fail for false positive. However, that being said, shouldn't be a problem since there are snapshot tests that indirectly care about the order.

Here I added some dummy tests that just expect true, but no changes to offset.

image

src/Concerns/Testable.php
jordanbrauer: Updated!

PR #361, opened 2021-07-27
Add `toThrow` expectation

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | -

I first learned about this technique from @adamwathan's TDD course years ago. The existing `throws()` has the following limitations:
1. Can only assert 1 exception is thrown per test
2. Cannot make assertions on the exception itself
3. Cannot make post-exception assertions (think side effects)

**Example:**

```php
test('all within a single test', function () {
    expect(fn() => $code->thatThrowsRuntimeException())->toThrow(RuntimeException::class);

    // a parent exception is allowed
    expect(fn() => $code->thatThrowsRuntimeException())->toThrow(Exception::class);

    // the toThrow closure's first parameter is used as the expected exception, the closure is executed after exception is caught
    expect(fn() => $code->thatThrowsRuntimeException())->toThrow(function(RuntimeException $e) {
        // make follow-up assertions on the exception itself
        expect($e->getMessage())->toBe('my message');

        // ... exceptions to check side effects (ie something doesn't exist in the database)
    });

    // "not" works
    expect(function () { throw new Exception(); })->not->toThrow(RuntimeException::class);
});
```

If this is of interest, I can polish up this PR as required.


src/Expectation.php
```diff
@@ -715,6 +715,33 @@ public function toMatchConstraint(Constraint $constraint): Expectation
         return $this;
     }
 
+    /**
+     * Asserts that executing value throws an exception.
+     *
+     * @param string|\Closure $exception string: the exception class
```
nunomaduro, line 721: If the given string, is not a class, can you assert that an exception as thrown with the given message?

src/Expectation.php
```diff
@@ -715,6 +715,33 @@ public function toMatchConstraint(Constraint $constraint): Expectation
...
+     *                                   Closure: first parameter = exception class
+     */
+    public function toThrow($exception): Expectation
+    {
+        $callback = function () {};
```
nunomaduro, line 726: There is an helper for this stuff on the core: NullClosure or something.

kbond:
I added some checks (that throw `LogicException`'s) around the expected format of the passed closure. I'm not sure if this is something you'd like to have - I can remove if desired.

src/Expectation.php
```diff
@@ -715,6 +716,53 @@ public function toMatchConstraint(Constraint $constraint): Expectation
         return $this;
     }
 
+    /**
+     * Asserts that executing value throws an exception.
+     *
+     * @param string|\Closure $exception string: the exception class
+     *                                   Closure: first parameter = exception class
+     */
+    public function toThrow($exception): Expectation
```
nunomaduro, line 725: I am sorry, but this is missing the same API as `throws` where you can specify the Exception class + the message.

kbond:
I believe this is ready but let me know if you'd like any other changes 🙂

PR #362, opened 2021-07-27
Add `throwsIf` exception

mertasan:
Moved: I submitted a new PR #371


owenvoke (commented):
Maybe `throwsIf()` would be more consistent for the naming.

mertasan:
@owenvoke you're right, I updated 👍

nunomaduro:
I feel that `throwsWhen` reads better. What do you think?

owenvoke (commented):
Would you mind running `composer update:snapshots` to update the snapshots?

This looks good to me otherwise though. 👍🏻

mertasan:
Why does not work? What am I forgetting? @owenvoke 

https://user-images.githubusercontent.com/13007665/128540237-0ffa1579-8424-4e31-a2bc-90f135b55e09.mov

PR #364, opened 2021-07-28
Datasets can now access the test case

lukeraymonddowning:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

Howdy!

This PR adds support for bound datasets. These datasets have full access to the test case and are resolved after the `beforeEach` has run.

This allows for some really cool scenarios that were never possible before in datasets. For example, you could provide full blown models via datasets:

```php
it('allows any user to be deleted', function(User $user) {
    $user->delete();
    expect(User::all())->toBeEmpty();
})->with([
    fn() => User::factory()->create(),
    fn() => User::factory()->blocked()->create(),
    fn() => User::factory()->unverified()->create(),
]);
```

This also works with shared datasets:

```php
// Datasets/Users.php
dataset('users', function() {
    yield fn() => User::factory()->create();
    yield fn() => User::factory()->blocked()->create();
    yield fn() => User::factory()->unverified()->create();
});

// Test
it('allows any user to be deleted', function(User $user) {
    $user->delete();
    expect(User::all())->toBeEmpty();
})->with('users');
```

Note that any closure will be resolved, so in order to create a dataset that returns closures, you'd need to return a closure from the closure. I've never ever returned a closure from a dataset before though, so can't see this being a common thing.

Excited to hear your thoughts!

Kind Regards,
Luke

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,7 @@
     "scripts": {
         "lint": "php-cs-fixer fix -v",
         "test:lint": "php-cs-fixer fix -v --dry-run",
-        "test:types": "phpstan analyse --ansi --memory-limit=0",
+        "test:types": "phpstan analyse --ansi",
         "test:unit": "php bin/pest --colors=always --exclude-group=integration",
         "test:integration": "php bin/pest --colors=always --group=integration",
         "update:snapshots": "REBUILD_SNAPSHOTS=true php bin/pest --colors=always",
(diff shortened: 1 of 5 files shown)


fabio-ivona:
This sounds great!

nunomaduro (approved):
@lukeraymonddowning Can you fix conflicts?

PR #367, opened 2021-08-04
Add `toBeTruthy` and `ToBeFalsy`

gabbanaesteban:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

This PR adds `toBeTruthy` and `toBeFalsy`.

Useful when you don't care what a value is and you want to ensure a value is true or false in a boolean context.

```php
expect("true")->toBeTruthy(); // passes
expect("0")->toBeTruthy(); // doest not pass

expect("0")->toBeFalsy(); // passes
expect("false")->toBeFalsy(); // doest not pass
```

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -206,6 +206,16 @@ final class Expectation
     }
 
     /**
+     * Asserts that the value is truthy.
+     */
+    public function toBeTruthy(): Expectation
+    {
+        Assert::assertTrue((bool) $this->value);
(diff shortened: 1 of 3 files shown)


nunomaduro (changes requested):

owenvoke (approved):
Nice new addition! LGTM after @nunomaduro's comments have been addressed.

PR #368, opened 2021-08-04
Install php-cs-fixer globally in workflow

abenerd:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Fixed tickets | no 

PHPCSFixer does not require the dependencies to be installed, installing only the fixer as a global dependency should speed up the workflow and save resources as less has to be downloaded and installed.

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -16,14 +16,11 @@ jobs:
       uses: shivammathur/setup-php@v2
       with:
         php-version: 8.0
-        tools: composer:v2
+        tools: php-cs-fixer:3
         coverage: none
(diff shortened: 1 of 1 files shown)


owenvoke (approved):
I don't see any real downside of this to be honest, if it's marginally faster, then probably worth it. 👍🏻 If CI passes, then should be fine to merge.

nunomaduro:
I rather prefer to use the local phpcs fixer.

owenvoke:
Fair enough. 👍🏻 I guess it is better to keep it consistent with local use.

PR #370, opened 2021-08-06
Parallel additions

nuernbergerA:
* extracted stuff to `MapArguments`
* implemented `--coverage` output for paratest


lukeraymonddowning:
Obviously we need to fix the static analysis issues, but we need to do that anyway on the parallel branch, so for the sake of clarity and not swamping, I'll merge this in. Looks great!

PR #371, opened 2021-08-07
Add `throwsIf` exception

mertasan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

(snapshots were not updating in this [PR](#362). I forked again.)

I have a test content is about 90 lines. This test content is only compatible with MySQL 8.0+. An exception will be thrown if the MySQL version is less than 8.

I know I can skip this test using Skip. But in this case I will have to create two of the same test. If one is MySQL 8 and above and the other is MySQL 5 it should work.
I can add if-else to test the content.
Or I can create another test with 90 lines.
I chose the fourth alternative, that is, I created a PR :)

```php
function dbVersion() {
    return (int) \DB::connection()->getPdo()->getAttribute(\PDO::ATTR_SERVER_VERSION);
}
```

```php
function data() {

    if (dbVersion() < 8) {
        throw new Exception('You need to use MySQL 8+ to make this work');
    } else {
        return [1, 2, 3, 4];
    }
    
}
```

```php
it('example', function () {

    $this->assertEquals([1, 2, 3, 4], data());
    
})->throwsIf(fn() => dbVersion() < 8, 'You need to use MySQL 8+ to make this work');
```

```php
it('example two', function () {

    $this->assertEquals([1, 2, 3, 4], data());
    
})->throwsIf(dbVersion() < 8, Exception::class, 'You need to use MySQL 8+ to make this work');
```

### DIFF
--- a/src/PendingObjects/TestCall.php
+++ b/src/PendingObjects/TestCall.php
@@ -79,6 +79,26 @@ final class TestCall
     }
 
     /**
+     * Asserts that the test throws the given `$exceptionClass` when called if the given condition is true.
+     *
+     * @param Closure|bool|int $condition
+     */
+    public function throwsIf($condition, string $exception, string $exceptionMessage = null): TestCall
+    {
+        $condition = is_callable($condition)
+            ? $condition
+            : Closure::fromCallable(function () use ($condition): bool {
+                return (bool) $condition;
+            });
+
+        if ($condition() === true) {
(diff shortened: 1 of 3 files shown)


owenvoke (approved):

nunomaduro:
Can you pull request docs?

PR #372, opened 2021-08-11
Allows the higher order scope to be reset using `and`

lukeraymonddowning:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

Hello there!

This PR adds the ability for a user to reset scope whilst using higher order expectations. Allow me to explain the need for this using an example:

```php
expect(['foo' => ['bar' => 'baz']])
    ->foo->bar->toBe('baz')
    ->toHaveCount(1);
```

The above code will fail because `toHaveCount` is still in the context of `->foo->bar`, which is a string. To resolve this issue, I have proposed being able to pass the `and` property or method without any arguments, which causes scope to be reset to the original expectation value:

```php
expect(['foo' => ['bar' => 'baz']])
    ->foo->bar->toBe('baz')
    ->and->toHaveCount(1); // As a property

expect(['foo' => ['bar' => 'baz']])
    ->foo->bar->toBe('baz')
    ->and()->toHaveCount(1); // As a method
```

Another example of a situation solved by this code can be found in this tweet by @Gummibeer: https://twitter.com/devgummibeer/status/1425451138384637956?s=20

Kind Regards,
Luke

### DIFF
--- a/src/HigherOrderExpectation.php
+++ b/src/HigherOrderExpectation.php
@@ -70,10 +70,14 @@ final class HigherOrderExpectation
      *
      * @param TValue $value
      *
-     * @return Expectation<TValue>
+     * @return Expectation<TValue>|Expectation
      */
-    public function and($value): Expectation
+    public function and($value = null): Expectation
     {
+        if ($value === null) {
+            return $this->original;
+        }
+
         return $this->expect($value);
     }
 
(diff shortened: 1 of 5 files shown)


nunomaduro:
Holding this one for now.

PR #387, opened 2021-08-28
Fix case sensitivity when finding test file

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no

This fixes a really weird issue cause by how debug_backtrace works on case-insensitive file systems.

When using `debug_backtrace` on case insensitive file system it reports the file path as how `pwd` would report it. 
Meaning if I have a folder named `something` and `cd` into it as `SOMEtHing` and run my script, then `debug_backtrace` would report that folder as `SOMEtHing`.
However doing a `realpath` would report it as `something`. 

In pest we have a case where wo look at both file path from `debug_backtrace` and `realpath`, so to make sure we don't hit issues with this, we should lower case both file paths, so this won't happen. 

One potential problem with doing this is one actually case sensitive file system, this could in theory give false positives.

### DIFF
--- a/src/Support/Backtrace.php
+++ b/src/Support/Backtrace.php
@@ -25,8 +25,10 @@ final class Backtrace
     {
         $current = null;
 
+        $filerLoaderPath = mb_strtolower((string) realpath('vendor/phpunit/phpunit/src/Util/FileLoader.php'));
+
         foreach (debug_backtrace(self::BACKTRACE_OPTIONS) as $trace) {
-            if (Str::endsWith($trace[self::FILE], (string) realpath('vendor/phpunit/phpunit/src/Util/FileLoader.php'))) {
+            if (Str::endsWith(mb_strtolower($trace[self::FILE]), $filerLoaderPath)) {
                 break;
             }
 


nunomaduro:
@lukeraymonddowning @owenvoke What are your thoughts about this?

nunomaduro (approved):

lukeraymonddowning:
My honest opinion is that this is user error.

olivernybroe:
@lukeraymonddowning I think I would say I agree with you here 👍 

A possible better solution would be a better error handling of it? 
So right now we throw a `ShouldNotHappen` exception if the stacktrace is empty, however if the stacktrace is there, but the `FileLoader` is just not found, it actually returns that the test file is the second last parameter in the stacktrace, resulting in all sort of weird errors.

I think it would make sense to make it throw a `ShouldNotHappen` exception, if the `FileLoader` class is not found. WDYT?

nunomaduro:
@lukeraymonddowning Why this is a user error?

lukeraymonddowning:
@nunomaduro because they cd'd into the wrong directory by case. In an operating system that was case sensitive, this could be an entirely different folder in the file system. It just happens to be the case that MacOS is case insensitive.

olivernybroe:
So just for a little reference of why I even discovered this error. 
If you have created a project in PhpStorm and setup your configs ( e.g. where to find pest executable and so on), and then rename the root folder. PhpStorm will not automatically rename all references in the config with it, meaning if you change the casing of the folder, they are still kinda valid, because filesystem being case insensitive, but php is not 🤔

nunomaduro:
@owenvoke What is your take on this?

owenvoke:
Hmm. I'd have to agree that this is user error, however this could trip people up. I think we should probably leave it as is, however maybe flag this up in the docs somewhere? 🤔

olivernybroe:
@owenvoke Yes, this is also why I suggest at least to make it throw an exception.

We could then implement ignition solutions to it, to show people about how to solve this problem 😃

nunomaduro:
Going to hold this one.

PR #391, opened 2021-09-03
Add `toHaveProperties`

gabbanaesteban:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

This PR adds `ToHaveProperties`.

It is like [toHaveKeys](https://pestphp.com/docs/expectations#expect-toHaveKeys) but for object.
Useful when you want to assert the object has the expected properties but you don't care about the values.

```php
$obj->name = "Jhon";
$obj->age = 21;

expect($obj)->toHaveProperties(["name", "age"]); // passes
expect($obj)->toHaveProperties(["name", "age", "gender"]); // doest not pass
```

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -389,9 +389,9 @@ final class Expectation
     /**
      * Asserts that the value contains the provided properties $names.
      *
-     * @param array<string> $names
+     * @param iterable<array-key, string> $names
      */
-    public function toHaveProperties(array $names): Expectation
+    public function toHaveProperties(iterable $names): Expectation
     {
         foreach ($names as $name) {
             $this->toHaveProperty($name);
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -387,6 +387,20 @@ final class Expectation
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Expectation.php
```diff
@@ -386,6 +386,20 @@ public function toHaveProperty(string $name, $value = null): Expectation
         return $this;
     }
 
+    /**
+     * Asserts that the value contains the provided properties $names.
+     *
+     * @param array<string> $names
```
nunomaduro, line 392: Can you use `iterable<array-key, string>`?

src/Expectation.php
```diff
@@ -386,6 +386,20 @@ public function toHaveProperty(string $name, $value = null): Expectation
...
+     */
+    public function toHaveProperties(array $names): Expectation
```
nunomaduro, line 394: `array $names` should be `iterable $names`.

nunomaduro (changes requested):
Also, can you prepare the docs pull request too?

nunomaduro:
Perfect, only docs missing now.

PR #399, opened 2021-09-16
Fix: `sequence()` can generate false positives

### LINKED ISSUE #398: `sequence()` can generate false positives
I think this test should fail.

https://github.com/pestphp/pest/blob/2f0cd7a4e3a36ac51367ea45d65fc09403ec32c7/tests/Features/Expect/sequence.php#L29-L38

Useful when working with static data. Even so, why test more than the actual data?

But it generates false positives in a use case like the one below.

```php
expect(new Model())
    ->all()
    ->toHaveCount(2) // << Assuming this isn't added
    ->sequence(
        fn ($model) => $model->title->toEqual('foo'),
        fn ($model) => $model->title->toEqual('bar'),
        fn ($model) => $model->title->toEqual('baz'), // << There is no such item.
    );
```

my solution:

```php
Assert::assertLessThanOrEqual(count($value), count($callbacks));
```

https://github.com/pestphp/pest/blob/2f0cd7a4e3a36ac51367ea45d65fc09403ec32c7/src/Expectation.php#L167

This assertion should be added to this line.

Do you support this idea?

Thanks.

mertasan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #398 

This resolves #398

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -154,15 +154,20 @@ final class Expectation
             throw new BadMethodCallException('Expectation value is not iterable.');
         }
 
-        $value  = is_array($this->value) ? $this->value : iterator_to_array($this->value);
-        $keys   = array_keys($value);
-        $values = array_values($value);
+        $value          = is_array($this->value) ? $this->value : iterator_to_array($this->value);
+        $keys           = array_keys($value);
+        $values         = array_values($value);
+        $callbacksCount = count($callbacks);
 
         $index = 0;
 
         while (count($callbacks) < count($values)) {
             $callbacks[] = $callbacks[$index];
             $index       = $index < count($values) - 1 ? $index + 1 : 0;
+        }
+
+        if ($callbacksCount > count($values)) {
+            Assert::assertLessThanOrEqual(count($value), count($callbacks));
         }
 
         foreach ($values as $key => $item) {
--- a/tests/.snapshots/success.txt
+++ b/tests/.snapshots/success.txt
@@ -168,7 +168,7 @@
   ✓ an exception is thrown if the the type is not iterable
   ✓ allows for sequences of checks to be run on iterable data
   ✓ loops back to the start if it runs out of sequence items
-  ✓ it works if the number of items in the iterable is smaller than the number of expectations
+  ✓ fails if the number of iterable items is greater than the number of expectations
   ✓ it works with associative arrays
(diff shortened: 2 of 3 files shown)


src/Expectation.php
```diff
@@ -165,6 +166,10 @@ public function sequence(...$callbacks): Expectation
             $index       = $index < count($values) - 1 ? $index + 1 : 0;
         }
 
+        if ($callbacksCount > count($values)) {
```
mertasan, line 169: The reason I added this condition:

https://github.com/pestphp/pest/blob/2f0cd7a4e3a36ac51367ea45d65fc09403ec32c7/tests/Features/Expect/sequence.php#L18-L27

nunomaduro (approved):
This looks good to me. @lukeraymonddowning what do you think?

lukeraymonddowning:
Makes sense to me. I can't see a use case where you'd want to have unmet expectations.

src/Expectation.php
```diff
@@ -154,9 +154,10 @@ public function sequence(...$callbacks): Expectation
             throw new BadMethodCallException('Expectation value is not iterable.');
         }
 
-        $value  = is_array($this->value) ? $this->value : iterator_to_array($this->value);
-        $keys   = array_keys($value);
-        $values = array_values($value);
+        $value          = is_array($this->value) ? $this->value : iterator_to_array($this->value);
+        $keys           = array_keys($value);
+        $values         = array_values($value);
+        $callbacksCount = count($callbacks);
```
lukeraymonddowning, line 160: I don't really see the need for the temporary variable, especially as we don't use it later. Might as well just inline it further down

mertasan: The count of `$callbacks` may increase in this line. So I used temporary variable.

https://github.com/pestphp/pest/blob/a16a19e1214f4e9e817b775ab395dcb52e45c42b/src/Expectation.php#L165

The count is increasing for the following feature.

https://github.com/pestphp/pest/blob/2f0cd7a4e3a36ac51367ea45d65fc09403ec32c7/tests/Features/Expect/sequence.php#L18-L27

lukeraymonddowning: My bad. Doh! It's getting late over here 😁

mertasan: No one saw it, don't worry. hahah

PR #401, opened 2021-09-19
Adds new `to()` method

mertasan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

It allows us to change the value/values in `expect()`.

Also useful for Higher Order Tests. Also, there is no need to add temporary variables to pass new values to expectations.

Like this: `->and($data)`

It's like a merge of the `and()` and `sequence()` methods. It has the advantage of being able to change the data for the continuation of expectations.

To use the `and()` method, you must pass a variable. However, with this method, we can do it all in one.

It's as useful as the `toArray()` and `json()` methods, but gives more freedom for the type of value to convert.

```php
test('parse markdown', function () {

    expect('# Foo')
        ->toBeString()
        ->to(
            fn ($value) => Parser::markdown($value)
        )
        ->toBe('<h1>Foo</h1>');
});
```

```php
test('parse currency', function () {

    expect([1, 2, 9999.99])
        ->toBeArray()
        ->toHaveCount(3)
        ->to(
            fn ($value) => Currency::parse($value)
        )
        ->toMatchArray(['1.00 USD', '2.00 USD', '10,000 USD']);
});
```

```php
test('complex', function () {

    expect(['# Foo', 2])
        ->toBeArray()
        ->toHaveCount(2)
        ->to(
            fn ($value) => Parser::markdown($value),
            fn ($value) => Currency::parse($value),
        )
        ->toMatchArray(['<h1>Foo</h1>', '2.00 USD']);
});
```

```php
it('validates the user names')
    ->expect(fn() => Auth::user())
    ->email->toEndWith('bar.com')
    ->to(fn($user) => ValidateUserName::isValid($user->username))
    ->toBeTrue();
```

For more examples, please review the tests.

Thank you.

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -850,6 +850,51 @@ final class Expectation
     }
 
     /**
+     * The current value can be changed before the next expectation.
+     *
+     * @template TToValue
+     *
+     * @param (callable(TToValue): TToValue)|mixed ...$callbacks
+     */
+    public function to(...$callbacks): Expectation
+    {
+        if (is_array($this->value) || is_iterable($this->value)) {
+            $value          = is_array($this->value) ? $this->value : iterator_to_array($this->value);
+            $keys           = array_keys($value);
+            $values         = array_values($value);
+            $callbacksCount = count($callbacks);
+
+            $index = 0;
+
+            while (count($callbacks) < count($values)) {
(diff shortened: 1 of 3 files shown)


nunomaduro:
No plans for this one.

PR #405, opened 2021-09-22
Add new --ci option to pest bin

fabio-ivona:
Hi guys!

as already discussed on the discord channel with  @nunomaduro , this PR wants to put a base for adding a new "environment aware" behaviour to pestphp

**background** 

my main idea comes from my bad habit to forget to remove `->only` method from a test (maybe added when fixing a bug) and this results the pipeline to execute only a partial set of my test suite, with the risk to deploy some other bug in production.

**the idea** 

Pestphp could be able to detect the environment it is running into and sligthing changing its behaviour according to it. 

to hint pest it is running in a CI pipeline, it could be launched with the `--ci` option:

`vendor/bin/pest --colors=always --ci`

and it would ignore all `->only()` directives given to it

**the implementation**

I added a new `$workingEnv` attribute to the `TestSuite` class that holds the current working environment. Its default value is 'local' and it is set to `ci` if the pest binary is called with the `--ci` option

this way the `->only()` behaviour can be disabled when in a CI environment:

```php
final class TestRepository
{
  //...
  private function testsUsingOnly(): array
      {
          if (TestSuite::getInstance()->workingEnv === 'ci') {
              return [];
          }
  
          return array_filter($this->state, function ($testFactory): bool {
              return $testFactory->only;
          });
      }
   //...
```

**future uses**

the `TestSuite::getInstance()->workingEnv` attribute could be used in future implementation to fine tuning pest behaviour when in a continuous integration environment (or other ones, defined as necessary)


nunomaduro:
Can you refactor this, in a way that the option gets "setted" using a plugin class? https://github.com/pestphp/pest/tree/master/src/Plugins.

fabio-ivona:
> Can you refactor this, in a way that the option gets "setted" using a plugin class? https://github.com/pestphp/pest/tree/master/src/Plugins.

Done, I've extracted environment information in `Context` Plugin, this way we can add more contextual info in next developments, if necessary

fabio-ivona:
@nunomaduro there are some strange errors in tests, are there currently any issues with them?

nunomaduro:
Not sure why, but can't resolve the conflicts. Can you do it please?

fabio-ivona:
> Not sure why, but can't resolve the conflicts. Can you do it please?

@nunomaduro I've solved merge conflicts

CI errors keep showing, I think they are related to the ones showing in #409

nunomaduro:
@fabio-ivona Can you add docs about this?

fabio-ivona:
> @fabio-ivona Can you add docs about this?

Sure! It's already written, will open a PR this night

PR #406, opened 2021-09-23
Adds new `when()` method

mertasan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

It makes some expectations work conditionally.

`when()` is preferably used as an alternative to the `skip()` method, no need to write extra tests.

#### Before:
```php
it('mysql test', function () {

   expect(new Foo())
         ->mysql
         ->toBeTrue();

})->skip(fn() => DB::getDriverName() === 'mysql', 'Only runs when using mysql');

it('sqlite test', function () {

   expect(new Foo())
         ->sqlite
         ->toBeTrue();

})->skip(fn() => DB::getDriverName() === 'sqlite', 'Only runs when using sqlite');
```

#### After:
```php
it('database test', function () {

    $driver = DB::getDriverName();
    
    expect(new Foo())
        ->when($driver === 'mysql',
            fn($foo) => $foo->mysql->toBeTrue()
        )
        ->when($driver === 'sqlite',
            fn($foo) => $foo->sqlite->toBeTrue()
        )
        ->bar->toEqual('bar');

});
```

For more examples, please review the tests.

Thank you.

PR #407, opened 2021-09-23
Adds new `match()` method

mertasan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

This is an alternative to #406.

```php
it('database test', function () {
    
    expect(new Foo())
        ->foo->toBe('foo')
        ->match(DB::getDriverName(), [
            'mysql'  => fn ($foo) => $foo->mysql->toBeTrue(),
            'sqlite' => fn ($foo) => $foo->sqlite->toBeTrue(),
        ])
        ->bar->toBe('bar');

});
```

```php
it('builder test', function ($driver) {
    
    expect(new Builder())
        ->match($driver, [
            'mysql'   => fn ($b) => $b->where('...')->toBe(...),
            'postgre' => fn ($b) => $b->whereRaw('...')->toBe(...),
        ]);
    
})->with([
    'mysql',
    'postgre',
]);
```

For more examples, please review the tests.

Thank you.

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -212,7 +212,7 @@ final class Expectation
                 continue;
             }
 
-            (new self($this->value))->toEqual($callback);
+            $this->and($this->value)->toEqual($callback);
 
             break;
         }
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -178,6 +178,56 @@ final class Expectation
     }
 
     /**
+     * If the subject matches one of the expressions, the callback in the expression is run.
+     *
+     * @template TMatchValue
+     *
+     * @param Closure|bool|string                            $subject
+     * @param array<mixed, callable(self): void|TMatchValue> $expressions
+     *
+     * @return \Pest\Expectation
+     */
+    public function match($subject, array $expressions): Expectation
+    {
+        $subject = is_callable($subject)
+            ? $subject
+            : function () use ($subject) {
+                return $subject;
+            };
+
+        $subject   = $subject();
+        $keys      = array_keys($expressions);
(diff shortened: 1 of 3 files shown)


nunomaduro:
@lukeraymonddowning What do you think?

src/Expectation.php
```diff
@@ -177,6 +177,56 @@ public function sequence(...$callbacks): Expectation
...
+                continue;
+            }
+
+            (new self($this->value))->toEqual($callback);
+
+            break;
+        }
+
+        if (!$matched) {
+            test()->addWarning('No item found matching "' . $subject . '".');
```
nunomaduro, line 224: This should be removed.

src/Expectation.php
```diff
@@ -177,6 +177,56 @@ public function sequence(...$callbacks): Expectation
...
+            }
+
+            $matched = true;
+
+            if (is_callable($callback)) {
+                $callback(new self($this->value));
+                continue;
+            }
+
+            (new self($this->value))->toEqual($callback);
```
nunomaduro, line 218: `this->and`

src/Expectation.php
nunomaduro: And add a `$this->toHaveKey(` before all this code.

mertasan: @nunomaduro Can you explain this in more detail please? I don't understand where to add it. `$this->toHaveKey(..?.)`

lukeraymonddowning:
I actually like this a lot. It's a sharp knife, for sure, but a valuable one I think.

PR #409, opened 2021-09-24
fix: adds failing test

nunomaduro:
cc @lukeraymonddowning is  this failing test normal?

### DIFF
--- a/tests/Features/Expect/sequence.php
+++ b/tests/Features/Expect/sequence.php
@@ -28,6 +28,17 @@ test('loops back to the start if it runs out of sequence items', function () {
     expect(static::getCount())->toBe(16);
 });
 
+test('it works if the number of items in the iterable is bigger than the number of expectations', function () {
+    expect([1, 2, 3, 4])
+        ->sequence(
+            function ($expectation) { $expectation->toBeInt()->toEqual(1); },
+            function ($expectation) { $expectation->toBeInt()->toEqual(2); },
+            function ($expectation) { $expectation->toBeInt()->toEqual(3); },
+        );
+
+    expect(static::getCount())->toBe(6);
+});
+
 test('fails if the number of iterable items is greater than the number of expectations', function () {
     expect([1, 2])
         ->sequence(


mertasan:
I think failing is normal.

The fourth loop matches the first `sequence` item:

```php
test('it works if the number of items in the iterable is bigger than the number of expectations', function () {
    expect([1, 2, 3, 4])
        ->sequence(
            function ($expectation) { $expectation->toBeInt()->toEqual(1); }, // 1 - 4
            function ($expectation) { $expectation->toBeInt()->toEqual(2); }, // 2
            function ($expectation) { $expectation->toBeInt()->toEqual(3); }, // 3
        );

    expect(static::getCount())->toBe(6);
});
```

```php
expect([1, 2, 3, 4, 5, 6, 7, 8])
    ->sequence(
        function ($expectation) { $expectation->toBeInt()->toBeIn([1, 4, 7]); },
        function ($expectation) { $expectation->toBeInt()->toBeIn([2, 5, 8]); },
        function ($expectation) { $expectation->toBeInt()->toBeIn([3, 6]); },
    );

expect(static::getCount())->toBe(16);
```

lukeraymonddowning:
As originally intended, sequence will loop back over the closures until all items in the iterable had been checked. So in the case of this failing test, there would be 8 assertions, not 6, because the first expectation in the sequence would be called twice (for the first and last item in the sequence).

fabio-ivona:
Another error seems to rise in `src/PendingObjects/TestCall.php:91`

I think `mixed` return will work only from php8.0 onward ([see here](https://php.watch/versions/8.0/mixed-type))

image


Why did you need to use `mixed` instead of a simple `bool`? wasn't $condition parameter of type `bool`?

`@param (callable(): bool)|bool $condition`

otherwise the return type on line 91 should not be defined at all

nunomaduro:
@fabio-ivona fixed.

@lukeraymonddowning "because the first expectation in the sequence would be called twice (for the first and last item in the sequence)." This seems weird to me. I would expect the first expectation called once, on the first element of the iterable. There is a reasoning about this behaviour ?

lukeraymonddowning:
The reasoning was to match how sequence works in Laravel factories, which is loosely what this was based on.

nunomaduro:
Thank you @lukeraymonddowning

PR #413, opened 2021-10-01
Fix skipping tests with exception asserting

fabio-ivona:
| Q             | A
| ------------- | ---
| Bug fix?      | yes

This PR will fix skipped text failing if there is a throw expectation (see the test I've added in `tests/Feature/Exceptions.php`)

this test won't be skipped:

```php
it('can handle a skipped test if it is trying to catch an exception', function () {
    expect(1)->toBe(2);
})->throws(ExpectationFailedException::class)->skip('this test should be skipped');
```

but it will trigger an error instead:

```
Failed asserting that exception of type "PHPUnit\Framework\SkippedTestError" matches expected exception "PHPUnit\Framework\ExpectationFailedException". Message was: "this test should be skipped"
```

**Edit** I've proposed a cleaner solution than this one, see [this comment](https://github.com/pestphp/pest/pull/413#issuecomment-932698555)
~~The fix is made by adding two ne methods to `HigherOrderMessageCollection` (`hasMessage($name)` and `forgetMessage($name)`) and using them in `TestCaseFactory` to detect skipped tests and removing all exception expectation from them:~~

```php
/// TestCaseFactory.php

 if ($chains->hasMessage('markTestSkipped')) {
      $proxies->forgetMessage('expectException');
  }
```

### DIFF
--- a/src/Factories/TestCaseFactory.php
+++ b/src/Factories/TestCaseFactory.php
@@ -140,6 +140,10 @@ final class TestCaseFactory
         $proxies     = $this->proxies;
         $factoryTest = $this->test;
 
+        if ($chains->hasMessage('markTestSkipped')) {
+            $proxies->forgetMessage('expectException');
+        }
+
         /**
          * @return mixed
          */
--- a/src/Factories/TestCaseFactory.php
+++ b/src/Factories/TestCaseFactory.php
@@ -140,16 +140,12 @@ final class TestCaseFactory
         $proxies     = $this->proxies;
         $factoryTest = $this->test;
 
-        if ($chains->hasMessage('markTestSkipped')) {
-            $proxies->forgetMessage('expectException');
-        }
-
         /**
          * @return mixed
          */
         $test = function () use ($chains, $proxies, $factoryTest) {
-            $proxies->proxy($this);
             $chains->chain($this);
+            $proxies->proxy($this);
 
             /* @phpstan-ignore-next-line */
             return call_user_func(Closure::bind($factoryTest, $this, get_class($this)), ...func_get_args());
--- a/src/Support/HigherOrderMessageCollection.php
+++ b/src/Support/HigherOrderMessageCollection.php
@@ -69,24 +69,4 @@ final class HigherOrderMessageCollection
             0,
         );
     }
-
-    public function hasMessage(string $name): bool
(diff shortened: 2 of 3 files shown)


nunomaduro:
Interesting. How PHPUnit behaves in this scenario?

fabio-ivona:
It seems that PHPUnit can handle such a scenario without any issue:

```php
class ExceptionTest extends TestCase
{
    /** @test */
    public function test_can_handle_skipping_the_test_if_it_is_expecting_an_exception()
    {
        $this->expectException(\PHPUnit\Framework\MockObject\BadMethodCallException::class);
        $this->markTestSkipped("non si fa");
    }
}
```

```
PHPUnit 9.5.10 by Sebastian Bergmann and contributors.

S                                                                   1 / 1 (100%)

Time: 00:00.002, Memory: 4.00 MB

OK, but incomplete, skipped, or risky tests!
Tests: 1, Assertions: 0, Skipped: 1.
```




@nunomaduro another thing before rewieving this implementation:

I noticed that this issue can be solved by simply inverting these two calls in `TestCaseFactory.php`, without the need to add new methods to `HigherOrderMessageCollection`:

```diff
$test = function () use ($chains, $proxies, $factoryTest) {
-     $proxies->proxy($this);
-     $chains->chain($this);

+    $chains->chain($this);
+    $proxies->proxy($this);
      

      /* @phpstan-ignore-next-line */
      return call_user_func(Closure::bind($factoryTest, $this, get_class($this)), ...func_get_args());
 };
```

Tests are running without any error, do you think this would be a cleaner solution?

fabio-ivona:
@nunomaduro another thing before rewieving this implementation:

I noticed that this issue can be solved by simply inverting these two calls in `TestCaseFactory.php`, without the need to add new methods to `HigherOrderMessageCollection`:

```diff
$test = function () use ($chains, $proxies, $factoryTest) {
-     $proxies->proxy($this);
-     $chains->chain($this);

+    $chains->chain($this);
+    $proxies->proxy($this);
      

      /* @phpstan-ignore-next-line */
      return call_user_func(Closure::bind($factoryTest, $this, get_class($this)), ...func_get_args());
 };
```

Tests are running without any error, do you think this would be a cleaner solution?

PR #418, opened 2021-10-07
fix test breaks

fabio-ivona:
@nunomaduro I just got aware that my previuos PR #413 contained two breaking problems:

- a forgotten `->only()` statement limited unit tests to a single test (the one I was working on)

- other tests where failing because of my implementation and escaped without triggering an error 😫 (and it was neither detected by success snapshot)

I've fixed it with a new implementation (the original one described in #413):

The fix is made by adding two new methods to HigherOrderMessageCollection (hasMessage($name) and forgetMessage($name)) and using them in TestCaseFactory to detect skipped tests and removing all exception expectation from them:

```php
/// TestCaseFactory.php

 if ($chains->hasMessage('markTestSkipped')) {
      $proxies->forgetMessage('expectException');
  }
```

let me know if it is ok, otherwise we shuold simply revert my previous PR

### DIFF
--- a/tests/.snapshots/success.txt
+++ b/tests/.snapshots/success.txt
@@ -102,7 +102,7 @@
   ✓ it can resolve a dataset after the test case is available with shared array sets with (Closure Object (...)) #1
   ✓ it can resolve a dataset after the test case is available with shared array sets with (Closure Object (...)) #2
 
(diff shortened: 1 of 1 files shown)


nunomaduro:
Reverted #413 .

Issue #247, opened 2021-01-02
[Bug] Phpstan failing to find expectation methods


R3VoLuT1OneR:
**Description**
PHPStan level 7 failing to find methods, but the tests works fine.
Probably problem in the PHPStan but I think you should know about it.

**Steps to reproduce**
1. Prepare environment for running phpstan
```shell
mkdir /tmp/failingphpstan && /tmp/failingphpstan
composer require pestphp/pest phpstan/phpstan
mkdir tests
```

2. Create next file with failing code `PhpstanFailingTest.php`:
```php
<?php 

test('phpstan fails to found expectations', function () {

    expect(new \DateTime())->toBeInstanceOf(\DateTime::class);
    expect([])->toHaveCount(0);

});
```

3. Run phpstan analyze
```shell
vendor/bin/phpstan analyze --level=7 tests
```

**Actual result:**
```
 ------ ------------------------------------------------------------------------------------------------------------------- 
  Line   PhpstanFailingTest.php                                                                                             
 ------ ------------------------------------------------------------------------------------------------------------------- 
  5      Call to an undefined method Pest\Expectations\Expectation|Pest\Expectations\Support\Extendable::toBeInstanceOf().  
  6      Call to an undefined method Pest\Expectations\Expectation|Pest\Expectations\Support\Extendable::toHaveCount().     
 ------ ------------------------------------------------------------------------------------------------------------------- 
 [ERROR] Found 2 errors                                                                                                 
```
**Expected (Success) result:**
```                                         
 [OK] No errors                                                                                                         
```

nunomaduro:
Fell free to create a phpstan plugin or pull request something that fixes this.

PR #423, opened 2021-10-18
fix Phpstan failing to find expectation methods #247.

ericsssan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #247

fix Phpstan failing to find expectation methods #247. 

Convert Extendable class to Extenability interface.
Have Expectation implements Extendability and implements extend method. 

Rename Extendable trait extend method to macro to avoid method signature complain.


nunomaduro:
We've fixed this.

PR #424, opened 2021-10-19
[2.x] Vastly improves the logic around bound datasets to make them more user friendly

lukeraymonddowning:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

After using bound datasets and receiving feedback for some time now, I wanted to go in and overhaul how they work to make them smarter and more intuitive.

This PR adds the following smarts:

1) If your dataset returns more than one argument per set, nothing will be bound and it will be assumed that you want a closure

```php
it('resolves a potential bound dataset logically', function ($foo, $bar) {
    expect($foo)->toBe('foo');
    expect($bar())->toBe('bar');
})->with([
    ['foo', function () { return 'bar'; }], // Here we pass 2 arguments, so no binding will occur
]);
```

2) If you type hint the property in your test method as a `Closure` or `callable`, we also won't bind

```php
it('will not resolve a closure if it is type hinted as a closure', function (Closure $data) { // Note the type hint here
    expect($data())->toBeString();
})->with([
    function () { return 'foo'; }, 
    function () { return 'bar'; },
]);

it('will not resolve a closure if it is type hinted as a callable', function (callable $data) { // and here
    expect($data())->toBeString();
})->with([
    function () { return 'foo'; },
    function () { return 'bar'; },
]);
```

3) If your bound dataset returns an array, we now intelligently check the number of expected arguments and base what we return on that

```php
it('can correctly resolve a bound dataset that returns an array', function (array $data) { // Only one argument, so we'll return it as an array
    expect($data)->toBe(['foo', 'bar', 'baz']);
})->with([
    function () { return ['foo', 'bar', 'baz']; },
]);

it('can correctly resolve a bound dataset that returns an array but wants to be spread', function (string $foo, string $bar, string $baz) { // Multiple arguments, so we'll spread the result into the test function
    expect([$foo, $bar, $baz])->toBe(['foo', 'bar', 'baz']);
})->with([
    function () { return ['foo', 'bar', 'baz']; },
]);
```

I hope this makes sense. It should make it a lot more intuitive to work with datasets without getting confused.

Kind Regards,
Luke


nunomaduro:
Please adjust docs if needed.

PR #430, opened 2021-10-31
[v2.x] expectation pipes and interceptors

fabio-ivona:
This PR sparks from ideas and brainstorming that took place between me, @lukeraymonddowning and @olivernybroe , lot of credit shoud go to them

## Main idea

You can't, currently, change or override an expectation in order to adapt it's behavior to the actual value you are expecting. Let's assume you are trying to build a pest plugin that works with Laravel Model (thing that I'm actually doing :innocent:)

you'll likely want to define a new expectation:

```php
$user = Auth::user();

expect($user)->toBeSameModelAs($anotherUser);
```

and you are compelled to use that weird `toBeSameModelAs` because `toBe` is reserved

My main concern when this idea come up was to be able to override `toBe` in order to be able to write

```php
expect($user)->toBe($anotherUser);
```

and checking that the two models refers to the same database row

got the point? let's see what this has become after a nice brainstorming with Luke and Oliver

## How this evolved

the first thing we decided is that this feature should allow two different method that would team up with `expect()->extend()`

### pipe

`→pipe(...)` will allow to run the expectation through a pipeline that can be fully executed or interrupted at any time (like a middleware for functions)

let's suppose we implement the `expect($user)->toBe($anotherUser);` above, with this PR you could define this in your Pest.php file:

```php
expect()->pipe('toBe', function ($next, $expected){
    if ($this->value instanceof Model) {
        assertTrue($this->value->is($expected);

        //returning nothing stops pipeline execution
        return;
    }

    //calling $next(); let the pipeline to keep running
    $next();
});
```

### intercept

Usually you'll want to override the `toBe` expectation with a model check only when the expected value is a model, for this type of implementations there is a shorthand, `->intercept(...)`, that applies the assertion and stops the pipeline execution:

```php
expect()->intercept('toBe', Model::class, function($expected){
    assertTrue($this->value->is($expected))
});
```

### Step 1: extract core expectations

in order to pass all expectation calls through a pipeline, pest's core expectations can't be defined as methods in `Expectation.php`, but every call should go through its `__call()` magic method

I achieved this by exctracting all expectation from `Expectation.php` and moving them in a separate class `CoreExpectation.php`. This way `Expectation.php` works as a proxy for the other class and it uses `@mixin CoreExpectation` to allow IDE's code completion

for the implementation see [this commit](https://github.com/pestphp/pest/commit/8174f2d973a81cc6192195ad29fbf94040bd4249)

**alternative**

if the extraction to a `CoreExpectation` is too much, another solution would be to define all core expectation as extensions through the `expect()->extend(..)` method

### Step 2: pipes implementation

After getting all expectations run through the `Expectation.php::__call()` method, I changed it to apply a pipeline before the called expectation:

```php
public function __call(string $method, array $parameters)
{
    if (!$this->hasExpectation($method)) {
        return new HigherOrderExpectation($this, $this->value->$method(...$parameters));
    }

    ExpectationPipeline::for($this->getExpectationClosure($method))
        ->send(...$parameters)
        ->through($this->pipes($method, $this, Expectation::class))
        ->run();

    return $this;
}
```

The `ExpectationPipeline` class is a pipeline implementation inspired to Laravel's one, with a couple of differences:

- the values passed through the pipeline can be multiple
- each pipe of the pipeline don't return anything (we don't want that one pip alters the content of the expected values for the next pipes in the pipeline)

### Step 3: Interceptors implementation

Interceptors are implemented as a pipe that is applied only if the type checking is met and, if applied, stops the pipeline after it gets executed 

### Step 4: tests

I added new tests that try to both verify the correctness of my implementation and to explain the behavior of the pipes and interceptors system (see `tests/Features/Expect/pipes.php`)

### interceptors checks can be a function

this will allow complex checking:

```php
/*
 * Overrides toBe to assert all integers are allowed if value is a wildcard (*)
 */
expect()->intercept(
    'toBe', 
    fn ($value, $expected) => $value === '*' && is_numeric($expected), 
    function ($expected) use ($state) {
        //.. always passes
    }
);
```

### pipes and interceptors can add additional optional parameters to the base expectation:

```php
/*
 * Overrides toBe to allow ignoring case when checking strings
 */
expect()->intercept('toBe', fn ($value) => is_string($value), function ($expected, $ignoreCase = false) {
    if ($ignoreCase) {
        assertEqualsIgnoringCase($expected, $this->value);
    } else {
        assertSame($expected, $this->value);
    }
});
```

### why isn't `$next` argument placed at the end (like Laravel's pipes)?

we discussed about that, writing

```php
expect()->pipe('toBe', function ($expected, $next) {
    //..
});
```

would have been more familiar for who is accustomed with Laravel syntax for pipes

**but** Laravel pipes don't have optional parameters, so we decided to implement it in this way, so you can write:

```php
expect()->pipe('toHaveKey', function ($next, $key $value = null) {
    //..
});
```

### next implementations

- expectation failures from an override should report the plugin or the piece of code that triggered the failure. This will avoid lot of issues in Pest github when another plugin has done something wrong

## Wrap up

That's all! (or better.. that's just a starting point!)

let me know what do you think about it!


fabio-ivona:
@nunomaduro unit tests are failing because `pestphp/pest-plugin-mock` requires pest v1.0`, should I open a PR to it in order to change its requirement from `v1.0` to `v1.0|v2.0` ?

lukeraymonddowning:
Will do a code review shortly

src/Concerns/Extendable.php
```diff
@@ -35,6 +40,71 @@ public static function hasExtend(string $name): bool
...
+     */
+    public static function intercept(string $name, string|Closure $filter, Closure $handler): void
+    {
+        if (is_string($filter)) {
+            $filter = fn ($value, ...$arguments): bool => $value instanceof $filter;
+        }
+
+        self::pipe($name, function ($next, ...$arguments) use ($handler, $filter): void {
+            /** @phpstan-ignore-next-line  */
+            if ($filter($this->value, ...$arguments)) {
```
lukeraymonddowning, line 62: I know this is currently a single use trait, but we have no dependence on `$this->value` in this trait other than here, so if this was placed in a class that didn't have `$value` as a property, then wouldn't it blow up?

lukeraymonddowning: May be cleaner to reverse this and do an early `return $next();` inside the `if` statement. Might have less visual overload.

src/Concerns/Extendable.php
```diff
@@ -35,6 +40,71 @@ public static function hasExtend(string $name): bool
...
+                /** @phpstan-ignore-next-line  */
+                $handler = $handler->bindTo($this, get_class($this));
+
+                if($handler instanceof Closure){
```
lukeraymonddowning, line 66: Perhaps better to write this as 

```php 
if ($handler = $handler->bindTo($this, $this::class) !== false) {
    $handler(...$arguments);
}
```

?

src/Concerns/Extendable.php
```diff
@@ -35,6 +40,71 @@ public static function hasExtend(string $name): bool
...
+     */
+    public static function pipe(string $name, Closure $handler): void
+    {
+        self::$pipes[$name][] = $handler;
+    }
+
+    /**
+     * Register an interceptor that should replace an existing expectation.
+     */
+    public static function intercept(string $name, string|Closure $filter, Closure $handler): void
```
lukeraymonddowning, line 54: If we typehint the `Closure` with a structure, we could probably remove the later `phpstan-ignore-next-line` calls.

src/Concerns/Extendable.php
```diff
@@ -35,6 +40,71 @@ public static function hasExtend(string $name): bool
...
+    {
+        if (!self::hasPipes($name)) {
+            return [];
+        }
+
+        $pipes = [];
+        foreach (self::$pipes[$name] as $pipe) {
+            $pipe = $pipe->bindTo($context, $scope);
+
+            if($pipe instanceof Closure){
```
lukeraymonddowning, line 100: Hmmm, we do the same thing here. Why don't we do something slightly different:

I think there is already a support class in Pest for closures. We could have our own version of bind to that never returns false and instead throws an exception. That would remove all of these checks.

lukeraymonddowning: If we do that, this could be rewritten as 

return `array_map(fn ($pipe) => Closure::bindTo($context, $scope), self::$pipes[$name] ?? [])`

lukeraymonddowning (changes requested):

src/Concerns/Extendable.php
```diff
@@ -19,6 +21,9 @@ trait Extendable
      */
     private static array $extends = [];
 
+    /** @var array<string, array<Closure>> */
```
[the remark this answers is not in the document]
fabio-ivona: is this syntax correct?

```suggestion
    `/** @var array<string, array<Closure(Closure $next, mixed ...$arguments): void>> */`
```

phpstorm seems to break on it, but I can't see the issue

src/Concerns/Extendable.php
```diff
@@ -35,6 +40,71 @@ public static function hasExtend(string $name): bool
...
+            }
+
+            $next();
+        });
+    }
+
+    /**
+     * Checks if pipes are registered for a given expectation.
+     */
+    public static function hasPipes(string $name): bool
```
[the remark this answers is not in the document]
fabio-ivona: hmm, I think it is public because I used it outside this trait in a previous implementation, it can be private

```suggestion
    private static function hasPipes(string $name): bool
```

src/Concerns/Extendable.php
fabio-ivona: done!

src/Concerns/Extendable.php
fabio-ivona: done!

src/Concerns/Extendable.php
fabio-ivona: > I know this is currently a single use trait, but we have no dependence on $this->value in this trait other than here, so if this was placed in a class that didn't have $value as a property, then wouldn't it blow up?

Maybe I'm wrong, but here `$this` will refer to the binding that will be done when the pipe closure is executed (it is bound to Expectation on line 98 of this file), so it is not strictly tied to the trait 

don't know if I managed to explain myself, let me know if I'm not

lukeraymonddowning: Of course, my bad 👍

src/Concerns/Extendable.php
fabio-ivona: I like the idea, will work on it soon!

src/Concerns/Extendable.php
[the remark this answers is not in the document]
fabio-ivona: seems a phpstorm parser problem, I'll commit thas

src/Concerns/Extendable.php
fabio-ivona: it seems that, also here, proper typehinting solves the issue, so no need for a support class: 

```php
private function pipes(string $name, object $context, string $scope): array
{
    return array_map(fn(Closure $pipe) => $pipe->bindTo($context, $scope), self::$pipes[$name] ?? []);
}
```

and we can get rid of the `hasPipes` method

fabio-ivona:
moving to draft: new v2 refactors break my implementation. will work to fix this

fabio-ivona:
@nunomaduro this is ready for review

Issue #375, opened 2021-08-14
Pest output should match PHPUnit parity


Bilge:
Although Pest output is objectively better than the PHPUnit standard printer, it shouldn't remove features present in the underlying software. In particular, the following aspects are missing:

* Total memory consumption in summary output
* Total number of assertions run in summary output
* Software version in printer preamble (i.e. Pest version number plus PHPUnit version number)
* Random seed when `executionOrder="random"` (this is important to reproduce the same results using the same seed)

nunomaduro:
I think output the memory is a good idea - fell free to pull request that feature as an option like `--memory` or something. Regarding the Pest version + the PHPUnit version, you don't get that feature by using the `--version` option?

Bilge:
Passing `--version` shows the version information but doesn't run the tests. This is contrary to the behaviour of PHPUnit, which emits version information when running tests, without specifying any command line switches.

nunomaduro:
Regarding the `--version` no plans to change it's behavior. 

Regarding the `--memory` option no plans to add that information by default, still open for having that information, as opt-in using the `--memory` option. Fell free to pull request that.

Bilge:
Sad!

Bilge:
Sad!

And in prematurely closing this, not only do you fail to address the issues we discussed, but you completely fail to address the missing random seed and assertion counter (which actually does appear in the TeamCity output, just not the CLI).

I thought Pest was a good project, but you're an awful maintainer.

PR #433, opened 2021-11-03
feat: add `--memory` usage flag

owenvoke:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | ...

This is related to https://github.com/pestphp/pest/issues/375#issuecomment-903183321, and [this message in Discord](https://discord.com/channels/705126275232038944/710642595705126923/905233861615710299).

I've decided to go with outputting `MB`, but I guess we could make it so that it works it out the unit automatically (i.e. B, MB, GB) if that would be better?

Also, I just wanted someone to clarify if `memory_get_usage()` is the correct function to use for this? 🤔

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -79,6 +79,7 @@
         },
         "pest": {
             "plugins": [
+                "Pest\\Plugins\\Memory",
                 "Pest\\Plugins\\Coverage",
                 "Pest\\Plugins\\Init",
                 "Pest\\Plugins\\Version",
--- a/src/Plugins/Memory.php
+++ b/src/Plugins/Memory.php
@@ -1,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pest\Plugins;
+
+use Pest\Contracts\Plugins\AddsOutput;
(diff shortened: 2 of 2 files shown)


nunomaduro:
Can you target `next`?

owenvoke:
@nunomaduro, I've targeted `next`, looks like everything is failing now.

owenvoke:
@nunomaduro, is this all good? Just rebased it onto the latest so all tests are passing. 👍🏻

nunomaduro:
Yes - perfect buddy!

raftalks:
how does this work? I added flag but doesn't seem to work with latest version

owenvoke:
@raftalks, this is only available on Pest 2.x (which isn't stable yet). 👍🏻

owenvoke:
@raftalks, just for reference, I just extracted it into a package for Pest 1.x here:
https://github.com/owenvoke/pest-plugin-memory

It should be installable with:
```shell
composer require owenvoke/pest-plugin-memory --dev
```

---

Note, I haven't properly tested it as I'm away from my PC at home. 👍🏻

PR #438, opened 2021-11-15
Phpstan to 9

fabio-ivona:
This PR will upgrade phpstan level to 9

#### Disabled phpstan parallel mode

with phpstan on level 6, it triggers this error in `https://github.com/pestphp/pest/blob/master/src/OppositeExpectation.php`:

> Internal error: Internal error: Attempt to read property "name" on null in file pest/src/OppositeExpectation.php  
> Run PHPStan with --debug option and post the stack trace to:                                                                           
> https://github.com/phpstan/phpstan/issues/new?template=Bug_report.md                                                                   
> Child process error (exit code 1):

this error seems to happen only when phpstan is running in parallel mode (the default running mode)

can't reproduce the issue in --debug mode (as debug disables parallel by default)

disabling phpstan parallel mode seems to solve the issue, but it is only a workaround, can't get why this happens (maybe @canvural can help understanding what happens?)

#### Changed a parameter type

in `Datasets.php(52)` I changed the signature for `$with` from

```php
/**
 * @phpstan-param  Closure|iterable<int|string, mixed>|string $with
 */
public static function with(string $filename, string $description, Closure|iterable|string $with): void
 ```

to

```php
/**
 * @phpstan-param  array<Closure|iterable<int|string, mixed>|string> $with
 */
public static function with(string $filename, string $description, array $with): void
```

`$with` is assigned to an array of arrays and the method is called only in `TestCaseFactory.php(158)` and it receives an array


canvural:
This is really interesting. I don't have an idea right now.

Can you delete some code from `OppositeExpectation` and re-run it until it passes or you find a piece of code that triggers it?

canvural:
Ok I think it's because of `thecodingmachine/phpstan-strict-rules` If you remove it from `phpstan.neon` it should work.

fabio-ivona:
> Ok I think it's because of `thecodingmachine/phpstan-strict-rules` If you remove it from `phpstan.neon` it should work.

@canvural you're right!

removing thecodingmachine/phpstan-strict-rules will make it work

I'm trying to remove pieces of code to detect where it fails

fabio-ivona:
> This is really interesting. I don't have an idea right now.
> 
> Can you delete some code from `OppositeExpectation` and re-run it until it passes or you find a piece of code that triggers it?

@canvural I found it!

```php
try {
    $this->original->toHaveKey($key);
} catch (ExpectationFailedException) {
    continue;
}
```

the missing exception paramenter breaks phpstan

I've fixed it by adding the `$exception` where was missing:

```php
try {
    $this->original->toHaveKey($key);
} catch (ExpectationFailedException $exception) {
    continue;
}
```

and now it works again

any idea of why `thecodingmachine/phpstan-strict-rules` crashes when the paramenter is missing?

fabio-ivona:
@nunomaduro I've taken phpstan to lvl8, currently working for 9 :innocent:

nunomaduro:
@fabio-ivona wow, good job! I will take over this PR once is level 9.

fabio-ivona:
@nunomaduro we have level 9 :sunglasses: 

**notes**

- I cheated a bit, had to mark 12 `@phpstan-ignore` for parts I couldn't solve 
- `Expectations.php` had a lot of mixed issues on lvl9, that's because `$this->value` is a mixed, have added checks to enforce correct types, throwing an exception if not compliant
- `Datasets.php` was quite tough, lot of nested arrays processing and transformation made it difficult to follow types along the code, maybe a second look is worth for it

nunomaduro:
Excellent job @fabio-ivona

PR #443, opened 2021-11-27
Small Tweaks for Pest v2

lukeraymonddowning:
Just going through the Pest source code and made a few little tweaks. These are opinions so feel free to disagree.

The main change is moving the responsibility of generating the test *method* string to `TestCaseMethodFactory`. This serves to reduce the complexity of the large `evaluate` function to make it easier to understand.


nunomaduro:
More of those are welcomed.

PR #444, opened 2021-11-27
[2.x] Improved generics for higher order

lukeraymonddowning:
The purpose of the PR is to improve PHPStan generics on expectations. The goal is that as PhpStorm improves its support for generics, we should get better IDE autocompletion essentially free of charge.

I'm currently getting one issue, which is due to the `Each` type in `HigherOrderExpectation`. If anybody has an idea how we can get this sorted, let me know.

### DIFF
--- a/src/Each.php
+++ b/src/Each.php
@@ -7,9 +7,9 @@ namespace Pest;
 /**
  * @internal
  *
- * @template TEachValue
+ * @template TValue
  *
- * @mixin Expectation<TEachValue>
+ * @mixin Expectation<TValue>
  */
 final class Each
 {
@@ -18,7 +18,7 @@ final class Each
     /**
      * Creates an expectation on each item of the iterable "value".
      *
-     * @param Expectation<TEachValue> $original
+     * @param Expectation<TValue> $original
      */
     public function __construct(private Expectation $original)
     {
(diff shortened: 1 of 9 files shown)


src/Each.php
```diff
@@ -7,22 +7,31 @@
 /**
  * @internal
  *
- * @mixin Expectation
+ * @template TEachValue
```
nunomaduro, line 10: This should be TValue.

src/Each.php
```diff
@@ -7,22 +7,31 @@
...
      */
     public function __construct(private Expectation $original)
     {
-        // ..
     }
 
     /**
      * Creates a new expectation.
+     *
+     * @template TValue
```
nunomaduro, line 30: This should be TAndValue

src/Expectation.php
```diff
@@ -66,6 +66,8 @@ public function and(mixed $value): Expectation
 
     /**
      * Creates a new expectation with the decoded JSON value.
+     *
+     * @return self<mixed>
```
nunomaduro, line 70: The mixed is always an array?

nunomaduro (changes requested):

src/Expectation.php
lukeraymonddowning: Well, technically you could be json_decoding 'true', or 'false' or another string value, in which case it wouldn't return an array. Thoughts?

src/Each.php
lukeraymonddowning: Changed

src/Each.php
lukeraymonddowning: Changed

PR #449, opened 2021-12-08
[2.x] Adds support for chaining `hasNoExpectations` to the test method.

lukeraymonddowning:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

A feature that is often useful in PHPUnit is `expectNotToPerformAssertions`. This is handy when you have a method that is simply testing that a method can be called without throwing any exceptions or causing any errors, without being interested in the end result.

This PR adds a new `hasNoExpectations` method to the `TestCall` class, which allows us to support this functionality in a style more aligned with Pest:

```php
it('does not throw any errors when run', function () {
    SomeComplexService::run();
})->hasNoExpectations();
```

By adding this, we provide 2 main benefits:

1) We allow for visual consistency with other Pest `TestCall` methods, such as `throws`
2) We use the term `expectations` instead of `assertions`, which Pest users are less familiar with

Kind Regards,
Luke

### DIFF
--- a/src/PendingCalls/TestCall.php
+++ b/src/PendingCalls/TestCall.php
@@ -161,9 +161,11 @@ final class TestCall
     }
 
     /**
-     * Informs the test runner that no expectations happen in this test.
+     * Informs the test runner that no expectations happen in this test,
+     * and its purpose is simply to check whether the given code can
+     * be executed without throwing exceptions.
      */
-    public function hasNoExpectations(): TestCall
+    public function throwsNoExceptions(): TestCall
     {
         $this->testCaseMethod->proxies->add(Backtrace::file(), Backtrace::line(), 'expectNotToPerformAssertions', []);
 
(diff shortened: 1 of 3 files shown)


freekmurze:
Very nice! You could opt to name `throwsNoExceptions` for even more visual consistency with `throws`.

lukeraymonddowning:
@freekmurze I like that, but would it lead to devs mistakenly using it incorrectly? 

For example, a junior dev may put it on every test they expect to work. 

I may be wrong here, just my initial thoughts on it...

freekmurze:
@lukeraymonddowning I think if `throwsNoExceptions` has good documentation, we can steer people away from abusing it.

Should it be decided that `hasNoExpectations` would be the way forward, I'll happily use it too 🙂

JustSteveKing:
I think the term `hasNoExpectations` is a lot more descriptive to the action. This test has no expectations, it's clear the purpose of it whereas `throwsNoExceptions` while still clear could be confusing if read in a glance

fabio-ivona:
+1 for `throwsNoExceptions`

PR #458, opened 2022-01-17
Add support for invokable classes in higher order expectations

sebastiandittrich:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | none

## Problem this PR tries to solve
Currently, only closures can be passed to higher order expectations, therefore it is impossible to pass invokable classes into the `test()->expect(...)` method.

## Solution
The naive approach of allowing all callables brings even more problems with it (see #344). This PR tries to allow invokable classes implementing a specific interface provided by pest to get executed by the higher order test expectation call. This way, the callables executed by higher order test expectations are more flexible instead of just accepting plain closures but do not unexpectedly evaluate strings or other callable classes.

## Example
```php
class HigherOrderCallable implements Pest\Contracts\IsHigherOrderCallable
{
    public function __invoke()
    {
        return 'foo';
    }
}

test('higher order callable')
    ->expect(new HigherOrderCallable())
    ->toBe('foo')
```

## Use case
I am currently building a function, that takes care of setting up some complex database situations in a laravel app that should then be tested. Accessing the generated database entries via the Eloquent models is done via a global `situation()` helper function, which returns an object containing all created models. Currently, I am accesing them in higher order expectations like this:

```php
test('has status created')
    ->expect(fn() => situation()->user)
    ->status()->toBe('created')
```

To make this a little bit easier, I wanted the `situation()` helper to return a kind of higher order function builder itself if it was called outside of a running testsuite, in a way that the previous code could be rewritten to: 

```php
test('has status created')
    ->expect(situation()->user) // Note the missing 'fn() => ' in this line
    ->status()->toBe('created')
```

With this approach, `situation()->user` is returning an invokable class, which doesn't get evaluated by pest currently. 

Happy to hear your feedback on this.

### DIFF
--- a/src/Contracts/IsHigherOrderCallable.php
+++ b/src/Contracts/IsHigherOrderCallable.php
@@ -1,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pest\Contracts;
+
+/**
+ * Interface used to mark a class callable by higher order expectations.
+ */
+interface IsHigherOrderCallable
+{
+    public function __invoke(): mixed;
+}
--- a/src/Support/HigherOrderCallables.php
+++ b/src/Support/HigherOrderCallables.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Pest\Support;
 
 use Closure;
+use Pest\Contracts\IsHigherOrderCallable;
 use Pest\Expectation;
 
 /**
@@ -31,6 +32,10 @@ final class HigherOrderCallables
      */
     public function expect(mixed $value): Expectation
     {
+        if ($value instanceof IsHigherOrderCallable) {
+            $value = fn (...$data) => $value(...$data);
+        }
+
         /** @var TValue $value */
         $value = $value instanceof Closure ? Reflection::bindCallableWithData($value) : $value;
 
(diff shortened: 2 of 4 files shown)


lukeraymonddowning:
@sebastiandittrich I appreciate the effort here, but for the sake of `fn () =>` over having to implement an interface I'm not 100% sure it saves all that much personally. What are your thoughts?

sebastiandittrich:
@lukeraymonddowning You are right, this saves only a few characters. But since the interface implementation part is mainly for package developers who want to better integrate their packages into pest, the character saving doesn't mean any extra effort to the end user. It just saves them the trouble to write `fn () =>` in their expectations and helps them make their tests more expressive.

Additionally, at least for me it was quite confusing why pest only accepts closures as arguments in the higher order `expect()` function. With this PR, pest will get more flexible in this respect without any drawbacks in case you don't need this flexibility and without any breaking changes.

So even if this saves "only" 6 characters per expectation, I think it's worth it since it comes at no cost (at least for the end user).

nunomaduro:
No plans for this. Thank you for the pull request.

Issue #452, opened 2021-12-19
High order expectation fail with json on object


amenophis:
Hi,
I'm trying to work on a symfony plugin.
I am unable to use high order expectation with json on object.

I created a failing test in my fork: https://github.com/amenophis/pest/blob/json-high-order/tests/Features/Expect/HigherOrder/methodsJson.php

The error is:

```
  • Tests\Features\Expect\HigherOrder\methodsJson > it can use json in method high order expectation
  Failed asserting that null is identical to 1.

  at src/Expectation.php:275
    271▕      * @param mixed $expected
    272▕      */
    273▕     public function toBe($expected): Expectation
    274▕     {
  ➜ 275▕         Assert::assertSame($expected, $this->value);
    276▕ 
    277▕         return $this;
    278▕     }
    279▕ 

  1   src/HigherOrderExpectation.php:137
      Pest\Expectation::toBe()

  2   src/HigherOrderExpectation.php:92
      Pest\HigherOrderExpectation::performAssertion()


```

Maybe i'm doing something bad.
Can i have your opinion ?

Thanks

amenophis:
Hello, 
Can you confirm it is a normal behavior or not ?

Thanks

nunomaduro:
cc @dansysanalyst - what were your conclusions on this?

dansysanalyst:
@nunomaduro on 21.12.2021 I had asked on the group regarding arrays and Higher Order and you replied `checking`.

Did I miss any further discussion for this topic?

![IMAGE 2022-01-03 15:00:23](https://user-images.githubusercontent.com/79267265/147939351-f1ba9ed9-eea0-4ea8-a20f-72c4520f52c3.jpg)
 `Failed asserting that null is identical to 'amenophis'.`

amenophis:
@dansysanalyst @nunomaduro Do you have news about this issue ??

dansysanalyst:
Just for investigation, I have returned an array instead of using the `json()` modifier. 

```php
//...
public function getArray(): array
{
    return json_decode($this->content, true);
}
```


```php
it('can access array', function () {
    expect(new Response())
        ->getArray()
        ->id->toBe(1)
        ->username->toBe('amenophis')
        ->email->toBe('amenophis@leherpeur.net');
});
```

Higher Order fails to read further array properties: 
`Failed asserting that null is identical to 'amenophis'.`

amenophis:
Can I help on this topic ?

[...]

PR #461, opened 2022-01-21
Fixes `json` in Higher Order Expectations

lukeraymonddowning:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #452

@dansysanalyst and I have been looking into possible solutions for this issue. We have decided on the following:

```php
it('can call the json method in higher order expectations', function () {
    expect(new HasMethods())
        ->getJsonContent()->json()->id->toBe(1)->toBeGreaterThan(0)
        ->getJsonContent()->json()->email->toBe('amenophis@leherpeur.net');
});
```

Note that at the end of an expectation chain with `json`, the scope will now completely reset, which is in line with how other expectation methods work. Previously, `json` in Higher Order Expectations wasn't working at all.

There was an alternative, which would allow for staying in the JSON object for more expectations on it, but that would have been out of place with how the rest of the expectation API works.

### DIFF
--- a/src/HigherOrderExpectation.php
+++ b/src/HigherOrderExpectation.php
@@ -77,6 +77,11 @@ final class HigherOrderExpectation
         return $this->expect($value);
     }
 
+    public function json(): HigherOrderExpectation
+    {
+        return new self($this->original, $this->expectation->json()->value);
+    }
+
     /**
      * Dynamically calls methods on the class with the given arguments.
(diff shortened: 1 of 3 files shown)


nunomaduro:
@lukeraymonddowning You will need to create another pull request for Pest v2. Alright?

PR #478, opened 2022-02-10
[2.x] Dataset preprocessing

### LINKED ISSUE #521: [feature] Public access method to Dataset
Hello,

In some tests I needed to access a `dataset` into a `test` for some processing before testing.

I used 

```php
foreach (\Pest\Datasets::get('my_dataset') as $data) {
    // ...
}
```

With success but with IDE warning about accessing internal function : `Class Datasets is marked as @internal`.

It could be good to allow users to access `dataset` "legally" with a function.

fabio-ivona:
**disclaimer:** this PR is a 4 hands (and 2 brains) work between me and @lukeraymonddowning , a great part of credits goes to his ideas

This PR wants to address the needs expressed in #319 and #470 and basically the generation of powerful context-dependant datasets

To clarify the improvements this PR brings, let's suppose we have a dataset like this:

```php
dataset('my dataset', function(int $count = 3, bool $uppercase = false){
    $dataset = Collection::make([
        'key1' => ['name' => 'foo', 'title' => 'mr'],
        'key2' => ['name' => 'bar', 'title' => 'mrs', 'age' => 21],
        'key3' => ['name' => 'baz', 'title' => 'ms'],
    ]);

    if($uppercase){
           $dataset = dataset->map(function($value){
                  $vaue['name'] = strtoupper($value['name']);
          });
    }

    return $dataset->take($count)->toArray();
});
```

## dataset generation closure parameters

with the implementation we set up, this PR will allow to pass parameters to the dataset generation closure:

```php

test('foo', function(){
    //...
})->with('my dataset', 2)

/* Will iterate over this dataset:
 * [
 *     'key1' => ['name' => 'foo', 'title' => 'mr'],
 *     'key2' => ['name' => 'bar', 'title' => 'mrs', 'age' => 21],
 * ]
 */

test('foo', function(){
    //...
})
->with('my dataset', uppercase: true);

/* [
 *     'key1' => ['name' => 'foo', 'title' => 'mr'],
 *     'key2' => ['name' => 'bar', 'title' => 'mrs', 'age' => 21],
 *     'key3' => ['name' => 'baz', 'title' => 'ms'],
 * ]
 */

```

## dataset builtin preprocessing functions

with the ability to use named parameters as arguments for the `->with()` function, a toolset of preprocessing functions can be applied to the dataset before starting iterate over it:

### only

```php
->with('my dataset', only: ['key1', 'key2']) 
/* [
 *     'key1' => ['name' => 'foo', 'title' => 'mr'],
 *     'key2' => [['name' => 'bar', 'title' => 'mrs', 'age' => 21],
 * ]
 */
```

### except

```php
->with('my dataset', except: 'key2') 
/* [
 *     'key1' => ['name' => 'foo', 'title' => 'mr'],
 *     'key3' => ['name' => 'baz', 'title' => 'ms'],
 * ]
 */
```

### pluck

```php
->with('my dataset', pluck: 'name') 
/* [
 *     'foo'
 *     'bar'
 *     'baz'
 * ]
 */
```

### map

```php
->with('my dataset', map: fn($element) => User::factory(['name' => $element['name'])->create())
/* [
 *     [User],
 *     [User],
 *     [User],
 * ]
 */
```

## DatasetRepository code optimization

along with these improvements, this PR will introduce a refactoring for the `DatasetRepository` code, extracting a lot of dataset resolving logic to a dedicated class and, thus, simplifying the typing system of the repository, i.e. the `$withs` parameters changes from:

```php
/**
 * Holds the withs.
 *
 * @var array<array<string, Closure|iterable<int|string, mixed>|string>>
 */
 private static array $withs = [];
```

to

```php
/**
 * Holds the withs.
 *
 * @var array<string, array<string, TestCaseDataset>>
 */
private static array $withs = [];
```

and lot of more cleanup

### DIFF
--- a/src/Datasets/PluckPreprocessor.php
+++ b/src/Datasets/PluckPreprocessor.php
@@ -23,7 +23,7 @@ final class PluckPreprocessor implements Preprocessor
             throw new InvalidArgumentException('The pluck preprocessor should be given a string.');
         }
 
-        // @phpstan-ignore-next-line
-        return array_map(fn (array|object $values) => $values[$value], $dataset);
+        /* @phpstan-ignore-next-line */
+        return array_map(fn (array|object $values) => is_array($values) ? $values[$value] : $values->{$value}, $dataset);
     }
 }
--- a/src/Datasets/PluckPreprocessor.php
+++ b/src/Datasets/PluckPreprocessor.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pest\Datasets;
+
+use InvalidArgumentException;
+use Pest\Contracts\Dataset\Preprocessor;
+
+/**
+ * @internal
+ */
+final class PluckPreprocessor implements Preprocessor
+{
+    public function getKey(): string
+    {
+        return 'pluck';
+    }
+
+    public function process(array $dataset, mixed $value): array
+    {
+        if (!is_string($value)) {
+            throw new InvalidArgumentException('The pluck preprocessor should be given a string.');
+        }
+
+        // @phpstan-ignore-next-line
+        return array_map(fn (array|object $values) => $values[$value], $dataset);
+    }
+}
--- a/src/Datasets/PreprocessorRepository.php
+++ b/src/Datasets/PreprocessorRepository.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pest\Datasets;
+
+use InvalidArgumentException;
+use Pest\Contracts\Dataset\Preprocessor;
+
+/**
+ * @internal
+ */
+final class PreprocessorRepository
+{
+    /**
+     * @var array<Preprocessor>
+     */
+    private array $preprocessors;
+
+    public function __construct()
+    {
+        $this->preprocessors = [
+            new ExceptPreprocessor(),
+            new MapPreprocessor(),
+            new OnlyPreprocessor(),
+            new PluckPreprocessor(),
+        ];
+    }
+
+    public function has(string|int $key): bool
+    {
+        return count($this->itemsForKey($key)) > 0;
+    }
+
+    public function get(string|int $key): Preprocessor
+    {
+        if (!$this->has($key)) {
+            throw new InvalidArgumentException("[$key] is not a supported preprocessor.");
+        }
+
+        return $this->itemsForKey($key)[0];
+    }
+
+    /**
+     * @return array<Preprocessor>
+     */
+    private function itemsForKey(string|int $key): array
+    {
+        return array_values(array_filter(
+            $this->preprocessors,
+            fn (Preprocessor $preprocessor) => $preprocessor->getKey() === $key
+        ));
+    }
+}
--- a/src/Factories/TestCaseMethodFactory.php
+++ b/src/Factories/TestCaseMethodFactory.php
@@ -5,12 +5,12 @@ declare(strict_types=1);
 namespace Pest\Factories;
 
 use Closure;
-use Pest\Dataset;
 use Pest\Exceptions\ShouldNotHappen;
 use Pest\Factories\Concerns\HigherOrderable;
 use Pest\Plugins\Retry;
 use Pest\Repositories\DatasetsRepository;
 use Pest\Support\Str;
+use Pest\TestCaseDataset;
 use Pest\TestSuite;
 use PHPUnit\Framework\Assert;
 use PHPUnit\Framework\TestCase;
@@ -30,7 +30,7 @@ final class TestCaseMethodFactory
     /**
      * The Test Case Dataset, if any.
      *
-     * @var array<Dataset>
+     * @var array<TestCaseDataset>
      */
     public array $datasets = [];
 
--- a/src/Factories/TestCaseMethodFactory.php
+++ b/src/Factories/TestCaseMethodFactory.php
@@ -34,6 +34,13 @@ final class TestCaseMethodFactory
     public array $datasets = [];
 
     /**
+     * The Test Case Dataset's parameters, if any.
+     *
+     * @var array<array<mixed>>
+     */
+    public array $datasetParameters = [];
+
+    /**
      * The Test Case depends, if any.
      *
      * @var array<int, string>
@@ -159,7 +166,7 @@ final class TestCaseMethodFactory
      */
     private function buildDatasetForEvaluation(string $methodName, string $dataProviderName): string
     {
-        DatasetsRepository::with($this->filename, $methodName, $this->datasets);
+        DatasetsRepository::with($this->filename, $methodName, $this->datasets, $this->datasetParameters);
 
         return <<<EOF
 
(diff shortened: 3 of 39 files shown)


src/Factories/TestCaseMethodFactory.php
```diff
@@ -29,7 +30,7 @@ final class TestCaseMethodFactory
     /**
      * The Test Case Dataset, if any.
      *
-     * @var array<Closure|iterable<int|string, mixed>|string>
+     * @var array<TestCaseDataset>
```
lukeraymonddowning, line 33: Is it a `TestCaseDataSet` or a `TestDataSet`? I would say that given it only pertains to a single test, it should be the latter?

lukeraymonddowning:
@fabio-ivona I've opted to move the preprocess functions to their own class. The idea being that we could support plugins registering their own preprocessors, or a user implementing a global preprocessor of their own.

src/Datasets/PluckPreprocessor.php
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    public function process(array $dataset, mixed $value): array
+    {
+        if (!is_string($value)) {
+            throw new InvalidArgumentException('The pluck preprocessor should be given a string.');
+        }
+
+        // @phpstan-ignore-next-line
+        return array_map(fn (array|object $values) => $values[$value], $dataset);
```
lukeraymonddowning, line 27: Would be good to have this work on objects and arrays.

src/Datasets/PreprocessorRepository.php
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Pest\Datasets;
+
+use InvalidArgumentException;
+use Pest\Contracts\Dataset\Preprocessor;
+
+/**
+ * @internal
+ */
+final class PreprocessorRepository
```
lukeraymonddowning, line 13: My eventual idea is to register this in the Pest container, then inject it into the `TestCaseDataSet`

lukeraymonddowning (commented):
Not finished with review/changes. But done for tonight :-)

src/Factories/TestCaseMethodFactory.php
fabio-ivona: Yeah, you're right, I think we should rename the class

src/Datasets/PluckPreprocessor.php
fabio-ivona: hmm, I think the dataset resolves to an array before calling preprocess functions, is there any way it can be an object?

also I guess you mean `$values` to be `$dataset`?

lukeraymonddowning: Doesn't this run on each item in the dataset? So for example, an array of Eloquent models?

src/TestCaseDataset.php
```diff
@@ -0,0 +1,150 @@
...
+    {
+        return is_string($this->dataset);
+    }
+
+    /**
+     * @return array<int|string, mixed>
+     */
+    public function resolve(): array
+    {
+        $this->resolveFromString();
```
lukeraymonddowning, line 40: I think this might be more clear if each of these were a pure method and we return the new dataset each time rather than overriding the `dataset` property.

src/Datasets/PluckPreprocessor.php
fabio-ivona: Hmm, no, pluck runs on each dataset and extracts only the key you want

that's the same behaviour of collection()->pluck(), it iterates over collection items and extract a key

or am I missing something?

fabio-ivona: so, to clarify my answer:

```php
$dataset = [
    ['first_name' => 'Fabio', 'last_name' => 'Ivona'],
    ['first_name' => 'Luke', 'last_name' => 'Downing'],
];


 array_map(fn (array|object $values) => $values[$value], $dataset); 
 // $values =  ['first_name' => 'Fabio', 'last_name' => 'Ivona'] on the first iteration
 // $values =  ['first_name' => 'Luke', 'last_name' => 'Downing'] on the second iteration

File: src/TestCaseDataset.php (hunk 4 above)
Comment (fabio-ivona) [reply src/TestCaseDataset.php:40]: I implemented the way you said initially, will check why I changed to this one, don't have my IDE in front of me ATM

File: src/Datasets/PluckPreprocessor.php (hunk 2 above)
Comment (lukeraymonddowning) [reply src/Datasets/PluckPreprocessor.php:27]: But a dataset could also be 

```php
$dataset = [
    new User(['first_name' => 'Fabio', 'last_name' => 'Ivona']),
    new Use(['first_name' => 'Luke', 'last_name' => 'Downing']),
];
```

lukeraymonddowning: In which case I'd want it to get the property `first_name` from the `User` object.

fabio-ivona: hmm, you're right, was focusing only on array[] datasets and missed single values one

let's change that value to array|object

src/TestCaseDataset.php
fabio-ivona: The main issue was in `getPreprocessParameters()` 

in that method, actual closure-dataset parameters are extracted from the list using reflection

but at the moment of calling that method, your dataset has already been resolved to an array

now I refactored to use pure methods that return a new dataset, but for actual parameters detection the method still uses the original dataset (resolving it from string, if needed)

this should work, let me know if something is not optimal or you have a better idea

src/Factories/TestCaseMethodFactory.php
fabio-ivona: Refactored to `TestDataset`

src/Datasets/PreprocessorRepository.php
fabio-ivona: I think this will be the greatest feature of this PR

src/Datasets/PluckPreprocessor.php
lukeraymonddowning, line 27: Won't this still fail on an object because you can't access a standard object property with `[]` syntax. We would have to implement something similar to Laravel's `data_get` or (perhaps simpler) use a ternary to decide how to access the property (`[]` or `->`).

lukeraymonddowning (approved):
Looking ace @fabio-ivona 

One last small change and I think we're good to merge

src/Datasets/PluckPreprocessor.php
fabio-ivona: Maybe a `data_get` is a bit overkill for this case alone. Do any other uses come in you mind for it to be worth to implement in pest?

fabio-ivona: meanwhile I added a ternary implementation and a test to point out this behavior

it seems fine to me, let me know :wink:

fabio-ivona:
@nunomaduro what's your tought on this one? 

I'm planning to update the code to match current v2, as it has been left behind, but it is worth doing it only if the feature has a real value

lukeraymonddowning:
@fabio-ivona I think we should scrap this for now. I think there is probably a better way to handle this with better IDE support.

Issue #483, opened 2022-02-17
Test passes when expecting an undefined exception


fran-f:
Hello there 👋 
It seems that Pest 1.21.0 lets a test pass if it throws and expects an exception that *does not exist*.
I can reproduce the issue with the examples below, on both Windows/PHP 7.4.23 and Ubuntu/PHP 7.4.3.

`ProblemTest.php`
```php
<?php
it("throws the exception", function() {
    expect(function() { throw new DoesNotExistException; })
        ->toThrow(DoesNotExistException::class);
});
```

`ControlTest.php`
```php
<?php
class ControlTest extends PHPUnit\Framework\TestCase
{
    public function testExceptionDoesNotExist()
    {
        $this->expectException(DoesNotExistException::class);
        throw new DoesNotExistException;
    }
}
```

Output of `vendor\bin\pest`
```
   FAIL  ControlTest
  ⨯ exception does not exist

   PASS  Tests\ProblemTest
  ✓ it throws the exception

  ---

  • ControlTest > exception does not exist
   PHPUnit\Framework\Exception

  Class DoesNotExistException does not exist

  Tests:  1 failed, 1 passed
  Time:   0.03s
```

The Pest test passes, while the "control" PHPUnit test fails. I would expect Pest to behave as PHPUnit 🤔

[edit] Possibly connected to this check?

https://github.com/pestphp/pest/blob/e64b6fe924df553006db2cdeeee7713f5859f7ac/src/Mixins/Expectation.php#L826,L830

fabio-ivona:
Hi @fran-f thanks for pointing it out!

we are working on a fix for this

fabio-ivona:
@fran-f  the issue is fixed by #487

fran-f:
Thank you @fabio-ivona !

PR #484, opened 2022-02-18
Fix toThrow expectation passing when exception doesn't exist

fabio-ivona:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| Fixed tickets | #483 

This PR will fix tests passing when expecting an exception that does not exist

as pointed out in #483 a test like this will pass without revealing the missing Class error:

```php
it("throws the exception", function() {
    expect(function() { throw new DoesNotExistException; })
        ->toThrow(DoesNotExistException::class);
});
```

this is dued to the `->toThrow` expectation checking if its first parameter is a message and not a class:

```php
//Expectation.php lines from 915
if (!class_exists($exception)) {
    Assert::assertStringContainsString($exception, $e->getMessage());

    return $this;
}
```
in this case:

`$exception` will be `DoesNotExistException`

the class not found error message contained in `$e->getMessage` will be `Class "DoesNotExistException" not found`

and the Assertion will evaluate true

this PR fix will add a check to avoid such a behaviour and rethrow the class not found error:

```php
//Mixin\Expectation.php lines from 828
if (!class_exists($exception)) {
    if ($e instanceof Error && $e->getMessage() === "Class \"$exception\" not found") {
        throw $e;
    }

    Assert::assertStringContainsString($exception, $e->getMessage());

    return $this;
}
```

**note** this should be fixed in v1.x as well, will open a PR for it as soon as we review and merge this

### DIFF
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -6,6 +6,7 @@ namespace Pest\Mixins;
 
 use BadMethodCallException;
 use Closure;
+use Error;
 use InvalidArgumentException;
 use Pest\Exceptions\InvalidExpectationValue;
 use Pest\Support\Arr;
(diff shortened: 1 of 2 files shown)


owenvoke (approved):

fabio-ivona (commented):
commented wrong PR

Issue #32, opened 2020-05-26
Support for PHPUnit @covers notation


Joucke:
PHPUnit supports coverage [annotations](https://phpunit.readthedocs.io/en/9.1/annotations.html), both test class wide and per test.

```
/**
 * @test 
 * @covers App\Model
 */

```

This will limit coverage calculation to only this class (or method, if specified, etc). 

Is it possible to do this in Pest?

owenvoke:
This would be quite interesting to have with methods for `covers(... $class)` and `coversNothing()`.

I've taken a quick look and I'm not sure if this could be implemented in a similar way to the [`group()` call](https://github.com/pestphp/pest/blob/master/src/PendingObjects/TestCall.php#L81).

We may be able to look at [this method in PHPUnit](https://github.com/sebastianbergmann/phpunit/blob/3da2ced818c857826dbb800ca963ff3dc7f7ea9b/src/Util/Test.php#L540) for more information.

nunomaduro:
@Joucke Can you try to develop this? Should be just adding something like this: https://github.com/pestphp/pest/pull/103/files#diff-4726f6affac796013ea01b71b2a6aae7R90.

Joucke:
Sorry, I've been looking at this - but I really don't understand how either pest or phpunit work internally. Thought it'd be a nice challenge, but I have no idea where to start.

DannyvdSluijs:
If this remains untouched I could take a stab at this somewhere in the near future. No promises though.
It seems only some glueing between Pest and the PHPUnit internals are needed. No need to mimic the annotations itself, right?

owenvoke:
@DannyvdSluijs, yep. I tried to get the 2 working together and passing the classes to PHPUnit, but hit a few issues which I didn't have time to overcome. 😬 Good luck if you do have a go, that'd be great!

DannyvdSluijs:
PHPUnit handles the `@covers` differently then a `@depends`. As with an `@depend` the meta data is set on the `Testcase` which was quite easily made possible in Pest as hinted by @nunomaduro. But the test coverage is part of the  `Testresult` class which is [invoking](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/TestResult.php#L843) the `PHPUnit\Util::getLinesToBeCovered()` directly as a static function in its `run` method. This has the effect that in its current form phpunit is wired to only support PHP DocBlock annotations on the test class and method, something which would not work for Pest.

[...]

PR #492, opened 2022-03-05
[2.x] Add `covers` attribute

danilopolani:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #32 

This PR _tries_ to implement the `@covers` annotation. Actually, following the issue https://github.com/sebastianbergmann/phpunit/issues/4502 , I decided to implement it as a PHP Attribute since the annotations will be deprecated in PHPUnit 11.

I haven't tested this yet because I started working on it on Windows and 90% of the commands do not work, will come back on this in the next few days.

Because of the logic of Pest (where final user does not have a class), I think it's not possible to implement the `@coversDefaultClass`, but that's not a big deal since the other methods works just fine without it.

**Usage**

```php
test('covers class')
    ->expect(true)->toBeTrue()
    ->coversClass(Foo::class, Bar::class); // Same as `@covers MyClass`

test('covers global function')
    ->expect(true)->toBeTrue()
    ->coversFunction('globalFunction', 'anotherFunction'); // Same as `@covers ::functionName`

test('covers nothing')
    ->expect(true)->toBeTrue()
    ->coversNothing(); // Same as `@coversNothing`
```

**Todo**

- [x] Move attributes above class and not method (`only allowed on classes; @covers was allowed on methods but it does not make sense to mix different coverage targets in a test case class`)
- [x] Check FQN when passing a class name string (when it's passed to the Attribute)
- [x] Lint
- [x] Add tests
- [x] Add smart "covers()" that accepts both classes and functions and checks if they exists (?)

### DIFF
--- a/src/Factories/Attributes/Attribute.php
+++ b/src/Factories/Attributes/Attribute.php
@@ -1,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pest\Factories\Attributes;
+
+/**
+ * @internal
+ */
+abstract class Attribute
+{
+    /**
+     * Determine if the attribute should be placed above the classe instead of above the method.
+     *
+     * @var bool
+     */
+    public const ABOVE_CLASS = false;
+}
--- a/src/Factories/Attributes/Attribute.php
+++ b/src/Factories/Attributes/Attribute.php
@@ -10,7 +10,7 @@ namespace Pest\Factories\Attributes;
 abstract class Attribute
 {
     /**
-     * Determine if the attribute should be placed above the classe instead of above the method.
+     * Determine if the attribute should be placed above the class instead of above the method.
      *
      * @var bool
      */
--- a/src/Factories/TestCaseFactory.php
+++ b/src/Factories/TestCaseFactory.php
@@ -38,7 +38,7 @@ final class TestCaseFactory
     /**
      * The list of annotations.
      *
-     * @var array<int, class-string>
+     * @var array<int, \Pest\Factories\Attributes\Attribute>
      */
     private static array $attributes = [
         Attributes\Covers::class,
@@ -155,6 +155,19 @@ final class TestCaseFactory
             $methods
         ));
 
+        $classAttributes = [];
+
+        foreach (self::$attributes as $attribute) {
+            if ($attribute::ABOVE_CLASS) {
+                /** @phpstan-ignore-next-line */
+                $classAttributes = (new $attribute())->__invoke($this, $classAttributes);
+            }
+        }
+
+        $classAttributes = implode('', array_map(
+            static fn ($attribute) => sprintf("\n                 %s", $attribute), $classAttributes,
+        ));
+
         try {
             eval("
                 namespace $namespace;
@@ -162,6 +175,7 @@ final class TestCaseFactory
                 use Pest\Repositories\DatasetsRepository as __PestDatasets;
                 use Pest\TestSuite as __PestTestSuite;
 
+                $classAttributes
                 final class $className extends $baseClass implements $hasPrintableTestCaseClassFQN {
                     $traitsCode
 
--- a/src/Factories/TestCaseFactory.php
+++ b/src/Factories/TestCaseFactory.php
@@ -166,7 +166,7 @@ final class TestCaseFactory
         }
 
         $classAttributes = implode('', array_map(
-            static fn ($attribute) => sprintf("\n                 %s", $attribute),
+            static fn (string $attribute) => sprintf("\n                 %s", $attribute),
             array_unique($classAttributes),
         ));
 
--- a/src/Factories/TestCaseFactory.php
+++ b/src/Factories/TestCaseFactory.php
@@ -36,6 +36,15 @@ final class TestCaseFactory
     ];
 
     /**
+     * The list of annotations.
+     *
+     * @var array<int, class-string>
+     */
+    private static array $attributes = [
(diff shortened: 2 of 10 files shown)


nunomaduro:
@danilopolani Looking very interesting so far!

src/Factories/Attributes/Attribute.php
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Pest\Factories\Attributes;
+
+/**
+ * @internal
+ */
+abstract class Attribute
+{
+    /**
+     * Determine if the attribute should be placed above the classe instead of above the method.
```
owenvoke, line 13: Minor typo
```suggestion
     * Determine if the attribute should be placed above the class instead of above the method.
```

src/Factories/TestCaseFactory.php
```diff
@@ -35,6 +35,15 @@ final class TestCaseFactory
         Annotations\Groups::class,
     ];
 
+    /**
+     * The list of annotations.
+     *
+     * @var array<int, class-string>
```
owenvoke, line 41: Can probably enforce that it's a class string of type `Attributes\Attribute` 🤷🏻
```suggestion
     * @var array<int, class-string<\Pest\Factories\Attributes\Attribute>>
```

src/PendingCalls/TestCall.php
```diff
@@ -168,6 +172,63 @@ public function skip(Closure|bool|string $conditionOrMessage = true, string $mes
...
+    {
+        foreach ($classesOrFunctions as $classOrFunction) {
+            $isClass  = class_exists($classOrFunction);
+            $isMethod = function_exists($classOrFunction);
+
+            if (!$isClass && !$isMethod) {
+                throw new InvalidArgumentException(sprintf('No class or method named "%s" has been found.', $classOrFunction));
+            }
+
+            if ($isClass) {
```
owenvoke, line 188: Nice! Definitely like having the single `covers()` method as an option. 👍🏻

Wonder if it's worth doing some early `continue` uses in here, not sure though.

src/PendingCalls/TestCall.php
```diff
@@ -168,6 +172,63 @@ public function skip(Closure|bool|string $conditionOrMessage = true, string $mes
...
+
+        return $this;
+    }
+
+    /**
+     * Sets that the current test covers nothing.
+     */
+    public function coversNothing(): TestCall
+    {
+        $this->testCaseMethod->covers[] = new CoversNothing();
```
owenvoke, line 227: Should this be the following so that you can't use a `coversNothing()` with a `covers()`? Not sure though, this should be fine. 👍🏻
```suggestion
        $this->testCaseMethod->covers = [new CoversNothing()];
```

owenvoke (commented):
Quick few things, but looks great to me! Thanks for adding the tests too! 🥳

lukeraymonddowning:
Brilliant work here 👌

src/Factories/TestCaseFactory.php
danilopolani: Didn't know this was possible, thank you!

src/PendingCalls/TestCall.php
danilopolani: Mhh  you're right, I mean, you could, but it would be useless haha.  
Reading a second time the issue on PHPUnit I noticed that `@coversNothing` could be in **both** classes and methods and I think it would be better, for us, to always consider it as "per-method" to avoid undesired behaviours where the whole tests file covers nothing.

Now the issue is that PHPUnit reads by priority comments and attributes:

> PHPUnit 10 will first look for metadata in attributes before it looks at comments

For this reason if we use `#[CoversNothing]` then the `@test` comment will be ignored and the test is not executed. IMHO we have three paths:

1. Use `@coversNothing` as comment instead of attribute;
2. Transform `@test` comment into `#[Test]` attribute to use `#[CoversNothing]`, although a bit out of scope;
3. Use, for now, `@coversNothing`, when this is merged I'll open another PR to transform `@test` and `@coversNothing` into their respective Attributes (now or later, I mean).

What do you think should we do?

src/PendingCalls/TestCall.php
danilopolani: Got your point, you tell me what would you prefer to see, at the end it's the same 😄

owenvoke: I think it's good to leave as is. 👍🏻

danilopolani:
@owenvoke thank you for the review! At the end I've moved `coversNothing` to method annotations and I've left the "methods attribute" code generation as well, although right now not used, because it will be needed in the future.

Let me know if you prefer me to remove it because not needed or if it's fine :)

nunomaduro:
Thank you for this! We may reach in the future, asking for docs about this. ✅

danilopolani:
> Thank you for this! We may reach in the future, asking for docs about this. ✅

Thank you Nuno! No problem, if you want I can already work on the PR for the docs 👀

nunomaduro:
@danilopolani Can you check what static types errors we have on master, and pull request a fix? Or they are not related to your PR?

danilopolani:
> @danilopolani Can you check what static types errors we have on master, and pull request a fix? Or they are not related to your PR?

Yep sorry, I forgot to run `test:types` but only ran lint and tests 🤦 Sent a new PR -> https://github.com/pestphp/pest/pull/498

Issue #494, opened 2022-03-07
Pest.php not available with different test folder name


jsonUK:
### Problem

`tests/Pest.php` does not autoload if your tests folder is different to `tests/`.

ie. I changed the phpunit.xml.dist folder to be `php-tests` instead.

### Steps to Reproduce
I was able to test this was true by:
- Making sure the function I want to access is present in `Pest.php`
- Remove local function in test file - re-test. Tests fail
- Move the `Pest.php` file to `tests/Pest.php` - re-test. Function is present again and the test passes.

### Proposed Solution

I believe the path for `Pest.php` just needs to be dictated by the phpunit.xml file, so that it loaded from the correct location.

Btw, first time using Pest and really happy with it so far, thank you 👍🎉

danilopolani:
Hey @jsonUK ,
can you try running your test suite with the argument `--test-directory` and see if that works? E.g.

```bash
./vendor/bin/pest --test-directory=php-tests
```

jsonUK:
Hi - with the test I have setup, the tests turn green after trying that command, though I would say its not ideal to override from the CLI.

danilopolani:
> Hi - with the test I have setup, the tests turn green after trying that command, though I would say its not ideal to override from the CLI.

It make sense that it should be read from the `phpunit.xml` configuration too before the default "tests" folder.  
PHPUnit already has a buch of useful methods to read and parse the configuration from a `phpunit.xml.*` file, but they are all private, but they can be a nice starting point.

Maybe, if @nunomaduro , @lukeraymonddowning , @fabio-ivona and other maintainers approve this request, should it address Pest v2 since it could lead to undesired effects? E.g. a phpunit configuration where tests are not inside the `tests` folder, although not a big issue for users to fix probably.

danilopolani:
> Howdy! Would you feel comfortable creating a PR to allow Pest to read the dir from the `phpunit.xml` file? We should also ensure that it always (even in case of exceptions) falls back to `tests`.

I think I can work on this in the next days 🤞

lukeraymonddowning:
Awesome! Thank you. And yes, please target v2 (main) branch.

danilopolani:
@lukeraymonddowning @nunomaduro should this be considered closed since it's going to be addressed in v2?

jsonUK:
Thank you for adding this as a feature. Looking forward to using it in v2 when you release it 🙂

PR #506, opened 2022-04-01
[2.x] Read tests folder from PHPUnit.xml file

danilopolani:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #494 

With this PR the tests folder will be read by default from the `phpunit.xml` file and, of course, can still be overwritten by the `--test-directory` arg. 
In the beginning I wanted to use the [TextUI Loader](https://github.com/sebastianbergmann/phpunit/blob/master/src/TextUI/Configuration/Xml/Loader.php) by PHPUnit but it's an internal class and could break at any time and does too much stuff for what was actually needed and, at the end, it was simpler to just use the native PHP XML library.

The code is mainly ispired to [PHPUnit XML Loader](https://github.com/sebastianbergmann/phpunit/blob/b1422ac347707e4039048bdc477ef64f12a05199/src/Util/Xml/Loader.php).

### DIFF
--- a/src/ConfigLoader.php
+++ b/src/ConfigLoader.php
@@ -13,14 +13,17 @@ use Throwable;
  */
 final class ConfigLoader
 {
-    private ?SimpleXMLElement $config = null;
-
     /**
      * Default path if config loading went wrong.
      *
      * @var string
      */
-    private const DEFAULT_TESTS_PATH = 'tests';
+    public const DEFAULT_TESTS_PATH = 'tests';
+
+    /**
+     * XML tree of the PHPUnit configuration file.
+     */
+    private ?SimpleXMLElement $config = null;
 
     /**
      * Creates a new instance of the config loader.
@@ -63,6 +66,26 @@ final class ConfigLoader
     }
 
     /**
+     * Get the configuration file path.
+     */
+    public function getConfigurationFilePath(): string|false
+    {
+        $candidates = [
+            $this->rootPath . '/phpunit.xml',
+            $this->rootPath . '/phpunit.dist.xml',
+            $this->rootPath . '/phpunit.xml.dist',
+        ];
+
+        foreach ($candidates as $candidate) {
+            if (is_file($candidate)) {
+                return realpath($candidate);
+            }
+        }
+
+        return false;
+    }
+
+    /**
      * Load the configuration file.
      */
     private function loadConfiguration(): void
@@ -86,25 +109,5 @@ final class ConfigLoader
 
         // Restore the correct error reporting
         error_reporting($oldReportingLevel);
-    }
-
-    /**
-     * Get the configuration file path.
-     */
-    private function getConfigurationFilePath(): string|false
-    {
-        $candidates = [
-            $this->rootPath . '/phpunit.xml',
-            $this->rootPath . '/phpunit.dist.xml',
-            $this->rootPath . '/phpunit.xml.dist',
-        ];
-
-        foreach ($candidates as $candidate) {
-            if (is_file($candidate)) {
-                return realpath($candidate);
-            }
-        }
-
-        return false;
     }
 }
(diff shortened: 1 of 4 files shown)


src/ConfigLoader.php
```diff
@@ -0,0 +1,113 @@
...
+     */
+    private function loadConfiguration(): void
+    {
+        $configPath = $this->getConfigurationFilePath();
+
+        if ($configPath === false) {
+            return;
+        }
+
+        $oldReportingLevel = error_reporting(0);
```
danilopolani, line 99: Like [PHPUnit does](https://github.com/sebastianbergmann/phpunit/blob/b1422ac347707e4039048bdc477ef64f12a05199/src/Util/Xml/Loader.php#L30), this is needed because both `file_get_contents` and `SimpleXMLElement` throw some **PHP Warning** when an error occurs, instead of just an Exception and this was the only way to suppress those errors.

In any case the original reporting level is restored at the end of the method.

tests/Unit/ConfigLoader.php
```diff
@@ -0,0 +1,19 @@
...
+    expect($instance->getConfigurationFilePath())->toBeFalse();
+    expect($instance->getTestsDirectory())->toBe(ConfigLoader::DEFAULT_TESTS_PATH);
+});
+
+it('fallbacks to default path if phpunit is not a valid XML')->skip();
+it('fallbacks to default path if failing to read phpunit content')->skip();
+it('fallbacks to default path if there is no test suites directory')->skip();
+it('fallbacks to default path if test suite directory has no value')->skip();
+it('fallbacks to default path if test suite directory does not exist')->skip();
+it('returns the parent folder of first test suite directory')->skip();
```
danilopolani, line 19: I don't really know how to implement these: usually I would just move functions like `is_dir`, `file_get_contents` etc. in different methods, mock those methods and that's cool. Here we have `final` classes and therefore I can't mock nor extend it for testing purposes 😕

nunomaduro:
Can you show me a real phpunit.xml example that contains that property?

danilopolani:
> Can you show me a real phpunit.xml example that contains that property?

Hey Nuno, are you talking about the `directory` property? I'm using the one contained in the `<testsuite>` element, picking the first testsuite and its directory, if any.

### Common Laravel example

In this case it will return `./tests` since it's the "base folder" of the first testsuite. I don't think anyone uses two different tests roots (e.g. `/tests` and `/tests-whatever`), and, it they would, it won't work anyway with the current Pest configuration.

```xml
<testsuites>
    <testsuite name="Feature">
      <directory suffix="Test.php">./tests/Feature</directory>
    </testsuite>
   <testsuite name="Unit">
      <directory suffix="Test.php">./tests/Unit</directory>
    </testsuite>
</testsuites>
```

### Current Pest phpunit.xml

Here it returns `./tests` as well, because, if there's only 1 directory separator (as in this scenario), we return the whole value if the directory exists.

```
<testsuites>
  <testsuite name="default">
    <directory suffix=".php">./tests</directory>
  </testsuite>
</testsuites>
```

nunomaduro:
Cool. Can you fix the test suite? Once tests are passing we are good to merge.

danilopolani:
@nunomaduro I don't think that's on me, apparently PHPUnit 10 has dropped the support of PHP 8.0: https://github.com/sebastianbergmann/phpunit/commit/95b61586c5d9b90bcaa516ddca0af00efc408940

I guess Pest too should drop it in v2?

nunomaduro:
Yes. Can you make a separate pull request for that @danilopolani ?

nunomaduro:
Had to revert this pull request. https://github.com/pestphp/pest/commit/b887116e5ce9a69403ad620cad20f0a029474eb5

PR #515, opened 2022-04-17
[1.x] remove constraining typehint

abenerd:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #... 

The `it` helper method has a typehint of `Testcall` which is incorrect as it can also have other return types as specified in the docblock, this PR removes this typehint to fix this and satisfy static analysis tools when using higher order tests with `it`.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -181,11 +181,6 @@ parameters:
 			path: src/Factories/TestCaseFactory.php
 
 		-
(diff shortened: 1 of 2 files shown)


lukeraymonddowning:
Seen as 2.x will be 8.0+, could we use typehints here with multiple return types and target master instead?

nunomaduro:
It's a TestCall returned, yet it magically proxies to other types.

PR #517, opened 2022-04-29
Make 'toHaveKeys' accept multi-dimensional associative arrays

fgaroby:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

This PR allows to use multi-dimensional arrays in the 'toHaveKeys()' method.

For example, from now we can do this:
```php
expect($response)
        ->getStatusCode()->toEqual(Response::HTTP_OK)
        ->getContent() ->json()
        ->toHaveKeys([
           'foo' => [
                'bar',
                'baz' => [
                    'biz',
                ],
            ]
        ]);
```
Of course, the dot notation still works !

### DIFF
--- a/src/Expectations/OppositeExpectation.php
+++ b/src/Expectations/OppositeExpectation.php
@@ -30,7 +30,7 @@ final class OppositeExpectation
     /**
      * Asserts that the value array not has the provided $keys.
      *
-     * @param array<int, int|string|array<int[]|string[]>> $keys
+     * @param array<int, int|string|array<int-string, mixed>> $keys
      *
      * @return Expectation<TValue>
      */
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -595,7 +595,7 @@ final class Expectation
     /**
      * Asserts that the value array has the provided $keys.
      *
-     * @param array<int, int|string|array<int[]|string[]>> $keys
+     * @param array<int, int|string|array<int-string, mixed>> $keys
      *
      * @return Expectation<TValue>
      */


fabio-ivona:
looks good to me

@fgaroby would you mind to fix static analysis failures?

fgaroby:
> @fgaroby would you mind to fix static analysis failures?

Done

src/Expectations/OppositeExpectation.php
```diff
@@ -29,15 +30,19 @@ public function __construct(private Expectation $original)
     /**
      * Asserts that the value array not has the provided $keys.
      *
-     * @param array<int, int|string> $keys
+     * @param array<int, int|string|array<int[]|string[]>> $keys
```
fabio-ivona, line 33: For a recursive array, this could be more readable by phpstan:


```suggestion
     * @param array<int, int|string|array<int-string, mixed>> $keys
```

Otherwise you could have a phpstan error on this:

`expect($foo)->toHaveKeys(['bar' => ['baz' => ['quuz', 'quz']]])`

As it will complain that the third array level should be made of strings or ints

fabio-ivona (changes requested):
Just needs a minor phpstan fix and it's fine for me

src/Expectations/OppositeExpectation.php
fgaroby: OK, I corrected the both files.

fabio-ivona (approved):
Lgtm

fabio-ivona:
@dansysanalyst should this be added to the docs?

PR #554, opened 2022-08-09
[2.x] Skipping tests based on group parameter

emtudo:
This PR brings the possibility to skip or run tests based on stipulated groups via the command line using using the `--groups=` parameter.

### Motivation

I would like my tests to reach the API and endpoints during ONLY deployment.  This could be achieved with phpunit.xml or with aliases or  makefile, but this would need external configuration by my other team members.
Also, I would like to avoid passing commands every time I run tests.

### Usage

The test below will be **skipped** unless the group is explicitly called using the `--groups=` parameter.

Runs the test: `./vendor/bin/pest --group=group1`
Skips the test: `./vendor/bin/pest`

```php
it('runs this test when group is called', function () {
    expect(true)->toBeTrue();
})
    ->group('group1') // it is not necessary
    ->skipUnlessGroup('group1');
```

### DIFF
--- a/src/PendingCalls/TestCall.php
+++ b/src/PendingCalls/TestCall.php
@@ -178,7 +178,7 @@ final class TestCall
      */
     public function skipUnlessGroup(string $group, string $message = ''): TestCall
     {
-        $has = (new ArgvInput())->hasParameterOption("--group={$group}");
+        $has = ((new ArgvInput())->getParameterOption('--group') === $group);
 
         return $this->skip(!$has, $message)
             ->group($group);
--- a/src/PendingCalls/TestCall.php
+++ b/src/PendingCalls/TestCall.php
@@ -15,6 +15,7 @@ use Pest\Support\HigherOrderCallables;
 use Pest\Support\NullClosure;
 use Pest\TestSuite;
 use SebastianBergmann\Exporter\Exporter;
+use Symfony\Component\Console\Input\ArgvInput;
 
 /**
  * @internal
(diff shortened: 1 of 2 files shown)


fabio-ivona (approved):

fabio-ivona:
Lgtm

src/PendingCalls/TestCall.php
```diff
@@ -172,6 +173,17 @@ public function skip(Closure|bool|string $conditionOrMessage = true, string $mes
         return $this;
     }
 
+    /**
+     * Skips the current test when no set group.
```
andersonef, line 177:
```suggestion
     * Skips the current test when no set group.
     @param $group string
     @param $message string
     @return TestCall
```

src/PendingCalls/TestCall.php
```diff
@@ -172,6 +173,17 @@ public function skip(Closure|bool|string $conditionOrMessage = true, string $mes
...
+     */
+    public function skipUnlessGroup(string $group, string $message = ''): TestCall
```
andersonef, line 179: I suggest the name of this method could be **skipUnlessGroupIs**

andersonef (changes requested):

src/PendingCalls/TestCall.php
emtudo: This is not the design default.

src/PendingCalls/TestCall.php
emtudo: ls at the end of the name can confuse the person and is "I" or "L"

src/PendingCalls/TestCall.php
fabio-ivona: I agree, no unnecessary docs

andersonef (approved):

emtudo:
I had to make an adjustment because it didn't work in parallel, since the behavior of ArgvInput in parallel is different.

nunomaduro:
No plans for this for now.

PR #560, opened 2022-08-24
[1.x] Add new command make:pest as an alias for pest:test

tobya:
| Q             | A
| ------------- | ---
| Bug fix?      |  no
| New feature?  | yes
| Fixed tickets | none

I would really like to be able to do

> php artisan make:pest

rather than

> php artisan pest:test

This PR adds a new command `make:pest`  that extends PestTestCommand but with simply a new command name.  This is inspired by @calebporzio who has lots of aliases for many of his livewire commands.

I could hide the command but it seems that it would be nice for it not to be hidden.

I didnt add tests as I couldnt find tests for pest:test, but it is possible I am being blind.

Let me know if you need me to make changes.

ps. thanks for a great tool, first time in 20 years I've started to really use tests in php.

### DIFF
--- a/src/Laravel/Commands/MakePestCommand.php
+++ b/src/Laravel/Commands/MakePestCommand.php
@@ -6,12 +6,7 @@ namespace Pest\Laravel\Commands;
 
 use Illuminate\Console\Command;
 use Illuminate\Support\Facades\File;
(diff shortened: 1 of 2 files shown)


nunomaduro:
No plans for this.

PR #561, opened 2022-08-25
[1.x] dynamic properties handling

### LINKED ISSUE #555: [PHP 8.2] Handling dynamic properties?
I'm wondering what is the correct way of handling dynamic properties defined in the `beforeEach` method?

In the documentation there is the following example:

```php
beforeEach(function () {
    $this->hey = 'artisan';
});
 
it('has artisan', function () {
    echo $this->hey;
});
```

But in PHP 8.2, dynamic properties are deprecated, and this will throw a notice that will fail the tests:

> Whoops\Exception\ErrorException: Creation of dynamic property P\Tests\Unit\Example\ExampleTest::$hey is deprecated

fabio-ivona:
Implements dynamic properties handling in order to avoid deprecation in PHP 8.2

fix #555

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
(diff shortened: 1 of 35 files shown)


fabio-ivona:
Paused, waiting for tests to be fixed #562

PR #563, opened 2022-08-25
[1.x] Dynamic properties handling

fabio-ivona:
Implements dynamic properties handling in order to avoid deprecation in PHP 8.2

fix https://github.com/pestphp/pest/issues/555

a PR for v2 will follow this one

### DIFF
--- a/src/Factories/TestCaseFactory.php
+++ b/src/Factories/TestCaseFactory.php
@@ -216,6 +216,7 @@ final class TestCaseFactory
             eval("
                 namespace $namespace;
 
+                #[AllowDynamicProperties]
                 final class $className extends $baseClass implements $hasPrintableTestCaseClassFQN {
                     $traitsCode
 


tests/Hooks/AfterEachTest.php
```diff
@@ -1,18 +1,14 @@
 <?php
 
 uses()->afterEach(function () {
-    expect($this)
-        ->toHaveProperty('ith')
```
fabio-ivona, line 5: `toHaveProperty` expectation doesn't work for properties handled with __get method

nunomaduro:
Can't you solve this issue using the `#[AllowDynamicProperties]` thing?

fabio-ivona:
> Can't you solve this issue using the #[AllowDynamicProperties] thing?

just wanted to fix the issue in a definitive way, but I can do that, np

fabio-ivona:
@nunomaduro done

nunomaduro:
Can you port that to pest v2?

dingo-d:
@fabio-ivona @nunomaduro this is not really a fix 😬 

You are just silencing the deprecation until PHP 9 🙈 

The underlying test case should probably have a `__set($property, $propertyValue)` method or something similar in order for the properties to be creatable on the fly. 

That way, any reference to the property inside the test case that doesn't exist on the `TestCase` class would be created on the fly.

fabio-ivona:
As it is a deprecation, it is fine to silence it, we'll plan a clean fix for php9

PR #565, opened 2022-08-26
[1.x] PHP 8.1 enums display value in console

fabio-ivona:
This PR will fix how PHP 8.1 Enums are displayed in dataset names and failure messages

they are currently exported as:

image

with the new code, the enum names/values will be exported as well

image

**Note**

I guess the issue will be gone in v2, as it has been fixed in sebastian/exporter master [branch](https://github.com/sebastianbergmann/exporter/blob/5b97ae9ab116bfa2b30559840005cb907bf7eeae/src/Exporter.php#L122)


fabio-ivona:
@nunomaduro hard to write tests for this feature, as every reference to `enum` breaks code in previous php versions

cannot find a solution, any ideas?

fabio-ivona:
@lukeraymonddowning and @nunomaduro 

I implemented a workaround to let us run Enums tests: 

I added a step in tests.yml workflow that parses all .php tests files for a comment like `/* @min-php-version 8.1 */` and append `.disabled` to the filename if the current php version doesn't match the requirement


**also** as we disable some test depending on php version, the visual snapshot of integration tests differs from one php version to another. 

to solve this I created an _integration tests_ workflow that will be run against the latest php versions (currently php 8.1 and 8.2)

fabio-ivona:
@owenvoke I added you to the reviewers as well as I changed the workflow setup

nunomaduro:
Can you point this to `2.x` instead?

fabio-ivona:
@nunomaduro V2 doesn't need this, it has already been fixed in PhpUnit V10

nunomaduro:
In that case, lets make this change Pest 2.x only.

PR #566, opened 2022-08-29
2.x dynamic properties fix

fabio-ivona:
- Fix missing `\` in #[\AllowDynamicProperties]  attribute
- Add PHP8.2 test workflow
- Fix tests failing with PHP8.2

### DIFF
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -710,16 +710,20 @@ final class Expectation
     /**
      * Asserts that the value array matches the given array subset.
      *
-     * @param iterable<int|string, mixed> $array
+     * @param iterable<int|string, mixed>|object $array
      *
      * @return Expectation<TValue>
      */
-    public function toMatchArray(iterable|object $array): Expectation
+    public function toMatchArray(array|object $array): Expectation
     {
         if (is_object($this->value) && method_exists($this->value, 'toArray')) {
             $valueAsArray = $this->value->toArray();
         } else {
             $valueAsArray = (array) $this->value;
+        }
+
+        if (is_object($array) && method_exists($array, 'toArray')) {
+            $array = $array->toArray();
         }
 
         foreach ($array as $key => $value) {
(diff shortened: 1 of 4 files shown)


src/Mixins/Expectation.php
```diff
@@ -710,18 +710,22 @@ public function toBeWritableFile(): Expectation
     /**
      * Asserts that the value array matches the given array subset.
      *
-     * @param iterable<int|string, mixed> $array
+     * @param iterable<int|string, mixed>|object $array
      *
      * @return Expectation<TValue>
      */
-    public function toMatchArray(iterable|object $array): Expectation
+    public function toMatchArray(array|object $array): Expectation
```
nunomaduro, line 717: This should be `iterable` only.

src/Mixins/Expectation.php
```diff
@@ -743,11 +747,11 @@ public function toMatchArray(iterable|object $array): Expectation
      * Asserts that the value object matches a subset
      * of the properties of an given object.
      *
-     * @param iterable<string, mixed>|object $object
+     * @param object|array<string, mixed> $object
      *
      * @return Expectation<TValue>
      */
-    public function toMatchObject(iterable|object $object): Expectation
+    public function toMatchObject(object|array $object): Expectation
```
nunomaduro, line 754: This should be `iterable` only.

nunomaduro (changes requested):

src/Mixins/Expectation.php
```diff
@@ -714,14 +714,18 @@ public function toBeWritableFile(): Expectation
...
-    public function toMatchArray(iterable|object $array): Expectation
+    public function toMatchArray(iterable $array): Expectation
     {
         if (is_object($this->value) && method_exists($this->value, 'toArray')) {
             $valueAsArray = $this->value->toArray();
         } else {
             $valueAsArray = (array) $this->value;
         }
 
+        if (is_object($array) && method_exists($array, 'toArray')) {
```
nunomaduro, line 725: This need code needs a test.

nunomaduro (changes requested):

src/Mixins/Expectation.php
fabio-ivona: I removed that code, added it under the wrong assumption that the expectation should accept an object

PR #583, opened 2022-09-19
[2.x] Add custom message to failed expectations

fabio-ivona:
This PR will implement custom messages for expectation failures

a trailing optional `message` parameter has been added to all expectations, in order to let the user to set a custom error message like this one:

```php
 expect($obj)->toHaveProperty(name: 'bar', message: 'oh no!');
```

**note**

1. couldn't add the custom message to the `toContain()` expectation, as it is the only expectation to have variadic parameters the trailing `$message`  is not allowed.

2. fixed a bug in expectation pipelines, they didn't handle named parameters in expectations


src/Mixins/Expectation.php
```diff
@@ -271,16 +272,16 @@ public function toHaveCount(int $count): self
      *
      * @return self<TValue>
      */
-    public function toHaveProperty(string $name, mixed $value = null): self
+    public function toHaveProperty(string $name, mixed $value = new NullValue(), string $failureMessage = ''): self
```
nunomaduro: @fabio-ivona Can you move this `NullValue` to an folder called `Matchers`, and call it `Any`?

That `Any` type should be created as the following:

`expect($obj)-> toHaveProperty('name', expect()->any(), 'my failure');`

nunomaduro (changes requested):

src/Mixins/Expectation.php
fabio-ivona: done

src/Expectation.php
```diff
@@ -339,4 +340,9 @@ public static function hasMethod(string $name): bool
             || method_exists(Mixins\Expectation::class, $name)
             || self::hasExtend($name);
     }
+
+    public function any(): Any
```
nunomaduro, line 347: Added a description, and mark this method as internal for now.

src/Matchers/Any.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pest\Matchers;
+
+final class Any
```
nunomaduro, line 10: Can you mark this call as internal for now?

src/Expectation.php
nunomaduro, line 347: Can you add a comment explain what this does? You may want to check Jest here.

fabio-ivona: Do you think that "Matches any argument" would be enough as explanation?

src/Matchers/Any.php
fabio-ivona: done

src/Expectation.php
fabio-ivona: done

nunomaduro: `Matches any value`

fabio-ivona: done

PR #584, opened 2022-09-19
[2.x] Dataset scopes

fabio-ivona:
This PR will implement **scoped datasets**

in v1 datasets can be defined in 3 ways:

1. inside a test file 
2. inside .php files in a `Datasets` directory placed it the test suite root
3. inside a `Datasets.php` file placed in the test suite root

with this PR datasets can be scoped with the following rules:

1. _a dataset defined inside a test file_ (using the `dataset()` function) **is scoped to the file itself** and will not be accessible from other test files 
2. _a dataset defined inside a test file_ **will have precedence** with datasets with the same name defined outside of the test file
3. _a dataset defined in a Datasets.php file_ will be scoped to all test defined in the same directory and/or child directories
4. _a dataset defined in a .php file in a Datasets directory_ will be scoped to all tests defined in directory where the Datasets directory belongs and all tests in any child directory
5. two datasets with the same name can't share the same scope

given these rules, each test file will see the datasets in its scope, with the nearest taking precedence to the farthest

**Overlapping datasets**

with this directory tree:

```
/tests
   |--Datasets
   |      |--Numbers.php (with a 'numbers' dataset inside)   [1]
   |
   |--Datasets.php (with a 'numbers' dataset inside)         [2]
   |--Pest.php

```
  
-  [1] and [2] will throw an exception as the two `numbers` datasets will have the same scope (/tests)

**Scope precedence**

```
/tests
   |--Datasets
   |      |--Numbers.php (with a 'numbers' dataset inside)   [1]
   |
   |--Feature
   |      |--Controllers
   |      |      |--ApiControllers
   |      |      |     |--PostApiController  (will use `numbers` dataset [2])
   |      |      |     
   |      |      |--Datasets.php  (with a 'numbers' dataset inside) [2]
   |      |      |
   |      |      |--PostController.php (will use `numbers` dataset [2])
   |      |
   |      |--Models
   |      |      |--Datasets
   |      |      |      |--Numbers.php  with a 'numbers' dataset inside) [3]
   |      |      |
   |      |      |--PostController.php (will use `numbers` dataset [3])
   |      |
   |      |--Actions
   |             |--SavePostTest.php   (will use `numbers` dataset [1])
   |--Pest.php

```

**TODO**

- [x] It should be avoided to define tests inside Datasets files, I have to find a way to prevent that (maybe disabling `test` and `it` functions during the `BootFiles` step

### DIFF
--- a/src/Exceptions/DatasetAlreadyExist.php
+++ b/src/Exceptions/DatasetAlreadyExist.php
@@ -17,8 +17,8 @@ final class DatasetAlreadyExist extends InvalidArgumentException implements Exce
     /**
      * Creates a new Exception instance.
      */
-    public function __construct(string $name)
+    public function __construct(string $name, string $scope)
     {
-        parent::__construct(sprintf('A dataset with the name `%s` already exist.', $name));
+        parent::__construct(sprintf('A dataset with the name `%s` already exist in scope [%s].', $name, $scope));
     }
 }
--- a/src/Functions.php
+++ b/src/Functions.php
@@ -9,8 +9,8 @@ use Pest\PendingCalls\TestCall;
 use Pest\PendingCalls\UsesCall;
 use Pest\Repositories\DatasetsRepository;
 use Pest\Support\Backtrace;
+use Pest\Support\DatasetInfo;
 use Pest\Support\HigherOrderTapProxy;
-use Pest\Support\Str;
 use Pest\TestSuite;
 use PHPUnit\Framework\TestCase;
 
@@ -61,16 +61,7 @@ if (! function_exists('dataset')) {
      */
     function dataset(string $name, Closure|iterable $dataset): void
     {
-        $file = Backtrace::datasetsFile();
-        $filename = Str::afterLast($file, DIRECTORY_SEPARATOR);
-        $scope = Str::beforeLast($file, DIRECTORY_SEPARATOR);
-
-        if (Str::afterLast($scope, DIRECTORY_SEPARATOR) === 'Datasets') {
-            $scope = Str::beforeLast($scope, DIRECTORY_SEPARATOR);
-        } elseif ($filename !== 'Datasets.php') {
-            $scope = $file;
-        }
-
+        $scope = DatasetInfo::scope(Backtrace::datasetsFile());
         DatasetsRepository::set($name, $dataset, $scope);
     }
 }
--- a/src/Functions.php
+++ b/src/Functions.php
@@ -10,6 +10,7 @@ use Pest\PendingCalls\UsesCall;
 use Pest\Repositories\DatasetsRepository;
 use Pest\Support\Backtrace;
 use Pest\Support\HigherOrderTapProxy;
+use Pest\Support\Str;
 use Pest\TestSuite;
 use PHPUnit\Framework\TestCase;
 
@@ -60,7 +61,17 @@ if (! function_exists('dataset')) {
      */
     function dataset(string $name, Closure|iterable $dataset): void
     {
-        DatasetsRepository::set($name, $dataset);
+        $file = Backtrace::datasetsFile();
+        $filename = Str::afterLast($file, DIRECTORY_SEPARATOR);
+        $scope = Str::beforeLast($file, DIRECTORY_SEPARATOR);
+
+        if (Str::afterLast($scope, DIRECTORY_SEPARATOR) === 'Datasets') {
+            $scope = Str::beforeLast($scope, DIRECTORY_SEPARATOR);
+        } elseif ($filename !== 'Datasets.php') {
+            $scope = $file;
+        }
+
+        DatasetsRepository::set($name, $dataset, $scope);
     }
 }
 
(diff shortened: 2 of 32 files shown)


src/Functions.php
```diff
@@ -60,7 +61,17 @@ function beforeEach(Closure $closure = null): BeforeEachCall
      */
     function dataset(string $name, Closure|iterable $dataset): void
     {
-        DatasetsRepository::set($name, $dataset);
+        $file = Backtrace::datasetsFile();
+        $filename = Str::afterLast($file, DIRECTORY_SEPARATOR);
+        $scope = Str::beforeLast($file, DIRECTORY_SEPARATOR);
+
+        if (Str::afterLast($scope, DIRECTORY_SEPARATOR) === 'Datasets') {
```
lukeraymonddowning: We do these same `Str` checks in multiple files. I wonder if we should introduce a `Support\Dataset` class?

src/Repositories/DatasetsRepository.php
```diff
@@ -67,9 +69,10 @@ public static function has(string $filename, string $description): bool
      */
     public static function get(string $filename, string $description)
     {
+        // dump("requesting file: " . $filename);
```
lukeraymonddowning: Left this in by mistake?

src/Repositories/DatasetsRepository.php
```diff
@@ -132,23 +135,53 @@ public static function resolve(string $description, array $dataset): array|null
         return $namedData;
     }
 
+    /**
+     * @return Closure|iterable<int|string, mixed>
+     */
+    private static function getScopedDataset(string $name, string $currentTestFile)
+    {
+        $matchingDatasets = array_filter(self::$datasets, function (string $key) use ($name, $currentTestFile) {
+            [$datasetScope, $datasetName] = explode('>>>', $key);
```
lukeraymonddowning: I don't really like that we carry this '>>>' string around everywhere. Should we extract it to a constant?

src/Repositories/DatasetsRepository.php
```diff
@@ -132,23 +135,53 @@ public static function resolve(string $description, array $dataset): array|null
...
+            }
+
+            if (! str_starts_with($currentTestFile, $datasetScope)) {
+                return false;
+            }
+
+            return true;
+        }, ARRAY_FILTER_USE_KEY);
+
+        $closestScopeDatasetKey = array_reduce(array_keys($matchingDatasets), function ($keyA, $keyB) {
```
lukeraymonddowning:
```suggestion
        $closestScopeDatasetKey = array_reduce(
            array_keys($matchingDatasets), 
            fn ($keyA, $keyB) => $keyA !== null && strlen($keyA) > strlen($keyB) ? $keyA : $keyB
        );
```

src/Exceptions/DatasetAlreadyExist.php
```diff
@@ -17,8 +17,8 @@ final class DatasetAlreadyExist extends InvalidArgumentException implements Exce
     /**
      * Creates a new Exception instance.
      */
-    public function __construct(string $name)
+    public function __construct(string $name, string $scope)
```
lukeraymonddowning, line 20: @nunomaduro shall we use the v2 opportunity to rename this to `DatasetAlreadyExists`? Grammatically correct :-)

src/Repositories/DatasetsRepository.php
```diff
@@ -132,23 +135,53 @@ public static function resolve(string $description, array $dataset): array|null
...
         foreach ($datasets as $index => $data) {
             $processedDataset = [];
 
             if (is_string($data)) {
-                if (! array_key_exists($data, self::$datasets)) {
-                    throw new DatasetDoesNotExist($data);
-                }
-
-                $datasets[$index] = self::$datasets[$data];
+                $datasets[$index] = self::getScopedDataset($data, $currentTestFile);
```
lukeraymonddowning, line 151: Small thing, but can the `getScopedDataset` method go below this method?

lukeraymonddowning (changes requested):
Looking pretty sweet so far!

src/Functions.php
fabio-ivona: @lukeraymonddowning I refactored this a bit, extracted to a support class and changed Str:: support functions with dirname and basename, so there is less code to maintain

you can find the changes here: 

https://github.com/pestphp/pest/pull/584/commits/80854d5f8700a958be28433cd60c2de74404c4a3

src/Repositories/DatasetsRepository.php
fabio-ivona: lol, thanks :rofl:

src/Repositories/DatasetsRepository.php
fabio-ivona: done

src/Repositories/DatasetsRepository.php
fabio-ivona: done

src/Exceptions/DatasetAlreadyExist.php
fabio-ivona: It's an internal class, so don't think there are any issues in doing it

src/Functions.php
lukeraymonddowning: So much nicer ❤️

nunomaduro (approved):
Good job!

PR #599, opened 2022-10-27
chore: narrow types for `it()` and `test()`

owenvoke:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Fixed tickets | #... 

Because `it()` sets a description, it will always return `TestCall`. This also adds a conditional PHPStan return type for `test()` as we can work out what it returns based on this. 👍🏻

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -181,11 +181,6 @@ parameters:
 			path: src/Factories/TestCaseFactory.php
 
 		-
-			message: "#^Function it\\(\\) should return Pest\\\\PendingObjects\\\\TestCall but returns mixed\\.$#"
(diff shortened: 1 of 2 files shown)


owenvoke:
> The it() function was added to the codebase and also returns a TestCall object.

Bad bot. 🤷🏻 🙂

mpociot:
It's still learning :)

nunomaduro:
@owenvoke Can you make this pull request against `2.x` instead? Not really in my plans change stuff on `1.x`.

owenvoke:
No problem, we'll stub it for our use needs, and I'll PR this to 2.x

PR #600, opened 2022-10-27
[2.x] chore: narrow types for `it()` and `test()`

owenvoke:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Fixed tickets | #... 

Because `it()` sets a description, it will always return `TestCall`. This also adds a conditional PHPStan return type for `test()` as we can work out what it returns based on this. 👍🏻

Original PR for 1.x: https://github.com/pestphp/pest/pull/599

### DIFF
--- a/src/Functions.php
+++ b/src/Functions.php
@@ -86,6 +86,8 @@ if (! function_exists('test')) {
      * a closure that contains the test expectations.
      *
      * @return TestCall|TestCase|mixed
+     *
+     * @phpstan-return ($description is null ? TestCase|mixed : TestCall)
      */
     function test(string $description = null, Closure $closure = null)
     {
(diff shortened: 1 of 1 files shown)


src/Functions.php
```diff
@@ -105,16 +107,13 @@ function test(string $description = null, Closure $closure = null)
      * is the test description; the second argument is
      * a closure that contains the test expectations.
      *
-     * @return TestCall|TestCase|mixed
+     * @return TestCall
```
nunomaduro, line 110: You sure this change does not cause auto-completion issues?

owenvoke: Just looking into it further. 👍🏻

nunomaduro:
Just re-open this pull request if necessary.

PR #613, opened 2022-11-23
Exit only if $exit is true

soilSpoon:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #... 

The `run` method of the `Pest\Console\Command` class executes `exit()` regardless of the `$exit` parameter.

Fix the above phenomenon.

### DIFF
--- a/src/Console/Command.php
+++ b/src/Console/Command.php
@@ -119,7 +119,11 @@ final class Command extends BaseCommand
         $result = parent::run($argv, false);
         $result = InteractsWithPlugins::addOutput($result);
 
-        exit($result);
+        if ($exit) {
+            exit($result);
+        }
+
+        return $result;
     }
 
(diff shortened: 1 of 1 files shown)


owenvoke:
I'd seen this before, but always assumed it was on purpose. Ping @nunomaduro to check, but I don't think this will probably be changed.

soilSpoon:
> I'd seen this before, but always assumed it was on purpose. Ping @nunomaduro to check, but I don't think this will probably be changed.

I'm experimenting with changing pest-parallel-plugin to WrapperRunner to speed it up. To do this, you need the non-exit option.

nunomaduro:
We don't to changes on Pest `v1.x` branch anymore. Can you pull request those changes on the `2.x`?

PR #644, opened 2023-01-15
Ensures toHaveKey() has a iterable value

dansysanalyst:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | -

This PR ensures the `$value` is of type `iterable` for the `toHaveKey()` and  `toHaveKeys()` Expectations. 

Both will throw an `InvalidExpectationValue` exception, like some other Expectations do.

This proposed change also improves consistency with PHPUnit 9.5, as the  [ArrayHasKey](https://phpunit.readthedocs.io/en/9.5/assertions.html?highlight=arrayHasKey#assertarrayhaskey) assertion throws an exception when a string is passed:

```php
$this->assertArrayHasKey('foo', 'hello world') 
//PHPUnit\Framework\Assert::assertArrayHasKey(): Argument #2 ($array) must be of type ArrayAccess|array, string given
```

### Motivation

As questioned by the user SomaiyaUtsav on [Twitter](https://twitter.com/SomaiyaUtsav/status/1612068864229257218), there is a strange/misleading behavior resulting in a false positive when checking if a `string` has keys:

Under the hood, the string `'for the best Pest content follow'` is converted to an array `[0 => 'for the best Pest content follow']` and the test passes.

### Open points

` 📚 ` The Pest Doc accordingly will be updated should this PR be accepted.
` ❓ ` Is this 2.x change or a bug fix to be included in 1.x?

### DIFF
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -560,14 +560,18 @@ final class Expectation
      */
     public function toHaveKey(string|int $key, mixed $value = new Any(), string $message = ''): self
     {
-        if (! is_iterable($this->value)) {
-            InvalidExpectationValue::expected('iterable');
-        }
+        $array = $this->value;
 
         if (is_object($this->value) && method_exists($this->value, 'toArray')) {
             $array = $this->value->toArray();
-        } else {
+        }
+
+        if (is_object($array)) {
             $array = (array) $this->value;
+        }
+
+        if (! is_iterable($array)) {
+            InvalidExpectationValue::expected('iterable');
         }
 
         try {
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -560,6 +560,10 @@ final class Expectation
      */
     public function toHaveKey(string|int $key, mixed $value = new Any(), string $message = ''): self
     {
+        if (! is_iterable($this->value)) {
+            InvalidExpectationValue::expected('iterable');
+        }
+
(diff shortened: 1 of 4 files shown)


src/Mixins/Expectation.php
```diff
@@ -560,6 +560,10 @@ public function toBeNull(string $message = ''): self
      */
     public function toHaveKey(string|int $key, mixed $value = new Any(), string $message = ''): self
     {
+        if (! is_iterable($this->value)) {
```
nunomaduro: This needs to be done within the `else` statement just bellow.

nunomaduro (changes requested):

src/Mixins/Expectation.php
dansysanalyst: I was guided by the idea to ensure the Expectation value is an `iterable` to avoid this behavior present in v1:

```php 
test('using string')->expect('long string to have key')->toHaveKey(0); //Passes
``` 

However, this also passes on v1 and checking for iterable would break it:

```php
class ExampleClass
{
    public $foo = 'foo';
}

test('pass with a class')->expect((new ExampleClass()))->toHaveKey('foo');

//is_iterable(new stdClass()) === false
```

I decided to avoid a breaking change in favor of user experience in v2.

So, I came up with some changes but I'm feeling a bit conflicted and not quite happy with the code.

nunomaduro:
@dansysanalyst I will visit this pull request in the future.

PR #646, opened 2023-01-17
fixes duplicate --coverage-php option

nicko170:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| Fixed tickets | #... 

This is a multi-tiered issue, running pest in my CI environment, all tests are failing as it's trying to write to the vendor directory - but given that directory is read only (it's mounted as RO in a container) - this fails.

Looking further into the issue - the --coverage-php option sets where coverage.php is actually written - though, when if you set the option in the CLI, it does not respect that and still tries writing to a temp directory within the vendor directory.

This PR just adds a check to only set that flag if it doesn't exist already - so no duplicate, and the override actually works.


nunomaduro:
Can you send this to the `2.x` if you are facing the same issue on Pest v2?

PR #651, opened 2023-01-28
[2.x] Docker support for OSS contributors

lukeraymonddowning:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

This PR adds support for building Docker images that are capable of running all the checks required to submit a PR to the Pest codebase. This will allow developers to very quickly get the project up and running locally so that they are able to contribute.

Here is how simple it is to get started:

1. `docker compose build` to build the Docker image
2. `docker compose run --rm composer install` to install Composer dependencies
3. `docker compose run --rm composer test` to run the project tests and analysis tools

The docker image is built on a minimal Alpine image to keep it as small as possible.

By default, the Docker image uses PHP 8.1, as this is the lowest supported PHP version. One nice thing about adding support for this is how easy it is to try out a PR against other PHP versions. For example, to test against PHP 8.2, a developer would run the following:

1. `docker compose build --build-arg PHP=8.2`
2. `docker compose run --rm composer update`
3. `docker compose run --rm composer test`

#LetsPestThis

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,5 @@
 /art                     export-ignore
+/docker                  export-ignore
 /docs                    export-ignore
 /tests                   export-ignore
 /scripts                 export-ignore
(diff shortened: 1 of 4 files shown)


nunomaduro:
@lukeraymonddowning can you tweet about this?

lukeraymonddowning:
Of course!

PR #659, opened 2023-02-08
Parallel support

lukeraymonddowning:


lukeraymonddowning:
TODO:

- [x] Fix `--dirty` integration
- [x] Fix Laravel integration when using `./vendor/bin/pest -p` instead of `php artisan test -p`
- [x] Test with `--coverage` and `--min`
- [x] Test with `--recreate-databases`
- [x] Test with `--drop-databases`
- [x] Get Pest todos to show as blue 't's in printer output
- [x] Refactor and cleanup
- [x] Fix PHPStan errors
- [x] Document removal or `pest-plugin-parallel` in upgrade guide
- [x] Test with `--only` and `--ci`
- [x] Test with `--retry`
- [x] Test with `--todo`
- [ ] Having a test name with a \' in it causes a segmentation violation (eg: `it("saves the user\'s surname correctly")`
- [ ] I noticed a strange bug where in the error report it would show the wrong test name. The exception it displayed was correct

> Note: `--coverage` works, but I noticed I had to set the `cacheDirectory` attribute on the `coverage` node in phpunit.xml. I don't have to do that outside of parallel. There is a `--cache-directory` you can set on the command alternatively

PR #662, opened 2023-02-13
2.x Parallel todo support

lukeraymonddowning:
This PR cleans up files for Pest Parallel and adds support for todo items in the progress printer.

CleanShot 2023-02-13 at 11 14 51@2x

We do this by overriding PHPUnit's `TestSkippedSubscriber` when in parallel. We do have to use reflection to achieve this, which I'm not best pleased about. There is an alternate solution, which is to create our own `ProgressPrinter`. That class would basically output the Pest compact printer output directly, and Parallel would concatenate the results and stream them directly to the console. The downside of that approach is that we would be entirely in charge of the `ProgressPrinter`.

So, it would be good to hear thoughts on the preferred approach here :-)


src/Plugins/Parallel/Handlers/Laravel.php
```diff
@@ -43,10 +46,8 @@ private static function whenUsingLaravel(array $arguments, Closure $closure): ar
     {
         $isLaravelApplication = InstalledVersions::isInstalled('laravel/framework', false);
         $isLaravelPackage = class_exists(\Orchestra\Testbench\TestCase::class);
-        if ($isLaravelApplication) {
-            return $closure($arguments);
-        }
-        if ($isLaravelPackage) {
+
+        if ($isLaravelApplication && ! $isLaravelPackage) {
```
lukeraymonddowning, line 50: @nunomaduro noticed you changed the logic here, but AFAIK we don't want to run the Laravel test runner in a package context because it won't work.

owenvoke (approved):
Worked on this with you, and I think this is the cleanest method (although it's a shame we had to use reflection). 😞

src/Plugins/Parallel/Handlers/Laravel.php
nunomaduro: Ups! Good point, can you update this?

lukeraymonddowning: Already have :-) It won't apply the Laravel runner anymore if it is a Laravel package.

nunomaduro (approved):

PR #667, opened 2023-02-18
Fix tests and static analysis

fabio-ivona:
Fix failing larastan tests and updates snapshost to match changes in `tests/Unit/Support/Str.php`

image


nunomaduro:
I've already fixed locally. Just missing push.

PR #676, opened 2023-03-03
[2.x] shorter if condition

alirezasalehizadeh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Fixed tickets | #...


alirezasalehizadeh:
@nunomaduro Hi nuno,
How you feel write `load` method like this?
```php 
private function load(string $filename): void
    {
        if (Str::endsWith($filename, '.php') || file_exists($filename)) {
            include_once $filename;
        }
    }
```

PR #689, opened 2023-03-15
See the underwater part of the iceberg

szepeviktor:
⛰️


szepeviktor:
@nunomaduro I'm wondering whether these kind of improvements find their place in Pest.

https://github.com/pestphp/pest/commit/3fc1540f581e0938407100d0fae38e1a5db3988d

`isset` is the tricky way to tell that the previous condition - `$this->options->needsTeamcity` - was true.

What do you think?

szepeviktor:
No comment?

PR #698, opened 2023-03-18
[chore] Better dataset arguments mismatch message

fabio-ivona:
This PR will show a better error message when dataset values dont match test arguments count

with this test:

```php
it("throws an understandable error message if datasets arguments number doesn't match", function (int $first, int $second){
    // ... 
})->with([
    ['first'=> 1]
]);
```

from: 

image

to:

image


fabio-ivona:
@nunomaduro rebased

paulshryock:
I feel the dataset name should be omitted.

What happens if the dataset happens to be very verbose? Or not a string?

```php
test()->with([
  'first' => [['hello', 'world', 'lorem', 'ipsum', 'dolor', 1, 2, 3, 4, 5], 'second parameter', 'third'],
]);
```

Looks like `sprintf` won't handle non-scalar values so well, and throws a warning:

```php
echo sprintf('echo dataset [%s]', ['hello', 'world']);
```

```bash
PHP Warning:  Array to string conversion in php shell code on line 1
PHP Stack trace:
PHP   1. {main}() php shell code:0
PHP   2. sprintf($format = 'echo dataset [%s]', ...$values = variadic()) php shell code:1

Warning: Array to string conversion in php shell code on line 1

Call Stack:
   24.1013     404088   1. {main}() php shell code:0
   24.1013     404088   2. sprintf($format = 'echo dataset [%s]', $values = [0 => 'hello', 1 => 'world']) php shell code:1

echo dataset [Array]
```

The message could just be:

```php
sprintf('Test expects %d arguments but dataset only provides %d', $requiredCount, $suppliedCount)
```

paulshryock:
I also strongly feel the file name and line number should be left in the error message. This change will make it much harder to debug without including `in /home/fabio/projects/pest/src/Factories/TestCaseMethodFactory.php on line 100`.

Honestly I kind of think the old message was a lot clearer in exactly what went wrong and how to fix it.

fabio-ivona:
Thanks @paulshryock I will take a look at it

PR #699, opened 2023-03-19
Refactor Expectation `unless` method

faissaloux:
Hello 👋🏻 

While checking `Expectation` class I have found that this code block is duplicated on `unless` and `when`, so when we call the `unless` method, this block got executed twice.
```php
$condition = is_callable($condition)
            ? $condition
            : static fn (): bool => $condition;
```
So in this PR I have tried to make it better.


nunomaduro:
It's a good idea. However, I still think the existing code is more clear.

PR #700, opened 2023-03-19
Fix test Case Name starts with a P

faissaloux:
| Q             | A
| ------------- | ---
| Bug fix?      | yes

In this Pr I have fixed a bug that removes the first P from test `Case Name`.

### Before
Screenshot 2023-03-19 174241

### After
Screenshot 2023-03-19 174345

### DIFF
--- a/src/Concerns/Testable.php
+++ b/src/Concerns/Testable.php
@@ -277,7 +277,7 @@ trait Testable
      */
     public static function getPrintableTestCaseName(): string
     {
-        return ltrim(self::class, 'P\\');
+        return preg_replace('/P\\\/', '', self::class, 1);
     }
 
     /**


faissaloux:
@nunomaduro Does the path supposed to start with uppercase? because it does anyway, in my case the real path here must be `packages\pestphp....`

If this is a bug I'll fix it on another PR.

nunomaduro:
That's good.

faissaloux:
Didn't understand! 👀

nunomaduro:
No need for more changes there.

shijunti19:
image
why
```
Note that I am not the root directory, it is the extension package packages\a031\comm\tests
```
An additional P\

PR #701, opened 2023-03-19
Fix: match() can generate false positives

mertasan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no

The following tests are failing, but their expected behavior is to pass.

In PHP, `true`, `1`, and `'1'` keys cannot coexist in the same array, as well as `false`, `0`, and `'0'` keys. Only the last assigned value will remain. Therefore, I made the changes accordingly.

## Test 1

```php
test('strict rules 1', function () {

    expect('PEST')
        ->toBeString()
        ->match(true, [
            'foo' => fn ($value) => $value->toBe('should not match'),
            true => fn ($value) => $value->toBe('PEST'),
        ]);

});
```

screenshot 22

---

## Test 2

```php
test('strict rules 2', function () {

    expect('PEST')
        ->toBeString()
        ->match(true, [
            2 => fn ($value) => $value->toBe('should not match'),
            true => fn ($value) => $value->toBe('PEST'),
        ]);

});
```

screenshot 23

---

## Test 3

```php
test('strict rules 3', function () {

    expect('PEST')
        ->toBeString()
        ->match(false, [
            false => fn ($value) => $value->toBe('PEST'),
            '' => fn ($value) => $value->toBe('should not match'),
        ]);

});
```

screenshot 24

---

screenshot 25


nunomaduro:
Going to wait until someone else reports this issue - to see if we actually consider this an issue.

Issue #682, opened 2023-03-10
Additional flags are not removed when using parallel


owenvoke:
When running Pest in parallel, and passing additional flags such as for our [GraphQL Coverage package](https://github.com/worksome/pest-graphql-coverage), the flags are not stripped as the parallel plugin doesn't seem to run plugin code.

This does not happen in standard execution (e.g. `pest --gql-coverage`).

```shell
> php vendor/bin/pest -p --gql-coverage

  The "--gql-coverage" option does not exist.
```

Blaming Luke 🤷🏻

nunomaduro:
You need to pass the flags before the --parallel flag.

owenvoke:
Hmm, weird. This worked before. Will try it on Monday. 👍🏻

owenvoke:
@nunomaduro, this doesn't work either way round. 🤔

![image](https://user-images.githubusercontent.com/1899334/226300511-665b860a-ebe3-43b1-a731-aef3896d53c2.png)

PR #703, opened 2023-03-20
[2.x] feat(parallel): Adds support for plugins to filter parallel arguments

lukeraymonddowning:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Fixed tickets | #682

Basically, `CallsHandlesArguments` is not executed fully in parallel (and rightly so). However, this means that plugins have no way of altering arguments and thus supporting parallel execution.

This PR adds a new action and contract that allows plugins to alter parallel arguments before parallel execution.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,7 +2,7 @@ ARG PHP=8.1
 FROM php:${PHP}-cli-alpine
 
 RUN apk update \
-    && apk add zip libzip-dev icu-dev
+    && apk add zip libzip-dev icu-dev git
 
 RUN docker-php-ext-configure zip
 RUN docker-php-ext-install zip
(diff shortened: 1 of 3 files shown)


owenvoke:
@lukeraymonddowning, I still don't think that this resolves the issue... 🤔 Unless I'm implementing the contract wrongly, it seems to have stripped the `--gql-coverage` flag before it even gets to the plugin.

src/Contracts/Plugins/HandlesParallelArguments.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pest\Contracts\Plugins;
+
+/**
+ * @internal
+ */
+interface HandlesParallelArguments
```
nunomaduro: Wondering it this shouldn't be `HandlesParallelWorkerArguments`?

nunomaduro (changes requested):

owenvoke (approved):

PR #721, opened 2023-03-21
[2.x] Fix `--dirty` not working on Windows

faissaloux:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | 

Due to separators `Dirty Files Filter` can't detect changed files on `Windows`.
so In this PR I have fixed it by making sure separators are the same so changed files can be found.

### DIFF
--- a/src/TestCaseFilters/GitDirtyTestCaseFilter.php
+++ b/src/TestCaseFilters/GitDirtyTestCaseFilter.php
@@ -32,6 +32,7 @@ final class GitDirtyTestCaseFilter implements TestCaseFilter
         assert(is_array($this->changedFiles));
 
         $relativePath = str_replace($this->projectRoot, '', $testCaseFilename);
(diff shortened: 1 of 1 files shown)


src/TestCaseFilters/GitDirtyTestCaseFilter.php
```diff
@@ -32,6 +32,7 @@ public function accept(string $testCaseFilename): bool
         assert(is_array($this->changedFiles));
 
         $relativePath = str_replace($this->projectRoot, '', $testCaseFilename);
+        $relativePath = str_replace(DIRECTORY_SEPARATOR, '/', $relativePath);
```
nunomaduro, line 35: @dansysanalyst Can you test this change please?

dansysanalyst: It works as expected.

CleanShot 2023-03-22 at 01 49 56@2x

owenvoke (approved):

Issue #752, opened 2023-03-30
Test in "Helpers" directory are not executed


gehrisandro:
After upgrading from v1 to v2 I run into the exact same problem already mentioned here: #734 

It worked perfectly fine with v1 but with v2 all tests in the "Helpers" directory are silently ignored. I only realized it because I tried to execute a specific test in the Helpers directory some days later.

Maybe we could add some information into the upgrade guide.

If you agree I can make a PR.

nunomaduro:
This is a bug.

gehrisandro:
Okay. I thought this is the expected behaviour. 🙃 

I will try to fix this later today.

PR #753, opened 2023-03-30
[2.x] fix: tests in Helpers directory are not executed

gehrisandro:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #752 

This PR allows to have tests in the "Helpers" directory.

The reason the tests are not executed is because all PHP files in the "Helpers" directory are loaded before building the test suite.
In this case they have a different backtrace which does not contain "overrides/Runner/TestSuiteLoader.php". Therefore I have extended the "Backtrace" class with an additional path to check for: "src/Bootstrappers/BootFiles.php"

Additionally I've added two new tests:

- `it executes tests in the Helpers directory` in "tests/Helpers/TestInHelpers.php" to ensure tests in the "Helpers" directory are executed.
- `it can use helpers from helpers directory` in "tests/Features/Helpers.php" to ensure the helpers still work (before there was no test covering this feature)

### DIFF
--- a/src/Support/Backtrace.php
+++ b/src/Support/Backtrace.php
@@ -30,7 +30,10 @@ final class Backtrace
 
             $traceFile = str_replace(DIRECTORY_SEPARATOR, '/', $trace[self::FILE]);
 
-            if (Str::endsWith($traceFile, 'overrides/Runner/TestSuiteLoader.php')) {
+            if (
+                Str::endsWith($traceFile, 'overrides/Runner/TestSuiteLoader.php') ||
+                Str::endsWith($traceFile, 'src/Bootstrappers/BootFiles.php')
+            ) {
                 break;
             }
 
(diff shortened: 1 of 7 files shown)


src/Support/Backtrace.php
```diff
@@ -30,7 +30,10 @@ public static function testFile(): string
 
             $traceFile = str_replace(DIRECTORY_SEPARATOR, '/', $trace[self::FILE]);
 
-            if (Str::endsWith($traceFile, 'overrides/Runner/TestSuiteLoader.php')) {
+            if (
+                Str::endsWith($traceFile, 'overrides/Runner/TestSuiteLoader.php') ||
+                Str::endsWith($traceFile, 'src/Bootstrappers/BootFiles.php')
```
nunomaduro, line 35: @gehrisandro can you explain me what's happening here buddy?

nunomaduro (changes requested):

PR #758, opened 2023-03-31
Skip os families

faissaloux:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

This PR adds ability to skip:

#### one os family
```php
test('test')->skipOsFamily('Windows');
```
#### multiple os families
```php
test('test')->skipOsFamily('Darwin', 'Windows');
```

### DIFF
--- a/src/PendingCalls/TestCall.php
+++ b/src/PendingCalls/TestCall.php
@@ -184,6 +184,8 @@ final class TestCall
                 );
             }
         }
+
+        return $this;
     }
 
     /**
--- a/src/PendingCalls/TestCall.php
+++ b/src/PendingCalls/TestCall.php
@@ -175,12 +175,15 @@ final class TestCall
     /**
      * Skips the current test if the given test is running on given os family.
      */
-    public function skipOsFamily(string $osFamily): self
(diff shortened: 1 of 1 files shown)


src/PendingCalls/TestCall.php
```diff
@@ -172,15 +172,28 @@ public function skip(Closure|bool|string $conditionOrMessage = true, string $mes
         return $this;
     }
 
+    /**
+     * Skips the current test if the given test is running on given os family.
+     */
+    public function skipOsFamily(string ...$osFamilies): self
```
nunomaduro, line 178: What value do I need to pass for MacOS?

nunomaduro (changes requested):

src/PendingCalls/TestCall.php
faissaloux: I think it's `Darwin`

faissaloux: https://github.com/php/php-src/blob/fced34ee1dd1af2439e88fecda299de44da4b860/main/php.h#L40-L53

PR #760, opened 2023-04-02
Add possibility to check property name and value with toHaveProperties

lucasgiovanny:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | no

This pull request proposes an enhancement to the existing code that deals with testing the presence and values of properties on an object. 

Currently, the `toHaveProperty()` and `toHaveProperties()` methods can be used to test for the existence of properties on an object. 

```php
expect($object)->toHaveProperty('name');
//or
expect($object)->toHaveProperties(['name', 'age']);
```

However, to check the value of a property, only `toHaveProperty()` is available. 

```php
expect($object)->toHaveProperty('name', 'Daniel San');
```

This PR adds the functionality of testing multiple properties and their values using the `toHaveProperties()` method using an associative array.

```php
expect($object)->toHaveProperties([
     'name' => 'Daniel San',
     'age' => 50,
]);
```

### DIFF
--- a/tests/Features/Expect/toHaveProperties.php
+++ b/tests/Features/Expect/toHaveProperties.php
@@ -4,30 +4,49 @@ use PHPUnit\Framework\ExpectationFailedException;
 
 test('pass', function () {
     $object = new stdClass();
-    $object->name = 'Jhon';
+    $object->name = 'John';
     $object->age = 21;
 
-    expect($object)->toHaveProperties(['name', 'age']);
+    expect($object)
+        ->toHaveProperties(['name', 'age'])
+        ->toHaveProperties([
+            'name' => 'John',
+            'age' => 21,
+        ]);
 });
 
 test('failures', function () {
     $object = new stdClass();
-    $object->name = 'Jhon';
+    $object->name = 'John';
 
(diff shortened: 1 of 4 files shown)


tests/Features/Expect/toHaveProperties.php
```diff
@@ -4,30 +4,49 @@
...
 
 test('failures', function () {
     $object = new stdClass();
-    $object->name = 'Jhon';
-
-    expect($object)->toHaveProperties(['name', 'age']);
+    $object->name = 'John';
+
+    expect($object)
+        ->toHaveProperties(['name', 'age'])
```
fabio-ivona, line 23: @lucasgiovanny maybe this group of failing tests could be split into two separate tests ? The second expectation could not fail and the exception will be thrown anyway because of the first one

fabio-ivona (changes requested):

jameshulse:
Is it plausible that this feature could be expanded at some stage to architecture testing? The use-case would be asserting that all classes in our "Actions" namespace have a single public `execute()` method.

PR #770, opened 2023-04-06
Update stub extensions

ganyicz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no

This PR fixes an issue with autocomplete inside TestCase class and PHPUnit tests when using VS Code with Intelephense.

Steps to reproduce:
1. Setup fresh Laravel installation
2. Install Pest
3. Try to use any method defined on Illuminate\TestCase

Screenshot 2023-04-06 at 20 02 39

VS Code doesn't pick up any methods defined inside base Laravel TestCase and only suggests methods defined inside base PHPUnit TestCase. This issue disappears after removing / renaming TestCase stubs in vendor/pestphp/pest/stubs

I guess this is caused by Intelephense picking up the non-laravel TestCase as a source for the method definitions elsewhere in the app.


nunomaduro:
Can you make all files stubs?

owenvoke:
I'd also say it's quite nice having `.*.stub` 👀 That way it's kind of consistent, and also allows you to assign the file type in an IDE (e.g. map `*.php.stub` to PHP)

So, rather than `TestCase.stub` it would be `TestCase.php.stub` 👍🏻

ganyicz:
I have updated all files to `.*.stub`

Not sure about the test stubs (Browser / Dataset / Feature / Unit) I haven't found a reference to them

owenvoke (approved):

ganyicz:
Any chance this could get merged? The checks were failing upstream 

Thanks

Issue #746, opened 2023-03-28
PCOV integration is not working


MaurizioMoreo:
Even if this test is green:
```
    public function testPcovEnabled(): void
    {
        $this->assertSame('1', ini_get('pcov.enabled'));
        $this->assertTrue(extension_loaded('pcov'));
    }
```
if I run the tests with the `--coverage` option, I've got this error:
```
 ERROR  Unable to get coverage using Xdebug. Did you set Xdebug's coverage mode?
 ```
 It seems related to `\Pest\Support\Coverage::isAvailable`. Am I missing something?

owenvoke:
PCov seems to work fine for me on PHP 8.2 🤔 That kind of sounds like both Xdebug and PCov are running at the same time, and Xdebug is taking precedence.

MaurizioMoreo:
In this case we could maybe consider to check if Xdebug is enabled: should we consider whether it is the enabling of a component that determines precedence? @nunomaduro make sense?

MaurizioMoreo:
We need another approval to merge the #772 PR.

MaurizioMoreo:
The problem is solved in [_v2.5.1_](https://github.com/pestphp/pest/commits/v2.5.1), but I see that the `cacheDirectory` XML attribute for the `coverage` element is ignored. Should we open a new issue @owenvoke?

nunomaduro:
Yes plz.

PR #772, opened 2023-04-07
Add PCOV availability check

MaurizioMoreo:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #746

This PR fixes the Xdebug failure when PCOV is enabled.

### DIFF
--- a/src/Support/Coverage.php
+++ b/src/Support/Coverage.php
@@ -46,6 +46,10 @@ final class Coverage
             return true;
         }
 
+        if ($runtime->hasPHPDBGCodeCoverage()) {
+            return true;
+        }
+
         if (! $runtime->hasXdebug()) {
             return true;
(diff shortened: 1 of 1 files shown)


src/Support/Coverage.php
```diff
@@ -42,6 +42,10 @@ public static function isAvailable(): bool
             return false;
         }
 
+        if ($runtime->hasPCOV()) {
+            return true;
```
owenvoke, line 46: Think maybe we should also add a check for PHPDBG?

MaurizioMoreo: Make sense, I'd improve readability also in that case but I'll try to change the minimum.

owenvoke (approved):

PR #797, opened 2023-05-01
[2.x] allow to expect `toThrow` against an exception instance

fabio-ivona:
This PR implements Throwable support as argument of a `toThrow` expectation. This will allow to:

```php
expect(fn() => $user->subscribe($team))
        ->toThrow(new TeamException("user is already subscribed to team"));
```

in this case, the expectation will pass if the Throwable is an instance of `TeamException` and has `user is already subscribed to team` as exception message

this will allow a fast check when using static exception constructors like the following one, without the need to enter the code and check the exception message: 

```php
expect(fn() => $user->subscribe($team))
        ->toThrow(TeamException::alreadySubscribed());
```

### DIFF
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -842,7 +842,7 @@ final class Expectation
      * @param (Closure(Throwable): mixed)|string $exception
      * @return self<TValue>
      */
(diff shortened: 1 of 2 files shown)


nunomaduro (approved):

nunomaduro:
Good job here. Can you add docs?

PR #805, opened 2023-05-06
Adding ability to set @define-env annotation for Orchestra/Testbench

titantwentyone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

This PR enables the ability to integrate Orchestra Testbench environment annotation @define-env. Usage as per test:

```php
function my_environment($app)
{
    //define environment
}

test('it will do soemthing', function() {
    
    //test

})->environment('my_environment');
```
I did look at doing this as a plugin but I'm not sure this is possible? If it is, please let me know and I will try and rework as such. If not, I can add the '@define-routes' annotation also in a future PR.

### DIFF
--- a/tests/Features/Testbench.php
+++ b/tests/Features/Testbench.php
@@ -1,0 +1,10 @@
+<?php
+
+function my_environment($app)
+{
+}
+
+test('it will add a define-env annotation', function() {
+    $phpDoc = (new ReflectionClass($this))->getMethod($this->name());
+    expect(str_contains($phpDoc->getDocComment(), '* @define-env my_environment'))->toBeTrue();
+})->environment('my_environment');
--- a/tests/Features/Testbench.php
+++ b/tests/Features/Testbench.php
@@ -4,7 +4,7 @@ function my_environment($app)
 {
 }
 
-test('it will add a define-env annotation', function() {
+test('it will add a define-env annotation', function () {
     $phpDoc = (new ReflectionClass($this))->getMethod($this->name());
     expect(str_contains($phpDoc->getDocComment(), '* @define-env my_environment'))->toBeTrue();
 })->environment('my_environment');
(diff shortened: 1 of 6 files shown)


tests/Features/Testbench.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+function my_environment($app)
+{
+}
+
+test('it will add a define-env annotation', function () {
+    $phpDoc = (new ReflectionClass($this))->getMethod($this->name());
+    expect(str_contains($phpDoc->getDocComment(), '* @define-env my_environment'))->toBeTrue();
```
owenvoke: We can simplify this test 🤷🏻
```suggestion
    expect($phpDoc->getDocComment())->toContain('* @define-env my_environment');
```

owenvoke (commented):
I feel like this would definitely make more sense as part of the Laravel plugin, or a separate plugin. However, I'm not sure we support adding annotations via plugins. 🤔

titantwentyone:
Have amended test.

I did look around for a way to hook into the annotations but couldn't see anything obvious. Would be nice to have an annotations API perhaps? Could be used within plugins or even provide the ability to define ad-hoc annotations per test?

EDIT: Just having a think but this could be reworked to provide a generalised approach to adding annotations. Something like:

```php
test('something', function() {
   //test
})->annotations([
   '@define-env my_environment',
   '...something else...'
]);
```

titantwentyone:
Since testbench needs a class method fot the envionment setup, the recent commit wraps the envronment function into such a method.

This works great for me but I'm moving this to a draft PR as @owenvoke is right in that this is too specific to Laravel. However there is no way to hook into the annotations. I do think an `annotations` API as described above would be useful along with a `addClassMethod` method to address the addition of a class method. This would be more generic and could be used within plugins/on an ad-hoc basis:

```php
addClassMethod([
    'my_environment' => function($app) {
        //set up enviroment
    }
]);

test('something', function() {
    //test
})->annotations([
    '@define-env my_environment'
]);
```
With these two things in place, a plugin (hopefully - I need to read more into the avilable API) could be written which supports something like:

```php
test('something', function() {
   //test
})->environment(function($app) {
   //environment
});
```

If this is something that would be considered, I can draft a PR.

nunomaduro:
No plans for this.

joaopalopes24:
@nunomaduro Any expectations of implementing this?

PR #844, opened 2023-06-20
[2.x] Make dataset method return values by name

hafezdivandari:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | -

This PR makes it possible to get the specified dataset values by its name:

```php
dataset('foo', [1, 2]);

test('can get dataset values', function () {
    expect(dataset('foo'))->toBe([1, 2]);
});
```

### DIFF
--- a/src/Functions.php
+++ b/src/Functions.php
@@ -57,10 +57,15 @@ if (! function_exists('dataset')) {
     /**
      * Registers the given dataset.
      *
-     * @param  Closure|iterable<int|string, mixed>  $dataset
+     * @param  Closure|iterable<int|string, mixed>|null  $dataset
+     * @return array|void
      */
-    function dataset(string $name, Closure|iterable $dataset): void
+    function dataset(string $name, Closure|iterable $dataset = null)
     {
+        if ($dataset === null) {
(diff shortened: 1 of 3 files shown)


src/Functions.php
```diff
@@ -57,10 +57,15 @@ function beforeEach(Closure $closure = null): BeforeEachCall
     /**
      * Registers the given dataset.
      *
-     * @param  Closure|iterable<int|string, mixed>  $dataset
+     * @param  Closure|iterable<int|string, mixed>|null  $dataset
+     * @return array|void
      */
-    function dataset(string $name, Closure|iterable $dataset): void
+    function dataset(string $name, Closure|iterable $dataset = null)
```
paulshryock, line 63: Should this still have a return type hint?

```php
function dataset(string $name, Closure|iterable $dataset = null): array|void
```

hafezdivandari: Void can only be used as a standalone type.

paulshryock: Of course; my bad.

nunomaduro:
No plans for this.

Issue #852, opened 2023-07-09
`toHaveProperties` does not handle nested array


abdul-alhasany:
Hi
I am adding tests for my [WordPress plugin](https://wordpress.org/plugins/nashaat-activity-log/) and using `toHaveProperties` to check database value.

Since the database value is an array, I am using this test code:
```php
expect($edit_comment) ->toHaveProperties(array(
				'event'    => 'edited',
				'context'  => 'comment',
				'log_info' => array(
					'post_id'   => '1',
					'post_type' => 'post',
				),
			));
```
But it fails with an error stating that `log_info` is missing an item. 

I checked the source code for Pest and it seems that `toHaveProperties` only check for top-level properties.

I am happy to submit a PR if you consider updating this method or creating a new one.

fabio-ivona:
Hi! Feel free to open a PR for it 😉

abdul-alhasany:
Thanks. I forked the repo and ran the tests (without changing anything) but 2 gave failed. 

Tests\Visual\TeamCity > visual snapshot of team city with ('SuccessOnly.php'…   
  Failed asserting that two strings are equal.
Tests:    2 failed, 11 skipped, 2 passed (4 assertions)

Also, when I tried to run `composer update:snapshots` I received this error:
```bash
'REBUILD_SNAPSHOTS' is not recognized as an internal or external command,
operable program or batch file.
```

fabio-ivona:
You can ignore integration tests, just run test:unit and test:parallel

abdul-alhasany:
@fabio-ivona Sorry I did not read your message earlier. That is what I ended up doing anyway. I already submitted a PR.

nunomaduro:
No plans for this.

PR #855, opened 2023-07-10
[2.x] Add support for nested properties for `toHaveProperties`

abdul-alhasany:
…ation

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #852 

This PR adds the ability for `toHaveProperties` expectation to handle nested objects and arrays. Extra tests have been added to check nested properties.

It should not break existing code.

### DIFF
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -294,7 +294,7 @@ final class Expectation
      * This has been abstracted out so that it can be called recursively.
      *
      * @param  iterable<array-key, mixed>  $incoming The incoming array
-     * @param  iterable<array-key, mixed>  $expected The expected array
+     * @param  array $expected The expected array
      * @param  string  $message The message to display if the assertion fails
      */
     private function assert_object(mixed $incoming, iterable $expected, string $message): void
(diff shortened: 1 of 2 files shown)


src/Mixins/Expectation.php
```diff
@@ -289,16 +289,63 @@ public function toHaveProperty(string $name, mixed $value = new Any(), string $m
...
 
     /**
-     * Asserts that the value contains the provided properties $names.
+     * Callback for toHaveProperties.
+     *
+     * This has been abstracted out so that it can be called recursively.
+     *
+     * @param  iterable<array-key, mixed>  $incoming The incoming array
+     * @param  array $expected The expected array
+     * @param  string  $message The message to display if the assertion fails
```
devajmeireles, line 298: There is no `@return void` missing here? 🤔

tests/Features/Expect/toHaveProperties.php
```diff
@@ -2,7 +2,7 @@
 
 use PHPUnit\Framework\ExpectationFailedException;
 
-test('pass', function () {
+test('pass objects', function () {
```
devajmeireles, line 5: I don't think we should transform the tests, but rather create new ones. Does this perhaps make way for: `toHavePropertyObject` with all that logic instead of changing `toHaveProperty`?

nunomaduro:
No plans for this.

PR #860, opened 2023-07-17
Feature: Introducing The Ability to Dump Conditionally

devajmeireles:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes
| Fixed tickets | -. 

## What

This PR aims to introduce the `ddWhen` and `ddUnless` to debug conditionally.

## Reason

Sometimes you might be working with `with` to pass data like `dataset`, but if you need to debug the behavior of the test it will give you an effort to do that . With `ddWhen` or `ddUnless` it will be very easy.

---

Hope this helps,

Thanks. ❤️

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -119,6 +119,48 @@ final class Expectation
     }
 
     /**
+     * Dump the expectation value when the result of the condition is truthy.
+     *
+     * @param  bool  $boolean
+     * @return never
+     */
+    public function ddWhen($boolean, mixed ...$arguments): void
+    {
+        if (! $boolean) {
+            return;
+        }
+
+        if (function_exists('dd')) {
+            dd($this->value, ...$arguments);
+        }
+
+        var_dump($this->value);
+
+        exit(1);
+    }
+
+    /**
+     * Dump the expectation value when the result of the condition is falsy.
+     *
+     * @param  bool  $boolean
+     * @return never
+     */
+    public function ddUnless($boolean, mixed ...$arguments): void
+    {
+        if ($boolean) {
+            return;
+        }
+
+        if (function_exists('dd')) {
+            dd($this->value, ...$arguments);
+        }
+
+        var_dump($this->value);
+
+        exit(1);
+    }
+
+    /**
      * Send the expectation value to Ray along with all given arguments.
      *
      * @return self<TValue>
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -130,13 +130,7 @@ final class Expectation
             return;
         }
 
-        if (function_exists('dd')) {
-            dd($this->value, ...$arguments);
-        }
-
-        var_dump($this->value);
-
-        exit(1);
+        $this->dd($this->value, ...$arguments);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Expectation.php
```diff
@@ -118,6 +118,48 @@ public function dd(mixed ...$arguments): void
...
+        }
+
+        if (function_exists('dd')) {
+            dd($this->value, ...$arguments);
+        }
+
+        var_dump($this->value);
+
+        exit(1);
+    }
```
faissaloux:
```suggestion
    public function ddUnless($boolean, mixed ...$arguments): void
    {
        $this->ddWhen(! $boolean, ...$arguments);
    }
```

faissaloux (changes requested):

src/Expectation.php
devajmeireles: Hey, @faissaloux ! Thank you for your suggestion. I'll apply this without accept the suggestion here because you removed the docblock.

faissaloux: Actually I didn't, the docblock still there.

It's okay anyways.

src/Expectation.php
```diff
@@ -118,6 +118,38 @@ public function dd(mixed ...$arguments): void
...
+     * @return never
+     */
+    public function ddWhen($boolean, mixed ...$arguments): void
+    {
+        if (! $boolean) {
+            return;
+        }
+
+        if (function_exists('dd')) {
+            dd($this->value, ...$arguments);
```
nunomaduro: Shouldn't this call our own `->dd` internally?

nunomaduro (changes requested):

src/Expectation.php
devajmeireles: Good captcha. I've updated it. Thank you!

src/Expectation.php
```diff
@@ -118,6 +118,32 @@ public function dd(mixed ...$arguments): void
         exit(1);
     }
 
+    /**
+     * Dump the expectation value when the result of the condition is truthy.
+     *
+     * @param  bool  $boolean
+     * @return never
+     */
+    public function ddWhen($boolean, mixed ...$arguments): void
```
faissaloux, line 127:
```suggestion
    public function ddWhen(bool $boolean, mixed ...$arguments): void
```

src/Expectation.php
```diff
@@ -118,6 +118,32 @@ public function dd(mixed ...$arguments): void
...
+        $this->dd($this->value, ...$arguments);
+    }
+
+    /**
+     * Dump the expectation value when the result of the condition is falsy.
+     *
+     * @param  bool  $boolean
+     * @return never
+     */
+    public function ddUnless($boolean, mixed ...$arguments): void
```
faissaloux, line 142:
```suggestion
    public function ddUnless(bool $boolean, mixed ...$arguments): void
```

faissaloux (changes requested):

src/Expectation.php
devajmeireles: Hey! @faissaloux . Thank you, but according to the all code style of Pest I think this can't be inserted here.

src/Expectation.php
devajmeireles: > 

Hey! @faissaloux . Thank you, but according to the all code style of Pest I think this can't be inserted here.

PR #872, opened 2023-07-24
[2.x] Add null ability to `and` method in Expectation

milwad-dev:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

We always use the `and` method like this:

```php
expect($name)
    ->toBe('milwad')
    ->and($name)
    ->toBeString();
```

But it's very impressive to use the `and` method like this:

```php
expect($name)
    ->toBe('milwad')
    ->and()
    ->toBeString();
```

Also, you can use `and` without parentheses:

```php
expect($name)->toBe('milwad')->and->toBeString();
```

Here we didn't need to pass the first value for the `and` method and changes are easier than the first way.

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -71,8 +71,12 @@ final class Expectation
      * @param  TAndValue  $value
      * @return self<TAndValue>
      */
-    public function and(mixed $value): Expectation
+    public function and(mixed $value = null): Expectation
     {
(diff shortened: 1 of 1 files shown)


nunomaduro:
Sorry, but this would a breaking change. Cuz the given value may be actually `null`.

PR #881, opened 2023-07-27
[2.x] Multiple snapshots

### LINKED ISSUE #871: `toMatchSnapshot` only supports one snapshot per test
It seems that `toMatchSnapshot` only supports one snapshot per test. I would have expected that either the index of the call or the name that is passed to the method will be used to generate the snapshot name.

fabio-ivona:
This PR will 

- [x] Cleanup `SnapshotRepository` code, as TestCase is retrieved inside the repository
- [x] Implement a snapshot expectations counter to keep track of multiple snapshots in the same test

**note** found an issue: when the first snapshot is created, a `$testCase::markTestIncomplete()` is called inside `toMatchSnapshot` expectation (Mixin/Expectation.php L837). This will stop the execution and prevent the creation of the second snapshot.

I commented out that line for now, resulting in a plain Risky test instead of an Incomplete one. Any idea on how to workaround on this?


src/Repositories/SnapshotRepository.php
```diff
@@ -25,11 +24,9 @@ public function __construct(
     /**
      * Checks if the snapshot exists.
      */
-    public function has(TestCase $testCase, string $description): bool
+    public function has(): bool
```
fabio-ivona, line 30: @nunomaduro maybe we shall rename this one to `exists`?

src/Mixins/Expectation.php
```diff
@@ -817,22 +823,18 @@ public function toMatchSnapshot(string $message = ''): self
...
                 $content,
                 $string,
                 $message === '' ? "Failed asserting that the string value matches its snapshot ($filename)." : $message
             );
         } else {
-            $filename = $snapshots->save($testCase, $string);
+            $filename = $snapshots->save($string);
 
-            $testCase::markTestIncomplete('Snapshot created at ['.$filename.'].');
+            //$testCase::markTestIncomplete('Snapshot created at ['.$filename.'].');
```
fabio-ivona: This one causes the test to end prematurely, preventing the creation of the second snapshot, I had to comment on it for now, any idea for a workaround?

src/Repositories/SnapshotRepository.php
nunomaduro: @fabio-ivona Yes.

src/Mixins/Expectation.php
nunomaduro: @freekmurze What do you do in your package here?

fabio-ivona: fixed

PR #890, opened 2023-08-01
Add .gitkeep in Tests folders

soymgomez:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

When init "Pest" I add some .gitkeep files to the test folders to make sure that these folders are comitted even if they don't have tests inside them.


enkhjargal (approved):

owenvoke:
Personally I don't think this makes sense, as the `ExampleTest.php` files act as a Gitkeep. 🤷🏻‍♂️

soymgomez:
They work until the sample tests are deleted, which is one of the first things we do in the projects.

And if you make a commit with these tests deleted and you have a working pipeline it won't pass because Pest always expects this folder structure.

I think it would be a good idea to leave the .gitkeep folders there in case something happens.

devajmeireles:
> They work until the sample tests are deleted, which is one of the first things we do in the projects.
> 
> And if you make a commit with these tests deleted and you have a working pipeline it won't pass because Pest always expects this folder structure.
> 
> I think it would be a good idea to leave the .gitkeep folders there in case something happens.

Thank you for your PR. I think this is unnecessary. Why will I remove files from the `Unit` folder but keep the `Unit` folder only with a single file (.gitkeep) if I can control when I have the `Unit` folder or not by editing `phpunit.xml`?

For a better organization, I will close this PR. @nunomaduro , will be notified and will be able to restore it if he agrees to keep the PR.

PR #891, opened 2023-08-01
Add `toBeInvokable` arch expectation

ash-jc-allen:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes
| Fixed tickets | N/A

Hey!

This PR is just a small one, but something that I think could be useful for some devs (including myself). This PR proposes a new `toBeInvokable` expectation that can be used to assert whether a class has an `__invoke` method.

I like to use invokable classes quite a bit (for things like action classes). An example action class might be something like this:

```php
use App\DataTransferObjects\User\NewUserData;

class CreateUser
{
    public function __invoke(NewUserData $userData): void
    {
        // ...
    }
}
```

And then it can be called like so in my controller:

```php
public function store(StoreUserRequest $request, CreateUser $createUser)
{
    // ...

    $createUser($request->toDto());

    // ...
}
```

I like to follow this pattern and like knowing that all my actions are invokable. So I thought it'd be handy to have an expectation that I can use to enforce this.

The expectation can be used like so:

```php
test('class is invokable')
    ->expect('App\Actions')
    ->toBeInvokable();
```

I'm not entirely sure if I've done everything right here, but I've tried my best to make sure there are tests covering the code. If I've done anything wrong though, please give me a shout!

Likewise, if this is something you think might be useful to other devs but you want something changing first, let me know and I'll get to it ASAP! 😄

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -701,7 +701,7 @@ final class Expectation
     {
         return Targeted::make(
             $this,
-            fn(ObjectDescription $object): bool => $object->reflectionClass->hasMethod('__invoke'),
+            fn (ObjectDescription $object): bool => $object->reflectionClass->hasMethod('__invoke'),
             'to be invokable',
             FileLineFinder::where(fn (string $line): bool => str_contains($line, 'class'))
         );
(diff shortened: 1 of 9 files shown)


JonPurvis (approved):

nunomaduro:
@ash-jc-allen can you add docs to this please?

ash-jc-allen:
Hey @nunomaduro! Yeah sure, I'll get some written now 😄

PR #895, opened 2023-08-02
[2.x] Add support for nested traversable in sequence

cerbero90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

This PR adds support for nested `Traversable`s when using the `sequence()` expectation.

The previous logic was trying to traverse an already closed generator due to the use of `iterator_to_array()`.

The new logic performs only one loop of the `Traversable` avoiding to loop through an already run generator or iterator.


tests/Features/Expect/sequence.php
```diff
@@ -40,7 +42,7 @@ function ($expectation) {
     expect(static::getCount())->toBe(16);
 });
 
-test('fails if the number of iterable items is greater than the number of expectations', function () {
```
nunomaduro, line 43: Why this test changed?

nunomaduro (changes requested):

tests/Features/Expect/sequence.php
cerbero90: there is a small mistake in the test description: when using `sequence()` we are allowed to have less callbacks than the number of iterable items.

we can even have just one callback which is called for every item.

we cannot - however - have less items than the callbacks, in that case it should fail.

the `sequence()` expectation is already behaving that way, there is just a little typo in the test description that is misleading 👍 

also, we are throwing `OutOfRangeException` instead of running `Assert::assertLessThanOrEqual()` because the latter shows a confusing error and runs an assertion that the end-user is not explicitly asking for. on the other hand the thrown exception is exhaustive and tells the user why the expectation failed 👍

nunomaduro:
@cerbero90 Can you resolve conflicts?

cerbero90:
@nunomaduro done

PR #898, opened 2023-08-05
TestSuiteLoader will always consider classes from the current file

dylanbr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no

I would like to have test classes which extend from other test classes. This allows for a set of test cases to run in the parent class, then the child class can override certain methods, and have the same test cases run.

This works with PHPUnit 10, but when running the same tests via Pest, some of the parent classes are ignored. This happens because of changes in the override to `TestSuiteLoader`. Once a class is loaded, it's not considered again, so if the child class is loaded before the parent, the parent will never be considered for test cases.

This PR is an attempt to resolve this, but I'd would also like to ask if there is any reason this should not be allowed?

### DIFF
--- a/overrides/Runner/TestSuiteLoader.php
+++ b/overrides/Runner/TestSuiteLoader.php
@@ -61,6 +61,11 @@ final class TestSuiteLoader
     private static array $loadedClasses = [];
 
     /**
+     * @psalm-var array<string, array<class-string>>
+     */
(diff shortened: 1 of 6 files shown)


dylanbr:
Thanks for the quick review and merge @nunomaduro 👍

PR #907, opened 2023-08-13
Add toBePluralOf expectation

JonPurvis:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | N/A

# Introduction 
This PR adds a new Expectation, to check that a value, is the pluralized version of another value.

## Usage
```php
it('succeeds as dogs is the plural of dog', function () {
    expect('dogs')->toBePluralOf('dog');
});

it('succeeds as loaves is the plural of loaf', function () {
    expect('loaves')->toBePluralOf('loaf');
});

it('succeeds as cat is not the plural of cats', function () {
    expect('cat')->not->toBePluralOf('cats');
});
```

Laravel handles pluralization incredibly well, so I've taken the bulk of the Pluralization class from that, and adapted it as needed to work with Pest. Thanks to datasets, I'm then able to validate a good handful of words to make sure they are correct.

Documentation PR: https://github.com/pestphp/docs/pull/212


19950512:
Desnecessário

owenvoke (commented):
I feel this may work better as a plugin (maybe a bit overkill for one new method), as this adds quite a bit of new code to maintain. If it was a plugin as well, it could just require the Illuminate Support package to handle this (meaning only one place needs to be maintained).

JonPurvis:
Hey,

Funnily enough, I actually did think about making it a plugin as I wanted to follow Luke Downing's Larabit, but then I too also thought it seemed a bit much just to introduce one method.

~~I guess from a maintenance perspective, the Pluralizer class in Laravel was last updated 9 years ago (closer to 10 years really), so I don't think this is something that would require any ongoing maintenance, I think any problems with the class would've been picked up over the years.~~

Turns out, I was looking at the class from a previous version 🙈 

I'd be happy to revisit this as a plugin if that's the general consensus though 😄 Just not sure it's plugin worthy!

fabio-ivona:
I see an issue here, the pluralizer works only for English strings, and it may lead to unwanted false positives/negatives in case of another language

this inconsistency could better be handled in a dedicated plugin that may also open up to other localizations

JonPurvis:
That's a fair point @fabio-ivona,

I'm going to mark this as a draft for now, I have 1 idea I want to play around with later on today, but I may just end up going down the Plugin route

JonPurvis:
Thanks @owenvoke and @fabio-ivona for taking a look over this, I think I'm going to down the plugin route as I still see some benefit to having it as a expectation. You've given me some stuff to think about, especially when it comes to multi-language so I'm going to close this PR and work on it separately. 

Thanks! 🍻

PR #911, opened 2023-08-14
[2.x] Introducing `toBeDigits` Expectation

devajmeireles:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

Considering the `toBeNumeric` expectation for cases where we have dots, for example, we succeed in the test case, because behind the scenes we use `is_numeric` [PHP function](https://www.php.net/manual/en/function.is-numeric.php) for the test:

image

This PR aims to introduce the `toBeDigits` expectation to Pest, causing the above test cases to fail due to the fact that we use the `ctype_digits` [PHP function](https://www.php.net/manual/pt_BR/function.ctype-digit.php), which will refuse the test due to the fact that there is a dot sign inside the represented number like string:

image

### DIFF
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -503,6 +503,18 @@ final class Expectation
     }
 
     /**
+     * Asserts that the value contains only digits.
+     *
+     * @return self<TValue>
+     */
+    public function toBeDigits(string $message = ''): self
+    {
+        Assert::assertTrue(ctype_digit((string) $this->value), $message);
+
+        return $this;
+    }
+
+    /**
      * Asserts that the value is of type object.
      *
      * @return self<TValue>
--- a/tests/Features/Expect/toBeDigits.php
+++ b/tests/Features/Expect/toBeDigits.php
@@ -1,0 +1,20 @@
+<?php
+
+use PHPUnit\Framework\ExpectationFailedException;
+
+test('pass', function () {
+    expect('123')->toBeDigits();
+    expect('123.14')->not->toBeDigits();
+});
+
+test('failures', function () {
+    expect('123.14')->toBeDigits();
+})->throws(ExpectationFailedException::class);
+
+test('failures with custom message', function () {
+    expect('123.14')->toBeDigits('oh no!');
+})->throws(ExpectationFailedException::class, 'oh no!');
+
+test('not failures', function () {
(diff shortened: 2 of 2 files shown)


19950512:
acredito que isso não seja necessário em vista que o numeric já em em vista estes pontos. É mais do mesmo.

devajmeireles:
> acredito que isso não seja necessário em vista que o numeric já em em vista estes pontos. É mais do mesmo.

Thank you for your review. Since Pest wants to achieve the ultimate test excellence, IMO this method is important for situations where we want to test safely, and the `is_numeric` function would not evaluate the entire character set, leading us to believe that "123.14 " is a number, when in certain situations we might want to know if it's just digits. The main idea we need to keep in mind to review this new expectation is this: _"this is only digits?"_

src/Mixins/Expectation.php
```diff
@@ -502,6 +502,18 @@ public function toBeNumeric(string $message = ''): self
         return $this;
     }
 
+    /**
+     * Asserts that the value contains only digits.
+     *
+     * @return self<TValue>
+     */
+    public function toBeDigits(string $message = ''): self
```
nunomaduro, line 510: Does Pest or Vitest have an equivalent on this?

nunomaduro (changes requested):

src/Mixins/Expectation.php
devajmeireles: > Does Pest or Vitest have an equivalent on this?

According to my review, none of them have an equivalent for this natively, including as Pest's `toBeNumeric` is using [PHP is_numeric](https://www.php.net/manual/en/function.is-numeric.php) the test below will result in success:

```php
expect('9001 ')->toBeNumeric(); //space
expect('1.23e5')->toBeNumeric(); //dot
expect('-35')->toBeNumeric(); //negative
```

**Manually**, the alternative to this would be:

```php
expect(ctype_digit('9001 '))->toBeTrue(); //false
expect(ctype_digit('1.23e5'))->toBeTrue(); //false
expect(ctype_digit('-35'))->toBeTrue(); //false
```

With addition of `toBeDigits` we have ensure that the string contains only digits.

devajmeireles:
If accepted, here is the doc: https://github.com/pestphp/docs/pull/213

PR #877, opened 2023-07-25
[2.x] Print only files below the min coverage

WendellAdriel:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes
| Fixed tickets |

This PR is adding a new `--errors-only` flag that when passed to the command call will print only the files that have coverage below the value set in with the `--min` argument

PR #919, opened 2023-08-18
[2.x] Print only files below the min coverage

WendellAdriel:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

This PR is adding a new `--errors-only` flag that when passed to the command call will print only the files that have coverage below the value set in with the `--min` argument


WendellAdriel:
@devajmeireles here's the PR with the code in sync with the 2.x branch.
Follow-up of the PR #877

devajmeireles (approved):

nunomaduro:
Had to revert this pull request. Seems that, if no --min is specified, no coverage of files is displayed at all.

PR #932, opened 2023-08-23
[2.x] Add `toBeEnumeration` Expectation

devajmeireles:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

Focusing on PHP 8.1, we have [Enumerations](https://www.php.net/manual/en/language.types.enumerations.php), something that is increasingly part of the ecosystem of well-built applications nowadays. This PR aims to introduce the `toBeEnumeration` expectation, allowing something like that:

```php
enum FooBar: string
{
    case Foo = 'foo';
    case Bar = 'bar';
}

expect('foo')->toBeEnumeration(FooBar::class); // ✅ 
expect(FooBar::Foo)->toBeEnumeration(FooBar::class); // ✅ 
expect('baz')->toBeEnumeration(FooBar::class); // ❌  
```

Behind the scenes, the `toBeEnumeration` will use the `tryFrom` from Enumerations combined with `assertNotNull` and `assertInstanceOf` to ensure that the value is part of the Enumeration.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,6 +43,7 @@
             "Tests\\Fixtures\\Covers\\": "tests/Fixtures/Covers",
             "Tests\\Fixtures\\Inheritance\\": "tests/Fixtures/Inheritance",
             "Tests\\Fixtures\\Arch\\": "tests/Fixtures/Arch",
+            "Tests\\Fixtures\\Features\\": "tests/Fixtures/Features",
             "Tests\\": "tests/PHPUnit/"
         },
         "files": [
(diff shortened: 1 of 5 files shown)


JonPurvis:
Nice work on this @devajmeireles, I'd love to see this get added to Pest. I've been meaning to look at this for a while 😆 

It'd definitely be a useful addition!

owenvoke (approved):
Love this! 😍 Guess the snapshot needs to be updated?

nunomaduro:
We've an `toBeEnum` already. Any chance you can combine both?

owenvoke:
Didn't realise we have `toBeEnum()`, looking at the code behind that, it probably makes more sense? 🤔

PR #948, opened 2023-09-05
[2.x] Add `toBeUuid` expectation

hungthai1401:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

This PR adds `toBeUUID` expectation.

Usage:

```php
expect('00112233-4455-8677-8899-aabbccddeeff')->toBeUUID();
```


devajmeireles (approved):
Thanks for your PR.

nexxai:
I have a decent amount of experience with UUIDs and this all looks decent to me.  I can't think of any edge cases that this regex wouldn't catch off the top of my head.

src/Mixins/Expectation.php
```diff
@@ -1111,4 +1112,20 @@ public function toBeStudlyCase(string $message = ''): self
 
         return $this;
     }
+
+    /**
+     * Asserts that the value is UUID.
+     *
+     * @return self<TValue>
+     */
+    public function toBeUUID(string $message = ''): self
```
nunomaduro:
```suggestion
    public function toBeUuid(string $message = ''): self
```

nunomaduro (changes requested):

inxilpro:
I believe a slightly better regex would be:

`/^(?:[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}|00000000-0000-0000-0000-000000000000)$/iD`

This accounts for known versions (1–5) and variants (8, 9, a, or b), while still allowing the NIL UUID. 

This pattern comes from [uuidjs](https://github.com/uuidjs/uuid)

hungthai1401:
@inxilpro I think the current regex is good. This regex pattern comes from [laravel](https://github.com/laravel/framework/blob/52b301797d7a229bf7ebd3e2eefc762487738578/src/Illuminate/Support/Str.php#L519)

gabbanaesteban:
There are different versions of UUIDs.

I would suggest using the approach used by https://www.npmjs.com/package/validator

Here is the validation https://github.com/validatorjs/validator.js/blob/master/src/lib/isUUID.js

nunomaduro (approved):

nunomaduro:
@nexxai Can you PR docs?

hungthai1401:
@nunomaduro I've created the documentation related to that: https://github.com/pestphp/docs/pull/229

Issue #936, opened 2023-08-24
[Bug]: Test Dependency not working on `describe` blocks


lidani:
### What Happened

when I made an dependency test inside a describe block it fails

### How to Reproduce

```php
// MercadoLivreAnnouncesTest.php
 
describe('mercado livre announces service', function () {

  it('should serialize an announce', function () {
    expect(true)->toBeTrue();
    return 'example-value';
  });

 it('should create an announce', function ($serialized) {
    var_dump($serialized); // should dump 'example-value'
    expect(true)->toBeTrue();
  })->depends('it should serialize an announce');

});

```

<img width="1503" alt="image" src="https://github.com/pestphp/pest/assets/29063458/885b4fb7-d9d5-4b5c-bc05-964dae81a352">

apparently, pest gets lost in the second test, even printing the wrong name.
the line 
`⨯ mercado livre announces service → it should serialize an announce`
should be something like 
`✓ mercado livre announces service → it should create an announce`

and when I remove the describe block

```php
it('should serialize an announce', function () {
  expect(true)->toBeTrue();
  return 'example-value';
});

it('should create an announce', function ($serialized) {
  var_dump($serialized); // should dump 'example-value'
  expect(true)->toBeTrue();
})->depends('it should serialize an announce');
```

it works as expected

<img width="500" alt="image" src="https://github.com/pestphp/pest/assets/29063458/cf0521c2-c6a2-45f6-9759-a5bdfddcf4e0">

### Pest Version

2.16

### PHP Version

8.2.9

### Operation System

macOS, Linux

### Notes

the error occurs both in my Mac and my alpine linux docker container

devajmeireles:
Hey, @lidani . Thanks for your report.

I was able to reproduce that. I will report this to the Pest team.

lidani:
perfect, thank you so much

devajmeireles:
Hey, @lidani ! This PR: https://github.com/pestphp/pest/pull/949 was created to fix this issue.

PR #949, opened 2023-09-05
[2.x] Fix the Usage of `depends` With `describe`

devajmeireles:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

This PR aims to introduce the ability to use `describe` with `depends`:

```php
describe('depends on describe', function () {
    test('foo', function () {
        expect('foo')->toBe('foo');
    });

    test('bar', function () {
        expect('bar')->toBe('bar');
    })->depends('foo');
});
```

### Related:

https://github.com/pestphp/pest/issues/936

### DIFF
--- a/overrides/Runner/TestSuiteLoader.php
+++ b/overrides/Runner/TestSuiteLoader.php
@@ -132,7 +132,7 @@ final class TestSuiteLoader
                     continue;
                 }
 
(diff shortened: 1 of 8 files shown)


devajmeireles:
Some changes here were made by Pint formation.

PR #952, opened 2023-09-07
[2.x] Add `toBeAnonymous` expectation

hungthai1401:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

This PR adds `toBeAnonymous` expectation.
Usage:

```php
$anonymousClass = new class()
{
};

expect($anonymousClass)->toBeAnonymous();
```


src/Mixins/Expectation.php
```diff
@@ -1128,4 +1129,19 @@ public function toBeUuid(string $message = ''): self
...
     }
+
+    /**
+     * Asserts that the value is an anonymous class.
+     *
+     * @return self<TValue>
+     */
+    public function toBeAnonymous(string $message = ''): self
+    {
+        $this->toBeObject();
```
owenvoke, line 1170: Do we only want to check classes? Functions can be anonymous too, so it might be useful for some people who work with anonymous callables, etc. to have support for `ReflectionFunction::isAnonymous()` too.

hungthai1401: @owenvoke We already have `toBeCallable` expection to check anonymous function. So I think no need

owenvoke: `toBeCallable` isn't necessarily anonymous though. 🤔

hungthai1401: @owenvoke `ReflectionFunction::isAnonymous()` only works from php 8.2. Should we add it?

owenvoke: Ah, probably not then. 👍🏻

hungthai1401: @owenvoke thanks

nunomaduro, line 1170: Is this true? Can't I provide a string that is an anonymous class?

nunomaduro (changes requested):

src/Mixins/Expectation.php
hungthai1401: @nunomaduro No, I think so

nunomaduro: Try to the get the class from the anonymous class like so `get_class($anonymousObject);` and use that string to check.

hungthai1401: @nunomaduro Sorry, I don't understand your mind

owenvoke: I think what Nuno is meaning is that you can pass an anonymous class reference, for example:
```php
<?php

$anonymousClass = new class() {};

$class = get_class($anonymousClass); // class@anonymous/tmp/preview:3$0

var_dump((new ReflectionClass($class))->isAnonymous()); // true
```

hungthai1401: but I don't understand what advantage of `get_class` at here

nunomaduro:
Not sure what happen here, but the contents of this pull request are wrong - and contain other stuff.

Issue #946, opened 2023-08-31
[Bug]: `->toMatchSnapshot()` will not work with `@csrf`


alexmanase:
### What Happened

When I use `expect($response)->toMatchSnapshot()` and I use the `@csrf` directory the test will fail because the csrf is changed on every request.`

I think there should be a way to ignore the `@csrf` directive on snapshot testing.

### How to Reproduce

1. Create a Laravel Project with Pest
2. Create a view that contains `@csrf`
3. Use the Pest API for testing the snapshot: `expect($response)->toMatchSnapshot()`.
4. Run tests.

### Sample Repository

https://github.com/alexmanase/pest-snapshost-csrf-issue

### Pest Version

2.16.1

### PHP Version

8.2.8

### Operation System

macOS

nunomaduro:
cc @freekmurze

freekmurze:
In our projects we add a function to strip out the token of a value before snapshotting that value.

It would be nice if there were some helper methods on the Pest snapshot methods for this. Or maybe let the `toMatchSnapshot()` method accept a canonicalising callable.

```
expect($response)->toMatchSnapshot($callableThatWillCanonicaliseResponse)
```

gehrisandro:
There is a similar problem with `wire:id="<random-string>"` when trying to take a snapshot of a Livewire component.

I think it would be nice if the most common "problems" would be handled automatically by Pest. But the idea with the callable sounds nice for custom use cases. 👍

owenvoke:
I think the callable makes the most sense. Perhaps for Livewire, it could be a hook or something via the Pest Livewire plugin, and `@csrf` output could be part of the Pest Laravel plugin. I think it seems a bit weird maybe to handle a bunch of Laravel-specific cases in Pest core. As people might ask us to add (and maintain) a bunch of other common cases, such as for Symfony and other frameworks. 🤷🏻

For example, we could have a method like `Snapshot::macro('laravel.csrf', function () { ... });`, and then in plugins they can register bits to strip out in snapshots. 🤷🏻 And then have like `Snapshot::disableMacro('laravel.csrf');` if you don't want to include a plugins macro.

gehrisandro:
@owenvoke Sounds amazing 🎉

nunomaduro:
@gehrisandro I like the idea of `nice if the most common "problems" would be handled automatically by Pest`.

gehrisandro:
Here my current quick and dirty fix for Livewire:

```php
[...]

PR #954, opened 2023-09-07
[2.x] Add macros and interceptors to snapshot testing

gehrisandro:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

The primary purpose of this PR is to ensure snapshot testing works with CSRF tokens and other random data.

As discussed [here](https://github.com/pestphp/pest/issues/946) the goal is handle common types of random data (CSRF token, Livewire wire:id, etc.) within Pest respectively within the Plugin it belongs to.
Besides that the new API is public allowing users to add their own macros and interceptors.

There are two related plugin PRs:
- Laravel Plugin: https://github.com/pestphp/pest-plugin-laravel/pull/48
- Livewire Plugin: https://github.com/pestphp/pest-plugin-livewire/pull/18

#### Intercept
This allows to intercept the `toMatchSnapshot` method for custom classes. This wouldn't be necessary to achieve the main goal but it makes the handling easier in the Plugins.

Example from the Laravel plugin `PestServiceProvider`. See this [PR](https://github.com/pestphp/pest-plugin-laravel/pull/48).
```php
Snapshot::intercept(TestResponse::class, function (TestResponse $response): string {
    return $response->getContent();
});
```

Example for a custom interceptor:
```php
test('create custom interceptor', function () {
    Snapshot::intercept(Color::class, function (Color $color): string {
        return $color->getStyle();
    });

    expect(new Color(255, 0, 0))
        ->toMatchSnapshot();
});
```

Nice side effect here, we can get rid of this unrelated `TestResponse` within the Pest core:
```php
$this->value instanceof \Illuminate\Testing\TestResponse => $this->value->getContent(), // @phpstan-ignore-line
```

#### Macros
This allows to add macros to modify the snapshot string before the match is performed.

From the Livewire plugin `PestServiceProvider`. See this [PR](https://github.com/pestphp/pest-plugin-livewire/pull/18).
```php
Snapshot::macro('livewire.wire-key', function (string $value) {
    return preg_replace('/wire:id="[0-9a-zA-Z]*"/', 'wire:id="..."', $value);
});
```

An example for a custom macro:
```php
Snapshot::macro('time', function (string $value): string {
    return preg_replace('/[0-9]{2}:[0-9]{2}:[0-9]{2}/', '...', $value);
});

$value = 'The current time is '.now()->toTimeString();

expect($value)
    ->toMatchSnapshot();
```

Users can disable a Macro if they don't want to use it:
```php
Snapshot::disableMacro('laravel.csrf');
```

#### Test repository
For an easy overview of all the new possibilities I've setup a test [repository](https://github.com/gehrisandro/test-pest-snapshot/tree/main/tests/Feature).

#### Documentation
For sure we have to update the documentation.
I am able to help there if you are happy with the decisions I made.

#### Draft
_This and all related Plugin PRs are drafts. I have to clean them up and add more tests when you agree to the API suggested here._

### Related:

https://github.com/pestphp/pest/issues/946


owenvoke (approved):
Personally, I love this and think it's really neat. 👍🏻 Will see what the others think though. 👀

nunomaduro:
Fixed here: https://github.com/pestphp/docs/commit/a0b413c22033d8af9c3cb0f4e0e9c0767244b53e.

PR #962, opened 2023-09-18
add toHaveConstructor() and toHaveDestructor() expectations

JonPurvis:
### What:

- [x] New Feature

### Description:

Adds a more fluent way to be able to check that classes have a constructor and/or destructor. Currently, you'd do something like:

```php
test('classes have a constructor and destructor')
    ->expect('App\Http\Example')
    ->toHaveMethod('__construct');
    ->toHaveMethod('__destruct');
```

with this PR, you can instead do:

```php
test('classes have a constructor and destructor')
    ->expect('App\Http\Example')
    ->toHaveConstructor()
    ->toHaveDestructor();
```

If you need to check to see if classes _don't_ have a constructor and/or destructor, then you can do:

```php
test('classes do not have a constructor and destructor')
    ->expect('App\Http\Example')
    ->not->toHaveConstructor()
    ->not->toHaveDestructor();
```

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -862,4 +862,20 @@ final class Expectation
             FileLineFinder::where(fn (string $line): bool => str_contains($line, 'class')),
         );
     }
+
+    /**
+     * Asserts that the given expectation target has a constructor method.
+     */
+    public function toHaveConstructor(): ArchExpectation
+    {
+        return $this->toHaveMethod('__construct');
+    }
+
+    /**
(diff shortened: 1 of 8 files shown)


JonPurvis:
@nunomaduro Thanks for merging, I'll submit a PR to update the Expectation docs later today if that's okay with you?

nunomaduro:
Yeah - we need docs on this.

JonPurvis:
@nunomaduro docs PR submitted 👍

PR #972, opened 2023-09-29
[2.x] Fixing Backtrace not found error if project dirname endswith pest

Carnicero90:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Currently, \Pest\Support\Backtrace::backtrace() is skipping every file which `str_contains($trace['file'], 'pest'.DIRECTORY_SEPARATOR.'src'`, so that for whatever reason the base folder of the project's name ends with pest (i.e.: _testpest_, which was actually my folder name), it fall back to `throw ShouldNotHappen::fromMessage('Backtrace not found.');`. I guess `DIRECTORY_SEPARATOR.'pestphp'.DIRECTORY_SEPARATOR.'pest'.DIRECTORY_SEPARATOR.'src'` should be safe enough to avoid that issue.

### DIFF
--- a/src/Support/Backtrace.php
+++ b/src/Support/Backtrace.php
@@ -115,7 +115,7 @@ final class Backtrace
                 continue;
             }
 
-            if (str_contains($trace['file'], 'pest'.DIRECTORY_SEPARATOR.'src')) {
+            if (str_contains($trace['file'], DIRECTORY_SEPARATOR.'pestphp'.DIRECTORY_SEPARATOR.'pest'.DIRECTORY_SEPARATOR.'src')) {
                 continue;
             }
 


devajmeireles:
Hey, @Carnicero90 ! Thanks for your PR.

Could you fix the errors here, please? Thank you.

Carnicero90:
Yeah I'll look into that, but I'm not sure it is related with my commit, since I get the very same error even if I run `make test` or `composer test:unit` in a fresh clone of the original repo (on branch 2.x)

Carnicero90:
@devajmeireles are you positive that those issue are caused by something in my PR? Cos I am getting the very same error (a `DatasetMissing` exception whenever you are trying to set the value for the test in a **describe** context via `beforeEach()->with(values);`) even without any change at all in the source code. Still, I could try to fix that, but it looks like a completely different issue, which deserves a dedicated issue/pr.  Dunno if I am missing something trivial, my apologies if that's the case.

PR #974, opened 2023-10-03
[2.x] Add `toContainEquals` expectation

erikgaal:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

Adds an expectation `toContainEquals` which is analogous to `Assert::assertContainsEquals` from PHPUnit.

### DIFF
--- a/tests/Features/Expect/toContainEquals.php
+++ b/tests/Features/Expect/toContainEquals.php
@@ -1,0 +1,36 @@
+<?php
+
+use PHPUnit\Framework\ExpectationFailedException;
+
+
+test('passes arrays', function () {
+    expect([1, 2, 42])->toContain('42');
+});
+
+test('passes arrays with multiple needles', function () {
+    expect([1, 2, 42])->toContain('42', '2');
+});
+
+test('failures', function () {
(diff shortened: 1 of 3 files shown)


tests/Features/Expect/toContainEquals.php
```diff
@@ -0,0 +1,36 @@
...
+    expect([1, 2, 42])->toContain('42');
+});
+
+test('passes arrays with multiple needles', function () {
+    expect([1, 2, 42])->toContain('42', '2');
+});
+
+test('failures', function () {
+    expect([1, 2, 42])->toContain('3');
+})->throws(ExpectationFailedException::class);
```
devajmeireles: I believe your goal with this was to prove that `toContain` does not analyze type and value, right? I don't know if we should keep these tests here because they are not part of the `toContainEquals` concept.

@nunomaduro , thoughts?

erikgaal: Actually these should use `toContainEquals`, good catch.

devajmeireles: > Actually these should use `toContainEquals`, good catch.

If this is wrong, please update the PR.

Issue #908, opened 2023-08-14
`repeat` method needs some improvements.


utsavsomaiya:
![image](https://github.com/pestphp/pest/assets/96036522/13874a9c-4f1a-47d3-acf4-9e722a852ffb)

devajmeireles:
> ![image](https://user-images.githubusercontent.com/96036522/260444612-13874a9c-4f1a-47d3-acf4-9e722a852ffb.png)

Would be a very good addition, @nunomaduro. I think I'm able to do that 😣 As a workaround, you can do that, @utsavsomaiya :

![CleanShot 2023-08-15 at 6 46 58](https://github.com/pestphp/pest/assets/60591772/793f9280-4008-4340-ac38-0a3193170da8)

nunomaduro:
@devajmeireles Can you do a pull request? Thanks!

utsavsomaiya:
I will try from my side.. If I am not getting into it. Then can you please help me?

Katalam:
@utsavsomaiya I created a PR for it, I wasn't sure how much work you already put into it, but I thought I try it anyway

utsavsomaiya:
> @utsavsomaiya I created a PR for it, I wasn't sure how much work you already put into it, but I thought I try it anyway

Okay.. you can do it.

nunomaduro:
Fixed in Pest `v3.x`.

PR #975, opened 2023-10-05
[3.x] Sharing `repeat` iteration as `dataset` variable

Katalam:
### What:

- [ ] Bug Fix
- [ ] New Feature
- [x] Improvement

### Description:

- Essentially, what was previously in [Issue #908](https://github.com/pestphp/pest/issues/908) requested
- When no dataset is provided, but the repetitions are above 1, the current iteration should be passed to the closure of the test

### Related:
[Issue #908](https://github.com/pestphp/pest/issues/908)

### DIFF
--- a/tests/Features/Repeat.php
+++ b/tests/Features/Repeat.php
@@ -17,6 +17,29 @@ test('multiple times with multiple dataset', function (int $numberA, int $number
         ->and([4, 5, 6])->toContain($numberB);
 })->repeat(times: 7)->with(['a' => 1, 'b' => 2, 'c' => 3], [4, 5, 6]);
 
-test('multiple times with iterator as argument', function (int $iteration) {
-    expect($iteration)->toBeGreaterThan(0);
-})->repeat(times: 8);
+test('multiple times with iterator', function (int $iteration) {
+    expect($iteration)
+        ->toBeNumeric()
+        ->toBeGreaterThan(0);
+})->repeat(times: 2);
+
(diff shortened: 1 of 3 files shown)


devajmeireles (approved):

devajmeireles:
Hey, @Katalam ! Thanks for your PR.

Everything works as expected. Approved. 👍🏻

tests/Features/Repeat.php
```diff
@@ -16,3 +16,7 @@
     expect([1, 2, 3])->toContain($numberA)
         ->and([4, 5, 6])->toContain($numberB);
 })->repeat(times: 7)->with(['a' => 1, 'b' => 2, 'c' => 3], [4, 5, 6]);
+
+test('multiple times with iterator as argument', function (int $iteration) {
+    expect($iteration)->toBeGreaterThan(0);
+})->repeat(times: 8);
```
utsavsomaiya: Can you please write a test with the dataset so we can get both as closure arguments!!

Katalam:
@devajmeireles hey, should the conflict between repeat and dataset at the same time be solved by passing both iteration variables in the closure? I can do that

tests/Features/Repeat.php
devajmeireles: > Can you please write a test with the dataset so we can get both as closure arguments!!

@Katalam , could you write a test that uses the `repeat` and `with` to validate if all will work as expected, please?

Katalam: @devajmeireles Done

devajmeireles (approved):

PR #977, opened 2023-10-08
[2.x] Adds `toBeUrl()` Expectation

JonPurvis:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

This PR adds a `toBeUrl()` expectation, to validate whether a value is a valid URL. This uses the same `isUrl()` method that Laravel has.

Example Usage:
```php
expect('https://pestphp.com')->toBeUrl();
expect('test')->not->toBeUrl();
```

### DIFF
--- a/src/Support/Str.php
+++ b/src/Support/Str.php
@@ -108,4 +108,12 @@ final class Str
     {
         return sprintf('`%s` → %s', $describeDescription, $testDescription);
     }
+
+    /**
+     * Determine if a given value is a valid URL.
+     */
+    public static function isUrl(string $value): bool
+    {
+        return (bool) filter_var($value, FILTER_VALIDATE_URL);
+    }
 }
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -1142,4 +1142,20 @@ final class Expectation
 
         return $this;
     }
+
+    /**
+     * Asserts that the value is a url
+     *
+     * @return self<TValue>
+     */
+    public function toBeUrl(string $message = ''): self
+    {
+        if ($message === '') {
+            $message = "Failed asserting that {$this->value} is a url.";
+        }
+
+        Assert::assertTrue(Str::isUrl((string) $this->value), $message);
+
+        return $this;
+    }
 }
--- a/tests/Features/Expect/toBeUrl.php
+++ b/tests/Features/Expect/toBeUrl.php
@@ -1,0 +1,24 @@
+<?php
+
+use PHPUnit\Framework\ExpectationFailedException;
+
+test('pass', function () {
+    expect('https://pestphp.com')->toBeUrl()
+        ->and('pestphp.com')->not->toBeUrl();
+});
+
+test('failures', function () {
+    expect('pestphp.com')->toBeUrl();
+})->throws(ExpectationFailedException::class);
+
+test('failures with custom message', function () {
+    expect('pestphp.com')->toBeUrl('oh no!');
+})->throws(ExpectationFailedException::class, 'oh no!');
+
+test('failures with default message', function () {
+    expect('pestphp.com')->toBeUrl();
+})->throws(ExpectationFailedException::class, 'Failed asserting that pestphp.com is a url.');
+
+test('not failures', function () {
+    expect('https://pestphp.com')->not->toBeUrl();
+})->throws(ExpectationFailedException::class);


src/Support/Str.php
```diff
@@ -108,4 +129,22 @@ public static function describe(string $describeDescription, string $testDescrip
     {
         return sprintf('`%s` → %s', $describeDescription, $testDescription);
     }
+
+    /**
+     * Determine if a given value is a valid URL.
+     *
+     * @param  array<string>  $protocols
+     */
+    public static function isUrl(mixed $value, array $protocols = []): bool
```
nunomaduro: Can you remove `protocols` argument?

src/Support/Str.php
```diff
@@ -108,4 +129,22 @@ public static function describe(string $describeDescription, string $testDescrip
...
+    {
+        if (! is_string($value)) {
```
nunomaduro: Please accept only a `string` on the parameter type.

src/Support/Str.php
```diff
@@ -108,4 +129,22 @@ public static function describe(string $describeDescription, string $testDescrip
...
+            return false;
+        }
+
+        $protocolList = $protocols === []
```
nunomaduro: Can we don't have a fixed list of domains?

JonPurvis:
Documentation PR: https://github.com/pestphp/docs/pull/237

owenvoke (approved):

nunomaduro (changes requested):

src/Support/Str.php
JonPurvis: Done!

src/Support/Str.php
JonPurvis: Done!

src/Support/Str.php
JonPurvis: I've made what I _think_ the change should be for this one

JonPurvis:
Hey @nunomaduro 

Thanks for the feedback, I think I've actioned all of your points 😄

src/Support/Str.php
```diff
@@ -20,6 +20,32 @@ final class Str
...
...ms-settings-displays-topology|ms-settings-emailandaccounts|ms-settings-language|ms-settings-location|ms-settings-lock|ms-settings-nfctransactions|ms-settings-notifications|ms-settings-power|ms-settings-privacy|ms-settings-proximity|ms-settings-screenrotation|ms-settings-wifi|ms-settings-workplace|ms-spd|ms-sttoverlay|ms-transit-to|ms-useractivityset|ms-virtualtouchpad|ms-visio|ms-walk-to|ms-whiteboard|ms-whiteboard-cmd|ms-word|msnim|msrp|msrps|mss|mtqp|mumble|mupdate|mvn|news|nfs|ni|nih|nntp|notes|ocf|oid|onenote|onenote-cmd|opaquelocktoken|openpgp4fpr|pack|palm|paparazzi|payto|pkcs11|platform|pop|pres|prospero|proxy|pwid|psyc|pttp|qb|query|redis|rediss|reload|res|resource|rmi|rsync|rtmfp|rtmp|rtsp|rtsps|rtspu|s3|secondlife|service|session|sftp|sgn|shttp|sieve|simpleledger|sip|sips|skype|smb|sms|smtp|snews|snmp|soap\.beep|soap\.beeps|soldat|spiffe|spotify|ssh|steam|stun|stuns|submit|svn|tag|teamspeak|tel|teliaeid|telnet|tftp|tg|things|thismessage|tip|tn3270|tool|ts3server|turn|turns|tv|udp|unreal|urn|ut2004|v-event|vemmi|ventrilo|videotex|vnc|view-source|wais|webcal|wpid|ws|wss|wtai|wyciwyg|xcon|xcon-userid|xfire|xmlrpc\.beep|xmlrpc\.beeps|xmpp|xri|ymsgr|z39\.50|z39\.50r|z39\.50s';
```
nunomaduro: Again, I don't want to maintain a list of all possible URLs in the world. Is not possible to use something like FILTER_VALIDATE_URL instead?

nunomaduro (changes requested):

src/Support/Str.php
JonPurvis: Ah yes, I'm with you now. Yeah, we can do that. I've just pushed an update that remove this and uses `FILTER_VALIDATE_URL`, same results, lets code to maintain 😄

JonPurvis:
Thanks @nunomaduro 🥳

PR #986, opened 2023-10-19
Fix toHaveProperty with dynamics properties (magically accessible)

### LINKED ISSUE #987: [Bug]: Magically set properties are currently ignored by toHaveProperty expectation
### What Happened

Magically set properties are currently ignored

### How to Reproduce

```php
test('dynamic properties', function () {
    $object = new class {
        public function __isset(string $name): bool {
            if ($name === 'foo') {
                return true;
            } else {
                return false;
            }
        }
    };

    expect($object)->toHaveProperty('foo');
    expect($object)->not->toHaveProperty('bar');
});
```

### Sample Repository

_No response_

### Pest Version

2.24.0

### PHP Version

8.2

### Operation System

Linux

### Notes

_No response_

julien-boudry:
### What:

- [x] Bug Fix

### Description:

Fix toHaveProperty() with dynamics properties that are magically accessible using the [__get](https://www.php.net/manual/en/language.oop5.overloading.php#language.oop5.overloading.members) magic method.

### Related:

Fix #987

### DIFF
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -301,7 +301,7 @@ final class Expectation
         $this->toBeObject();
 
         // @phpstan-ignore-next-line
-        Assert::assertTrue(property_exists($this->value, $name), $message);
+        Assert::assertTrue(property_exists($this->value, $name) || isset($this->value->$name), $message);
 
         if (! $value instanceof Any) {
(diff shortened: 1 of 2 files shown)


nunomaduro:
It's a breaking change, and I am not sure I want to change this behavior of this expectation.

PR #996, opened 2023-10-30
Create PHPStan extension and add `HigherOrderTapProxy` to `universalObjectCratesClasses`

calebdw:
### What:

- [x] New Feature

### Description:
Hello!

With the release of PHP 8.2, dynamic property creation is deprecated unless the class uses the `AllowDynamicProperties` attribute or defines the `__get`/`__set` magic methods. However, PHPStan still reports the dynamic properties as being undefined unless the class is added to the `universalObjectCratesClasses` config.

Because of this, PHPStan is throwing errors for the below example even though the `HigherOrderTapProxy` (returned by empty `test()` call) does implement the `__get`/`__set` magic methods:

```php
beforeEach(function () {
    test()->dateFormat = 'Y-m-d';
});

it('...', function () {
    // ...
    now()->format(test()->dateFormat);
});

// yields
Access to an undefined property Pest\PendingCalls\TestCall|Pest\Support\HigherOrderTapProxy::$dateFormat.
```

This PR creates an auto-registerable PHPStan extension that adds the `HigherOrderTapProxy` to the `universalObjectCratesClasses` config to fix these errors.

Thanks!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -109,6 +109,11 @@
                 "Pest\\Plugins\\Version",
                 "Pest\\Plugins\\Parallel"
             ]
+        },
+        "phpstan": {
+            "includes": [
+                "extension.neon"
+            ]
         }
     }
 }
--- a/extension.neon
+++ b/extension.neon
@@ -1,0 +1,3 @@
+parameters:
+    universalObjectCratesClasses:
+        - Pest\Support\HigherOrderTapProxy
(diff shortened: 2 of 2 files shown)


calebdw:
Note, I also added `Expectation` to `universalObjectCratesClasses` to fix the below error type:

```php
expect($log)->finished_at->toBeNull();

Access to an undefined property Pest\Expectation<mixed>::$finished_at.
```

PR #998, opened 2023-11-02
[2.x] Improve "not" modifier in order to accept condition

clementbirkle:
### What:

New Feature

### Description:

With this PR, this test:
```
it('example (before)', function (bool $hasMedia, int $mediaId) {

    $media = Media::find($mediaId);

    if ($hasMedia) {
        expect($media)->not->toBeNull();
    } else {
        expect($media)->toBeNull();
    }
})->with([
    [true, 1],
    [false, 2],
]);
```

Could be refactorized like that:
```
it('example (after)', function (bool $hasMedia, int $mediaId) {

    $media = Media::find($mediaId);

    expect($media)->not($hasMedia)->toBeNull();
})->with([
    [true, 1],
    [false, 2],
]);
```

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -180,10 +180,14 @@ final class Expectation
     /**
      * Creates the opposite expectation for the value.
      *
-     * @return OppositeExpectation<TValue>
+     * @return OppositeExpectation<TValue>|self<TValue>
      */
-    public function not(): OppositeExpectation
+    public function not(bool $condition = true): OppositeExpectation|self
     {
+        if (! $condition) {
+            return $this;
+        }
+
         return new OppositeExpectation($this);
     }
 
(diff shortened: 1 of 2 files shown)


devajmeireles:
Hey, @clementbirkle. Thanks for your PR!

Is [when](https://pestphp.com/docs/expectations#content-when) not enough to cover cases like the example you gave?

clementbirkle:
Hi @devajmeireles !

Of course we can do it with "when" but it's a bit cumbersome:

```
it('example (after)', function (bool $hasMedia, int $mediaId) {

    $media = Media::find($mediaId);

    expect($media)
        ->when($hasMedia, fn ($value) => $value->not->toBeNull)
        ->when(! $hasMedia, fn ($value) => $value->toBeNull);
})->with([
    [true, 1],
    [false, 2],
]);
```

nunomaduro:
I prefer to use when for these cases,

Issue #738, opened 2023-03-24
Impossible to combine flags with `--watch`


meduzen:
When running `php artisan test --watch --compact`, I expect to have a compact output in the terminal.

However, `--compact` is ignored because it is remove from the arguments array that is passed down to the plugin:
https://github.com/pestphp/pest/blob/6c8970e0a3b9bb36544bb1eacba0a4175dbafe97/bin/pest#L21-L39

I’m not familiar enough yet with the whole Pest ecosystem, so I don’t know if the fix would be as simple as “don’t `unset()` arguments in that `foreach` loop” nor if it would make sense to fix it that way for all flags or not.

ermesonqueiroz:
I investigated the issue, and here are my thoughts:

The `unset()` in that `foreach` loop is beacuse the PHPUnit doesn't have the same options as pest.
![image](https://github.com/pestphp/pest/assets/111156073/0bc74765-01cb-4bcd-85de-a90df1b77feb)

The `--watch` option call the command that you executed when a test is updated or a new is created. However, with the current use of `unset()`, the watch plugin isn't receiving all options.

I propose removing these options after passing them to the plugins. I'm already working on it, but I would appreciate your feedback and any suggestions you might have.

nunomaduro:
Fixed in `v2.33.0` that will be released today.

PR #1000, opened 2023-11-06
[2.x] fix: allow flags combination with --watch

ermesonqueiroz:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

This PR introduce the ability to use the `--watch` flag with other pest flags

### Related:

#738


nunomaduro:
This will be fixed in `v2.33.0`

PR #1001, opened 2023-11-07
Fix html in descriptions or datasets

faissaloux:
### What:

- [ x ] Bug Fix
- [ - ] New Feature

### Description:
Termwind raises a StyleNotFound exception when a test dataset or description includes html tags with class attribute, as Termwind tries to style it as html element but the class is not supported.

Example:
Screenshot from 2023-11-07 17-36-58
photo_2023-11-07_17-52-38

If the test desciption doesn't include html tags with class attribute, Termwind will raise the same exception for dataset.

I have converted html tags into HTML entities so that Termwind will treat it as normal text. And the result will shows as written in the test description or Dataset:

Screenshot from 2023-11-07 17-50-31

PS: Same thing applies on `describe()`.


tests/Unit/TestName.php
```diff
@@ -20,7 +20,7 @@
     '卜竹弓一十山' => '__pest_evaluable_卜竹弓一十山',
     'アゴデヸ' => '__pest_evaluable_アゴデヸ',
     '!p8VrB' => '__pest_evaluable__p8VrB',
-    '&xe6VeKWF#n4' => '__pest_evaluable__xe6VeKWF_n4',
+    '&amp;xe6VeKWF#n4' => '__pest_evaluable__amp_xe6VeKWF_n4',
```
nunomaduro, line 23: Why you had to change this?

tests/Unit/TestName.php
```diff
@@ -33,7 +33,7 @@
     'Каролин' => '__pest_evaluable_Каролин',
     'অ্যান্টার্কটিকা' => '__pest_evaluable_অ্যান্টার্কটিকা',
     'Frýdek-Místek"' => '__pest_evaluable_Frýdek_Místek_',
-    'Allingåbro&' => '__pest_evaluable_Allingåbro_',
+    'Allingåbro&amp;' => '__pest_evaluable_Allingåbro_amp_',
```
nunomaduro, line 36: Why you had to change this?

nunomaduro (changes requested):

tests/Unit/TestName.php
faissaloux: It's because `$__latestDescription` changes here after changing the description HTML tags to HTML entities.

https://github.com/pestphp/pest/blob/0a680dd06e02bdc58132d6bb3f155498355160a5/src/Concerns/Testable.php#L212

PR #1003, opened 2023-11-09
[2.x] Add contain with count methods

faissaloux:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:
In this PR I have added contain with count methods `toContainAtLeast` , `toContainAtMost` and `toContainExact`  which asserts that a substring exists in a string for a number of times (exact or at least) same thing for iterables it asserts that it contains an element exatly or at least a number of times.

PS: We can use `toContainOp` too which can be controlled by the operator passed. (`greater than or equal`, `greater than`, `less than or equal`, `less than` ...).

### Examples:
```php
it('to have count', function() {
    expect([[1, 2, 3], 2, 42])->toContain(42, [1, 2, 3]);
    expect([[1, 2, 3], 2, 42, 42])->toContainAtLeast(1, 42);
    expect([1, 42, 2, 42, 31, 42])->toContainAtMost(3, 42);
    expect([[1, 2, 3], 2, 42])->toContainExact(1, [1, 2, 3]);
    expect("pest is the best.")->toContain('pest');
    expect(["pest", "is", "the", "best", "pest"])->toContain('pest');
    expect("pest is the best")->toContainAtLeast(1, 'pest');
    expect(["pest", "is", "the", "best", "pest"])->toContainExact(2, 'pest');
})->only();
```

### DIFF
--- a/src/Exceptions/InvalidMethod.php
+++ b/src/Exceptions/InvalidMethod.php
@@ -1,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pest\Exceptions;
+
+use Exception;
+
+/**
+ * @internal
+ */
+final class InvalidMethod extends Exception
+{
+    /**
+     * Creates a new InvalidMethod instance from the given name.
+     */
+    public static function fromName(string $name): InvalidMethod
+    {
+        return new self("Method [$name] does not exist.");
+    }
+}
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -208,6 +208,7 @@ final class Expectation
         if (is_string($this->value)) {
             // @phpstan-ignore-next-line
             Assert::assertStringContainsString((string) $needle, $this->value);
+            // @phpstan-ignore-next-line
             Assert::$method($count, substr_count($this->value, $needle), $message);
         } else {
             if (! is_iterable($this->value)) {
@@ -220,6 +221,7 @@ final class Expectation
                     $found++;
                 }
             }
+            // @phpstan-ignore-next-line
             Assert::$method($count, $found, $message);
         }
 
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -214,13 +214,16 @@ final class Expectation
             if (! is_iterable($this->value)) {
                 InvalidExpectationValue::expected('iterable');
             }
+
             Assert::assertContains($needle, $this->value);
+
             $found = 0;
             foreach ($this->value as $val) {
                 if ($val === $needle) {
                     $found++;
                 }
             }
+
             // @phpstan-ignore-next-line
             Assert::$method($count, $found, $message);
         }
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -209,7 +209,7 @@ final class Expectation
             // @phpstan-ignore-next-line
             Assert::assertStringContainsString((string) $needle, $this->value);
             // @phpstan-ignore-next-line
-            Assert::$method($count, substr_count($this->value, $needle), $message);
+            Assert::$method($count, substr_count($this->value, (string) $needle), $message);
         } else {
             if (! is_iterable($this->value)) {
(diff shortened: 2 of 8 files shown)


src/Mixins/Expectation.php
```diff
@@ -182,20 +183,87 @@ public function toBeLessThanOrEqual(int|float|DateTimeInterface $expected, strin
...
+    /**
+     * Asserts that $needle exists in the value depending on $op $count times.
+     *
+     * @return self<TValue>
+     */
+    public function toContainOp(string $op, int $count, mixed $needle, string $message = ''): self
+    {
+        Assert::assertGreaterThanOrEqual(0, $count);
+
+        $op = str_replace(' ', '', ucwords($op));
```
devajmeireles, line 201: I'm not a big fan of this kind of approach. IMO, this is something that hurts readability. Why not pass the method name without the need of this formatting?

devajmeireles, line 201: I'm not a big fan of this kind of approach. IMO, this is something that hurts readability. Why not pass the method name without the need of this formatting?

src/Mixins/Expectation.php
```diff
@@ -182,20 +183,87 @@ public function toBeLessThanOrEqual(int|float|DateTimeInterface $expected, strin
...
+            // @phpstan-ignore-next-line
+            Assert::assertStringContainsString((string) $needle, $this->value);
+            // @phpstan-ignore-next-line
+            Assert::$method($count, substr_count($this->value, (string) $needle), $message);
+        } else {
+            if (! is_iterable($this->value)) {
+                InvalidExpectationValue::expected('iterable');
+            }
+            Assert::assertContains($needle, $this->value);
+            $found = 0;
```
devajmeireles:
```suggestion
            if (! is_iterable($this->value)) {
                InvalidExpectationValue::expected('iterable');
            }

            Assert::assertContains($needle, $this->value);

            $found = 0;
```

src/Mixins/Expectation.php
```diff
@@ -182,20 +183,87 @@ public function toBeLessThanOrEqual(int|float|DateTimeInterface $expected, strin
...
+            foreach ($this->value as $val) {
+                if ($val === $needle) {
+                    $found++;
                 }
-                Assert::assertContains($needle, $this->value);
             }
+            // @phpstan-ignore-next-line
+            Assert::$method($count, $found, $message);
```
devajmeireles:
```suggestion
            }

            // @phpstan-ignore-next-line
            Assert::$method($count, $found, $message);
```

src/Exceptions/InvalidMethod.php
```diff
@@ -0,0 +1,21 @@
...
+declare(strict_types=1);
+
+namespace Pest\Exceptions;
+
+use Exception;
+
+/**
+ * @internal
+ */
+final class InvalidMethod extends Exception
```
devajmeireles, line 12: This is only used in one place, and I don't think we'll ever need it again, or I'm wrong? Shouldn't we be worried about the amount of new, in theory, unnecessary files?

devajmeireles (changes requested):
Although Nuno validation is final, I made some observations that I believe are necessary to observe.

src/Mixins/Expectation.php
faissaloux: Fixed!

src/Mixins/Expectation.php
faissaloux: Fixed!

src/Exceptions/InvalidMethod.php
faissaloux: I think it's better for readability and testing, isn't it?

src/Mixins/Expectation.php
faissaloux: I made it this way to support both, and to not worry about how `$op` parameter is written.

For example `lessThanOrEqual`   we can pass it exactly how it's defined or just write it as simple as `less than or equal` .

nunomaduro:
I don't see myself using this. However, thanks for trying to contribute!

PR #1004, opened 2023-11-11
[2.x] Implementation to the `at` method

Tiskiel:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

Implementation of the `at` method. This will allow you to target an index or key on an iterable and apply other expectation methods to the value. It provides better readability and works on nested arrays.

Example below:

```php
it('ensures it work with nested array', function () {
    $nestedArray = [
        [1, 2, 3],
        ['foo' => 'bar', 'john' => 'doe'],
    ];

    expect($nestedArray)
        ->at(0)->at(1)->toBe(2)
        ->and($nestedArray)
        ->at(1)->foo->toBe('bar');
});
```

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -66,6 +66,20 @@ final class Expectation
     }
 
     /**
+     * Allows you to target a value in a table
+     *
+     * @return self<TAndValue>
+     */
+    public function at(string|int $key): Expectation
+    {
+        if (! is_iterable($this->value)) {
+            InvalidExpectationValue::expected('iterable');
+        }
+
+        return new self($this->value[$key]);
+    }
+
+    /**
      * Creates a new expectation.
      *
      * @template TAndValue
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -76,6 +76,31 @@ final class Expectation
             InvalidExpectationValue::expected('iterable');
         }
 
+        if(gettype($key) === 'string') {
+            if(strpos($key, '.') !== false) {
+                $keys = explode('.', $key);
+                $value = $this->value;
+                foreach($keys as $nestedKey) {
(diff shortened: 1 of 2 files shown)


src/Expectation.php
```diff
@@ -65,6 +65,20 @@ public function __construct(
         // ..
     }
 
+    /**
+     * Allows you to target a value in a table
+     *
+     * @return self<TAndValue>
+     */
+    public function at(string|int $key): Expectation
```
owenvoke, line 73: If this is to be implemented, I think an option for dot-notation syntax would be much nicer.

```php
->at('0.1.2')
```

Personally, I'm not sure of the benefit of this, can higher order access not be used?

```php
->{1}->{2}
->name->firstName
```

Tiskiel: > If this is to be implemented, I think an option for dot-notation syntax would be much nicer.

That's a great idea, I hadn't thought of that. I'm looking at reworking my PR in that direction this week. 

> Personally, I'm not sure of the benefit of this, can higher order access not be used?


Yes, it's usable but the function I want to implement is designed to add readability too. On a structure that's not too complex you can use higher order access but I think that on a more complex structure the risk of getting lost in the reading is greater.

I'd also like to find a way of being able to chain my function without having to use `and()` to provide even more readability. But at the moment I haven't figured out how I could do that.

Tiskiel:
Hi there, 

I have implemented the dot notation as spoked above.

Here is an example :

```php
it('ensures it work with dot notation', function () {
    $nestedDictionary = [
        'foo' => [
            'bar' => [
                'john' => 'doe',
            ],
        ],
    ];

    expect($nestedDictionary)
        ->at('foo.bar.john')->toBe('doe');
});
```

nunomaduro:
No plans for this.

Issue #1011, opened 2023-11-17
[Improvement]: Add fluent methods onlyOnWindows, onlyOnLinux ...


nahime0:
If a test should run only on a platform, it will be easy to specify onlyOnPlatform than skipping all the others. 
Internally Pest could still call the skipOn on all other platforms.

nahime0:
Opened PR: https://github.com/pestphp/pest/pull/1012

nahime0:
The PR https://github.com/pestphp/pest/pull/1012 has been merged

PR #1012, opened 2023-11-17
[2.x] Added onlyOn* methods to run the test only on a specific OS

nahime0:
### What:

- [ ] Bug Fix
- [X] New Feature

### Description:

Sometimes a test should run only on one explicit OS. It's easier to use onlyOn{OS}, rather than skipping all other OSes.
This PR introduces:

- onlyOnWindows()
- onlyOnMac()
- onlyOnLinux()

### Related:

Related to the issue/improvement https://github.com/pestphp/pest/issues/1011

### DIFF
--- a/src/PendingCalls/TestCall.php
+++ b/src/PendingCalls/TestCall.php
@@ -224,6 +224,30 @@ final class TestCall
     }
 
     /**
+     * Runs the current test only if the given test is running on Windows.
+     */
+    public function onlyOnWindows(): self
+    {
+        return $this->skipOnLinux()->skipOnMac();
+    }
+
+    /**
+     * Runs the current test only if the given test is running on Mac.
+     */
+    public function onlyOnMac(): self
+    {
+        return $this->skipOnWindows()->skipOnLinux();
+    }
+
+    /**
+     * Run the current test only if the given test is running on Linux.
+     */
+    public function onlyOnLinux(): self
+    {
+        return $this->skipOnWindows()->skipOnMac();
+    }
+
+    /**
      * Skips the current test if the given test is running on the given operating systems.
      */
     private function skipOn(string $osFamily, string $message): self
--- a/src/PendingCalls/TestCall.php
+++ b/src/PendingCalls/TestCall.php
@@ -224,35 +224,45 @@ final class TestCall
     }
 
     /**
-     * Runs the current test only if the given test is running on Windows.
-     */
-    public function onlyOnWindows(): self
-    {
-        return $this->skipOnLinux()->skipOnMac();
-    }
-
-    /**
-     * Runs the current test only if the given test is running on Mac.
-     */
-    public function onlyOnMac(): self
-    {
-        return $this->skipOnWindows()->skipOnLinux();
-    }
-
-    /**
-     * Run the current test only if the given test is running on Linux.
-     */
-    public function onlyOnLinux(): self
-    {
-        return $this->skipOnWindows()->skipOnMac();
-    }
-
-    /**
      * Skips the current test if the given test is running on the given operating systems.
      */
     private function skipOn(string $osFamily, string $message): self
     {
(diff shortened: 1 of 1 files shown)


src/PendingCalls/TestCall.php
```diff
@@ -223,6 +223,30 @@ public function skipOnLinux(): self
...
+        return $this->skipOnWindows()->skipOnLinux();
+    }
+
+    /**
+     * Run the current test only if the given test is running on Linux.
+     */
+    public function onlyOnLinux(): self
+    {
+        return $this->skipOnWindows()->skipOnMac();
+    }
```
owenvoke:
```suggestion
    /**
     * Skips the current test unless the given test is running on Windows.
     */
    public function onlyOnWindows(): self
    {
        return $this->onlyOn('Windows', 'This test is skipped unless on [Windows].');
    }

    /**
     * Skips the current test unless the given test is running on Mac.
     */
    public function onlyOnMac(): self
    {
        return $this->onlyOn('Darwin', 'This test is skipped unless on [Mac].');
    }

    /**
     * Skips the current test unless the given test is running on Linux.
     */
    public function onlyOnLinux(): self
    {
        return $this->onlyOn('Linux', 'This test is skipped unless on [Linux].');
    }
    
    /**
     * Skips the current test unless the given test is running on the given operating system.
     */
     public function onlyOn(string $osFamily, string $message): self
     {
         return $osFamily !== PHP_OS_FAMILY
             ? $this->skip($message)
             : $this;
     }
```

owenvoke (commented):
Personally I think this doesn't add much value. 🤔 Happy with this if the team want to maintain it though. 🤷🏻

I feel like it would make more sense to have an `onlyOn()` method that get's called though (as suggested).

src/PendingCalls/TestCall.php
nahime0: Thanks, I've updated the code based on your suggestions. The only difference is that I kept the 'onlyOn' method private.

src/PendingCalls/TestCall.php
```diff
@@ -233,6 +233,40 @@ private function skipOn(string $osFamily, string $message): self
             : $this;
     }
 
+    /**
+     * Skips the current test unless the given test is running on Windows.
+     */
+    public function onlyOnWindows(): self
+    {
+        return $this->onlyOn('Windows', 'This test is skipped unless on [Windows].');
```
nunomaduro: Can you call `skipOnLinux` and `skipOnMac` instead?

nunomaduro (changes requested):

src/PendingCalls/TestCall.php
nahime0: I think it's more fluent and easier to understand. 
If I have two alternative tests, one for a plaform and one for all the others, I can do onlyOnX on the first and skipOnX on the second. 
Moreover this will help in case other oses will be added. I'm using a bsd machine to launch  stress tests (using a patched stressless) and in that case i should skip 3 oses. If this get approved I'm planning to propose a PR also for skipOnBSD and onlyOnBSD.

owenvoke: 😅 Maybe I spoke to soon. 😂 Apologies if you have to reverse the changes I suggested, @nahime0

nahime0: Sorry, I didn't understood what requested by Nuno. I'll revert the changes 🙃

nahime0: @nunomaduro This [commit](https://github.com/pestphp/pest/pull/1012/commits/f7705fe1c1c007aa91aac9333f5039f2ae6c4908) solve that.

nunomaduro:
can you document this? thanks!

nahime0:
Sure @nunomaduro, https://github.com/pestphp/docs/pull/243

PR #1014, opened 2023-11-18
[2.x] Dockerfile Optimization

mjsafarali:
### What:

- [ ] Bug Fix
- [ ] New Feature
- [x] Improvement

### Description:

- Combined apk add commands: Instead of running two separate apk add commands, I combined them into a single command. This saves time and reduces the number of layers in the image.

- Combined docker-php-ext-configure, docker-php-ext-install, and docker-php-ext-enable commands: Similarly, I combined these commands for each extension into a single command. This further reduces the number of layers.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,21 +1,16 @@
 ARG PHP=8.1
 FROM php:${PHP}-cli-alpine
 
-RUN apk update \
-    && apk add zip libzip-dev icu-dev git
+RUN apk update && apk add \
+    zip libzip-dev icu-dev git \
 
-RUN docker-php-ext-configure zip
-RUN docker-php-ext-install zip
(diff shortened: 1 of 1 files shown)


devajmeireles (approved):

docker/Dockerfile
```diff
@@ -1,21 +1,16 @@
 ARG PHP=8.1
 FROM php:${PHP}-cli-alpine
 
-RUN apk update \
-    && apk add zip libzip-dev icu-dev git
+RUN apk update && apk add \
+    zip libzip-dev icu-dev git \
```
owenvoke, line 5: Won't this cause issues as it's a trailing newline?

genintho: It does. 

Screenshot 2023-12-27 at 7 48 08 PM

PR #1017, opened 2023-11-22
[2.x] Add `toSnapshot` early return

markhuot:
Sometimes objects need native toString() and toArray() methods that are different from what you want to snapshot.

This adds an explicit toSnapshot() method that will be called first (when set) allowing for better snapshot values than the generic methods offer.

### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

I have a number of objects in my project that already contain `toString()` and `toArray()` methods for different reasons than snapshotting. For example a user object that may downcast to a string by rendering the full name or to an array by returning the raw DB attributes.

But for snapshotting I want a mixture of both. I want the attributes with some computed fields and even some hidden fields. I'd like to include a user's "token" field in the snapshot, even though it's hidden from API responses so I can have confidence that the token logic isn't changing.

This adds an explicit `->toSnapshot()` method call above the more abstract methods allowing consumers to customize exactly what data is snapshotted.

### DIFF
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -844,6 +844,7 @@ final class Expectation
 
         $string = match (true) {
             is_string($this->value) => $this->value,
+            is_object($this->value) && method_exists($this->value, 'toSnapshot') => $this->value->toSnapshot(),
             is_object($this->value) && method_exists($this->value, '__toString') => $this->value->__toString(),
             is_object($this->value) && method_exists($this->value, 'toString') => $this->value->toString(),
             $this->value instanceof \Illuminate\Testing\TestResponse => $this->value->getContent(), // @phpstan-ignore-line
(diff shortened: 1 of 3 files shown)


src/Mixins/Expectation.php
```diff
@@ -844,6 +844,7 @@ public function toMatchSnapshot(string $message = ''): self
 
         $string = match (true) {
             is_string($this->value) => $this->value,
+            is_object($this->value) && method_exists($this->value, 'toSnapshot') => $this->value->toSnapshot(),
```
nunomaduro, line 847: Love it! Can you add a test on it?

nunomaduro (changes requested):

src/Mixins/Expectation.php
markhuot: Test added in https://github.com/pestphp/pest/pull/1017/commits/2a54b5819d5583698ba53e87d3d0bfa33fa1b2c8

markhuot:
Test added and this is ready for re-review. Test passes locally (as does the rest of the suite).

PR #1021, opened 2023-11-24
[2.x] Introduce Reminder When Filtering Using `only`

devajmeireles:
### What:

- [ ] Bug Fix
- [ ] New Feature
- [x] Enhancements

### Description:

This PR is intended to introduce a warning as a reminder when you are filtering tests using `->only()`, because without something like this, we can filter the tests but forget to remove the `->only()` filter.

CleanShot 2023-11-24 at 14 24 23

### DIFF
--- a/src/Plugins/Only.php
+++ b/src/Plugins/Only.php
@@ -34,7 +34,7 @@ final class Only implements Shutdownable
             unlink($lockFile);
         }
 
-        echo "The tests executed were tests filtered using <href=https://pestphp.com/docs/filtering-tests#content-only>only</>";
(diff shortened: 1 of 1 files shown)


src/Plugins/Only.php
```diff
@@ -33,6 +33,8 @@ public function shutdown(): void
         if (file_exists($lockFile)) {
             unlink($lockFile);
         }
+
+        echo 'The tests executed were tests filtered using <href=https://pestphp.com/docs/filtering-tests#content-only>only</>';
```
nunomaduro, line 37: Can you inject the output, like other plugins are doing, and use the info stuff?

nunomaduro (changes requested):

PR #1025, opened 2023-11-30
[2.x] Allow `--todo` argument.

xiCO2k:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

This allows `--todo` argument to be used as an alias for `--todos`.

I find myself running `pest --todo` all the time, and I figured that could be a good alias to have.

Cheers.

### DIFF
--- a/bin/pest
+++ b/bin/pest
@@ -38,7 +38,7 @@ use Symfony\Component\Console\Output\ConsoleOutput;
             unset($args[$key]);
         }
 
-        if ($value === '--todos') {
+        if (str_contains($value, '--todo')) {
             $todo = true;
             unset($args[$key]);
         }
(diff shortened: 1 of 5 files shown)


bin/pest
```diff
@@ -38,7 +38,7 @@ use Symfony\Component\Console\Output\ConsoleOutput;
             unset($args[$key]);
         }
 
-        if ($value === '--todos') {
+        if (str_contains($value, '--todo')) {
```
owenvoke: Perhaps use a strict check to be more specific.
```suggestion
        if (in_array($value, ['--todo', '--todos'], true)) {
```

Just in case someone wants to add a new flag like `--todo-only`

owenvoke (commented):
Looks good, I would also choose `--todo` over `--todos`. Just one suggestion.

bin/pest
xiCO2k: Yes, thanks @owenvoke

owenvoke (approved):
LGTM. Will leave for another review as well though.

PR #1038, opened 2023-12-19
PR for Error fixes and code improvement

saram-aman:
1. In `SnapshotRepository`:
   - Applied the `readonly` keyword correctly to class properties.

2. In `ChainableClosure`:
   - Addressed errors caused by using `$this` in static methods by introducing an explicit object instance (`$self`) within closures.

**Changes:**
- For `SnapshotRepository`:
  - Applied `readonly` to `$testsPath` and `$snapshotsPath` properties.

- For `ChainableClosure`:
  - Updated `boundWhen`, `bound`, and `boundStatically` methods to accept an object instance (`$self`) explicitly.
  - Replaced references to `$this` with `$self` within the closures.
  - Improved error handling for scenarios where the object is not provided.

### DIFF
--- a/src/Repositories/SnapshotRepository.php
+++ b/src/Repositories/SnapshotRepository.php
@@ -16,12 +16,22 @@ final class SnapshotRepository
     private static array $expectationsCounter = [];
 
     /**
+     * @readonly
+     */
+    private string $testsPath;
+
+    /**
+     * @readonly
+     */
+    private string $snapshotsPath;
+
+    /**
      * Creates a snapshot repository instance.
      */
-    public function __construct(
-        readonly private string $testsPath,
-        readonly private string $snapshotsPath,
-    ) {
+    public function __construct(string $testsPath, string $snapshotsPath)
+    {
+        $this->testsPath = $testsPath;
+        $this->snapshotsPath = $snapshotsPath;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Repositories/SnapshotRepository.php
```diff
@@ -15,13 +15,23 @@ final class SnapshotRepository
...
+    private string $snapshotsPath;
+
     /**
      * Creates a snapshot repository instance.
      */
-    public function __construct(
-        readonly private string $testsPath,
-        readonly private string $snapshotsPath,
-    ) {
+    public function __construct(string $testsPath, string $snapshotsPath)
```
owenvoke, line 31: Why would you change from PHP's built-in `readonly` keyword to a PHPDoc tag? 🤔 I don't think this change is necessary, or makes sense, when only PHP 8.1 and later are supported.

src/Support/ChainableClosure.php
```diff
@@ -15,31 +15,37 @@ final class ChainableClosure
     /**
      * Calls the given `$closure` when the given condition is true, "bound" to the same object.
      */
-    public static function boundWhen(Closure $condition, Closure $next): Closure
+    public static function boundWhen(object $self, Closure $condition, Closure $next): Closure
```
owenvoke, line 18: I'm not sure why `$this` is used in a static (assuming it's some magic), I think @nunomaduro might know, however these changes would be breaking changes if anyone is using these directly.

owenvoke (changes requested):

PR #1056, opened 2024-01-06
Skip on php versions

faissaloux:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

In this PR I have added the ability to skip tests on php versions.

```php
    it('skips specific php version')->skipOnPhp("8.3.0");
    it('skips multiple php versions depending on constraint')->skipOnPhp("<=7.4.0");
````

### DIFF
--- a/tests/Features/Skip.php
+++ b/tests/Features/Skip.php
@@ -60,5 +60,5 @@ it('can skip on specific php version')
     ->assertTrue(false);
 
 it('can skip on php versions depending on operator')
-    ->skipOnPhp('7.4.0', '>')
+    ->skipOnPhp('>=7.4.0')
     ->assertTrue(false);
--- a/tests/Features/Skip.php
+++ b/tests/Features/Skip.php
@@ -59,6 +59,6 @@ it('can skip on specific php version')
     ->skipOnPhp(PHP_VERSION)
     ->assertTrue(false);
 
-it('can skip on php versions depending on operator')
+it('can skip on php versions depending on constraint')
     ->skipOnPhp('>=7.4.0')
     ->assertTrue(false);
--- a/tests/Features/Skip.php
+++ b/tests/Features/Skip.php
@@ -54,3 +54,11 @@ it('can user higher order callables and skip')
         return $this->shouldSkip;
     })
     ->toBeFalse();
+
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: tests/Features/Skip.php [hunk 1]
```diff
@@ -54,3 +54,11 @@
         return $this->shouldSkip;
     })
     ->toBeFalse();
+
+it('can skip on specific php version')
+    ->skipOnPhp(PHP_VERSION)
+    ->assertTrue(false);
+
+it('can skip on php versions depending on operator')
+    ->skipOnPhp('7.4.0', '>')
```
nunomaduro: This API needs to be `->skipOnPhp('^7.4.0')` or `->skipOnPhp('>=7.4.0')` or  `->skipOnPhp('>7.4.0')`. Can you adjust that?

nunomaduro (changes requested):

tests/Features/Skip.php
owenvoke: That wouldn't work with `version_compare`, so we'd either have to pull in a package (e.g. [`composer/semver`](https://github.com/composer/semver)) or write our own parser.

I think `composer/semver` would work nicely with:

```php
Composer\Semver\Semver::satisfiedBy(PHP_VERSION, $version)
```

faissaloux: Yeah it's better, I have same idea as @owenvoke but I think it's better to use `composer/semver` than creating a parser for it on a test framework. I'm working on it...

Thanks everyone!

owenvoke (approved):

nunomaduro:
Added here: https://github.com/pestphp/pest/commit/5258e569c19cb782b07a12027f189b78b216dbd1.

owenvoke:
Hmm. 🤔 My main thought on using `composer/semver` was so that you could have tests which run between two versions of PHP, or similar cases (e.g. `>=8.1 || <=8.3` or `^8.3`).

I guess in theory you can just use 2 of them (e.g. `->skipOnPhp('>=8.1')->skipOnPhp('<=8.3')`), which is fine. 👍🏻

faissaloux:
Yeah same thought!

Issue #1054, opened 2024-01-04
[Improvement]: Extend 'depends' to allow referencing methods in other Testfiles


apmstage:
## Description
Allow the `test()->depends(...)` method to reference methods from other file like the underlying @depends annotation from PHPUnit can do.

## Motivation
In of our projects we've been using pest for integration testing on the staging system. These tests include general system requirements tests and some end-to-end feature tests that rely on the general system availability. Till now every test was in one file but this doesn't scale very well across a team so we would like to abstract the general tests into there own test file and reference theme via the `depends` on the feature tests.

## Interface (DX)

``` php
test('test model creation', function() { 
    # [...]
})->depends([ '\Tests\Feature\DbTest', 'successful db migration' ]);
```

## Implementation steps
1. Add class functionality to Pest\PendingCalls\TestCall -> depends
``` php
    /**
     * Sets the test depends.
     */
    public function depends(string|array ...$depends): self
    {
        foreach ($depends as $depend) {
            // Unpack and format class reference as array
            if (is_array($depend)) {
                if (count($depend) !== 2)
                    throw new InvalidArgumentException('Depends array must have 2 elements.');
                [$file, $test] = $depend;
                $file = ltrim($file, '\\');
                $depend = 'P\\'. $file . '::' . $test;
            }
            // Add depends
            $this->testCaseMethod->depends[] = $depend;
        }
        return $this;
    }
```
2. Adjust logic in @depends annotation generator Pest\Factories\Annotations\Depends -> __invoke
``` php
    /**
     * {@inheritdoc}
     */
    public function __invoke(TestCaseMethodFactory $method, array $annotations): array
    {
        foreach ($method->depends as $depend) {
            $method_name = $depend;

            // Split class and method name
            $class = null;
            if (str_contains($depend, '::')) {
                $parts = explode('::', $depend);
                [$class, $method_name] = $parts;
            }

            // Convert method name to evaluable string
            $depend = Str::evaluable($method->describing !== null ? Str::describe($method->describing, $method_name) : $method_name);

            // Add class name to method name and add annotation
            if ($class !== null)
                $depend = "$class::$depend";
            $annotations[] = "@depends $depend";
        }

        return $annotations;
    }
```

Abdulmajeed-Jamaan:
This would be helpful, my use case as package developer:

I have set of features based on a relation defined in eloquent model.

I want to run the some tests only if the relation has specific signature.

The tests could be in different describe context and/or different files.

[...]

PR #1059, opened 2024-01-08
[2.x] Support test dependencies across different classes

apmstage:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

Allow the test()->depends(...) method to reference methods from other file like the underlying https://github.com/Depends annotation from PHPUnit can do.

### Related:

Feature Request Issue: #1054 

### Motivation

In of our projects we've been using pest for integration testing on the staging system. These tests include general system requirements tests and some end-to-end feature tests that rely on the general system availability. Till now every test was in one file but this doesn't scale very well across a team so we would like to abstract the general tests into there own test file and reference theme via the `depends` on the feature tests.

### Interface (DX)

``` php
test('test model creation', function() { 
    # [...]
})->depends([ '\Tests\Feature\DbTest', 'successful db migration' ]);
```

### Implementation steps
1. Add class functionality to Pest\PendingCalls\TestCall -> depends
``` php
    /**
     * Sets the test depends.
     */
    public function depends(string|array ...$depends): self
    {
        foreach ($depends as $depend) {
            // Unpack and format class reference as array
            if (is_array($depend)) {
                if (count($depend) !== 2)
                    throw new InvalidArgumentException('Depends array must have 2 elements.');
                [$file, $test] = $depend;
                $file = ltrim($file, '\\');
                $depend = 'P\\'. $file . '::' . $test;
            }
            // Add depends
            $this->testCaseMethod->depends[] = $depend;
        }
        return $this;
    }
```
2. Adjust logic in @depends annotation generator Pest\Factories\Annotations\Depends -> __invoke
``` php
    /**
     * {@inheritdoc}
     */
    public function __invoke(TestCaseMethodFactory $method, array $annotations): array
    {
        foreach ($method->depends as $depend) {
            $method_name = $depend;

            // Split class and method name
            $class = null;
            if (str_contains($depend, '::')) {
                $parts = explode('::', $depend);
                [$class, $method_name] = $parts;
            }

            // Convert method name to evaluable string
            $depend = Str::evaluable($method->describing !== null ? Str::describe($method->describing, $method_name) : $method_name);

            // Add class name to method name and add annotation
            if ($class !== null)
                $depend = "$class::$depend";
            $annotations[] = "@depends $depend";
        }

        return $annotations;
    }
```
3. Added new tests for this feature

### DIFF
--- a/src/Factories/Annotations/Depends.php
+++ b/src/Factories/Annotations/Depends.php
@@ -19,8 +19,18 @@ final class Depends implements AddsAnnotations
     public function __invoke(TestCaseMethodFactory $method, array $annotations): array
     {
         foreach ($method->depends as $depend) {
+            // Split class and method name
+            $class = null;
+            if (str_contains($depend, '::')) {
+                [$class, $depend] = explode('::', $depend);
+            }
+
             $depend = Str::evaluable($method->describing !== null ? Str::describe($method->describing, $depend) : $depend);
 
+            // Add class name to method name and add annotation
+            if ($class !== null) {
+                $depend = "$class::$depend";
+            }
             $annotations[] = "@depends $depend";
         }
 
--- a/src/PendingCalls/TestCall.php
+++ b/src/PendingCalls/TestCall.php
@@ -149,10 +149,25 @@ final class TestCall
 
     /**
      * Sets the test depends.
+     *
+     * @param  array<string|array<string, string>>  $depends
      */
-    public function depends(string ...$depends): self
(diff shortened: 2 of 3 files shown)


nunomaduro:
Appreciate the effort on this feature. However, I don't think it is useful at all. People simply are not doing this.

apmstage:
Sad to see because from my perspective it would only widen the feature pool of pest. 
But thx for your opinion regarding this topic.

PR #1061, opened 2024-01-12
[3.x] Add a `test()->after()` helper to create test-specific teardown callbacks

emmadesilva:
## Abstract

Adds a chainable helper to add a teardown closure specific to the test function, something that has gotten great response on Twitter so far. https://twitter.com/CodeWithCaen/status/1745752831385055626/photo/1

Targets v3, as requested by Nuno on Twitter.

### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

This feature makes it really easy to create teardown logic specific to a single test function, by chaining the new `->after()` method to the test setup. 

```php
test('file parsing', function () {
    touch('example.txt');

    expect(Parser::parse('example.txt'))->toBe('foo');
})->after(function () {
    // This closure gets called regardless of the test outcome
    unlink('example.txt');
});
```

This is really useful for cleaning up after an impure test, or to clear a service container, but you only need it for a single test function. Of course these things could be done at the end of a test within its main function body, but if the test fails the cleanup wouldn't be executed. This solves that as the closure gets executed regardless of the test outcome. It also makes the code cleaner by keeping cleanup logic out of the actual test declaration.

### Implementation:

As this is my first PR to Pest, I'm not fully familiar with the internals of the framework, so I opted for what I after looking through the related code found to be the simplest implementation and tried to match the coding style of the related code. Upon code review I'm more than happy to adjust the implementation to fit the codebase better, if necessary.

### Future features:

This could of course be enhanced by allowing things like dependency injection for the `after` closure, but I wanted to keep the initial feature as light as possible. I could also see the use for adding a `before` helper for test setups, however I think the `after` helper has the broadest use case.

### DIFF
--- a/src/Factories/TestCaseMethodFactory.php
+++ b/src/Factories/TestCaseMethodFactory.php
@@ -214,7 +214,7 @@ final class TestCaseMethodFactory
      */
     public function __destruct()
     {
-        if ($this->after !== null) {
+        if ($this->after instanceof Closure) {
             ($this->after)();
         }
     }
--- a/src/Factories/TestCaseMethodFactory.php
+++ b/src/Factories/TestCaseMethodFactory.php
@@ -208,4 +208,14 @@ final class TestCaseMethodFactory
 
         EOF;
     }
+
+    /**
+     * Execute the 'after' callback, if set.
+     */
+    public function __destruct()
+    {
+        if ($this->after !== null) {
+            ($this->after)();
+        }
+    }
 }
--- a/src/Factories/TestCaseMethodFactory.php
+++ b/src/Factories/TestCaseMethodFactory.php
@@ -66,6 +66,11 @@ final class TestCaseMethodFactory
     public array $groups = [];
 
     /**
+     * Callback to run after the test has been executed.
+     */
+    public ?Closure $after = null;
+
+    /**
      * Creates a new test case method factory instance.
      */
     public function __construct(
--- a/src/PendingCalls/TestCall.php
+++ b/src/PendingCalls/TestCall.php
@@ -404,6 +404,16 @@ final class TestCall
     }
 
     /**
+     * Registers a closure to be called after the test has been executed.
+     */
+    public function after(Closure $after): self
+    {
+        $this->testCaseMethod->after = $after;
+
+        return $this;
+    }
+
+    /**
      * Saves the property accessors to be used on the target.
      */
     public function __get(string $name): self
--- a/tests/Hooks/AfterTest.php
+++ b/tests/Hooks/AfterTest.php
@@ -1,0 +1,15 @@
+<?php
+
+test('after helper can be chained', function () {
+    expect(true)->toBeTrue();
+})->after(function () {
+    // Example
+});
+
+test('after closure is called', function () use (&$afterWasCalled) {
+    $afterWasCalled = true;
+
+    expect(true)->toBeTrue();
+})->after(function () use (&$afterWasCalled) {
+    expect($afterWasCalled)->toBeTrue();
+});


devajmeireles (approved):

devajmeireles:
LGTM. Thanks!

src/PendingCalls/TestCall.php
```diff
@@ -403,6 +403,16 @@ public function throwsNoExceptions(): self
         return $this;
     }
 
+    /**
+     * Registers a closure to be called after the test has been executed.
+     */
+    public function after(Closure $after): self
```
sshymko, line 409: It would make sense to support any valid callback, not just closures, for instance:
- Closure
- Callable class with `__invoke()` method
- Array callable syntax: `[$instance, 'method']`, `['ClassName', 'staticMethod']`
- Function callable: `'function_name'`

It can be implemented using `callable` pseudo-type annotation, for example:
```php
public function after(callable $callback): self
{
    $this->testCaseMethod->after = Closure::fromCallable($callback);
}
```

src/Factories/TestCaseMethodFactory.php
```diff
@@ -203,4 +208,14 @@ public static function $dataProviderName()
 
         EOF;
     }
+
+    /**
+     * Execute the 'after' callback, if set.
+     */
+    public function __destruct()
```
sshymko, line 215: In my experience, putting logic in destructor is a dangerous practice. Even though it may work now, it relies on assumptions about the timing of when the object will be destroyed that can (accidentally) change in the future. For instance, the testing framework may keep objects in memory for longer, adding any circular references would make this behavior subject to the garbage collection, etc.

It would be more reliable (and less magical) to make the invocation explicit by calling a method.

tests/Hooks/AfterTest.php
```diff
@@ -0,0 +1,15 @@
...
+})->after(function () {
+    // Example
+});
+
+test('after closure is called', function () use (&$afterWasCalled) {
+    $afterWasCalled = true;
+
+    expect(true)->toBeTrue();
+})->after(function () use (&$afterWasCalled) {
+    expect($afterWasCalled)->toBeTrue();
```
sshymko, line 14: Commenting out the following line should fail the test, right?
```php
($this->after)();
```

It seems that this test passes even when a callback is never executed. If that's the case, the test is not really testing anything. The test should fail when the functionality under test is broken (the callback is never called).

It should be something like this:
```php
$afterWasCalled = false;

test('after helper can be chained', function () use (&$afterWasCalled) {
    expect($afterWasCalled)->toBeFalse();
})->after(function () use (&$afterWasCalled) {
    expect($afterWasCalled)->toBeFalse();
    $afterWasCalled = true;
});

test('after closure is called', function () use (&$afterWasCalled) {
    expect($afterWasCalled)->toBeTrue();
});

src/PendingCalls/TestCall.php
emmadesilva: Responding to all your reviews here. Thanks for taking the time to leave the great feedback! As this PR was already approved I'm going to hold off on making any changes until I hear more from the maintainers on which route they want to go in. Again, thanks for taking the time, I really appreciate it!

nunomaduro:
This feature cannot be coded this way. I can see, even without testing, that this won't work with `only` or filtering. I will handle this one myself on the 3.x branch. Thanks tho!

emmadesilva:
> This feature cannot be coded this way. I can see, even without testing, that this won't work with `only` or filtering. I will handle this one myself on the 3.x branch. Thanks tho!

Okay cool! Feel free to add me as a co-author of the PR if you are able to build upon this :)

PR #1069, opened 2024-01-20
[2.x] Improve coverage output message on failing minimum requirements

davybaccaert:
### What:

- [ ] Bug Fix
- [ ] New Feature
- [x] Improvement

### Description:
Improves the readability of the console output when the minimum required percentage is not met. 
As an added benefit, the value that is most important is at the end of the line, making it easier to spot.


owenvoke (approved):
This makes more sense grammatically, at least to me. :eyes:

Note: Not sure if this requires any changes to tests.

PR #1081, opened 2024-01-28
[2.x] Track vendor changes

nuernbergerA:
we override some vendor file
to keep track of changes of those file we keep the hash to compare it later

i had to move two files to match the vendor file structure
* WarmCodeCoverageCacheCommand
* TestSkippedSubscriber


nunomaduro:
This is crazy smart btw! Do you plan to finish this? Or should I do something?

nuernbergerA:
@nunomaduro im handing this over to you.
the question is if you wanna run this test on every matrix-run or however you wanna be notified about changes

nunomaduro:
> on every matrix-run 

This one.

nuernbergerA:
@nunomaduro can you help here with the snapshot situation?

PR #982, opened 2023-10-14
[2.x] add `toBeEmailAddress()` expectation

JonPurvis:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

This PR adds a `toBeEmailAddress()` Expectation, to validate a value is an email address. 

```php
<?php

it ('ensures the email of the user is actually an email', function () {
    expect($user->email)->toBeEmailAddress();
});
````

PR #1088, opened 2024-02-11
[2.x] Adds toBeEmail() Expectation

LucasLaurens:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

This PR adds a toBeEmail() expectation, to validate whether a value is a valid EMAIL.

### Related:

```php
expect('test@example.com')->toBeEmail();
expect('test@localhost')->not->toBeEmail();
```

### DIFF
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -1177,4 +1177,23 @@ final class Expectation
 
         return $this;
     }
+
+    /**
+     * Asserts that the value is an email
+     *
+     * @return self<TValue>
+     */
+    public function toBeEmail(string $message = ''): self
+    {
+        if ($message === '') {
+            $message = "Failed asserting that {$this->value} is an email.";
+        }
+
+        Assert::assertTrue(
+            Str::isEmail((string) $this->value),
+            $message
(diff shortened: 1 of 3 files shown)


LucasLaurens:
If the PR is accepted, I could create one for documentation.

LucasLaurens:
I also have a question. Do I need to update the snapshots? This would seem normal if this feature is accepted.

LucasLaurens:
I hadn't seen it before. But my PR matches [this one](https://github.com/pestphp/pest/pull/982) exactly. So it doesn't make sense anymore.

LucasLaurens:
Perhaps in the long term, if the intention is to add an expectation on the email, this could be improved with a DNS check.

> [!NOTE]
> I was able to extend an expectation on a personal project where I check for a specific use case the validity of the email according to an __RFC__ and __DNS__ validation.

```php
/** DNS Validation */
$dns = new DNSCheckValidation();

if (
    !$dns->isValid(
        email: $value,
        emailLexer: new \Egulias\EmailValidator\EmailLexer()
    )
) {
    return false;
}
```

PR #1090, opened 2024-02-12
[2.x] Adds toBePowerOf() Expectation

LucasLaurens:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

This PR adds a `toBePowerOf()` expectation, to validate whether a value is a power of another.
If the PR is accepted, I could create one for documentation.

> [!NOTE]
> What's more, if this code is accepted then I could also create expectations for `exponential` and `square root`, for example.

### Related:

```php
expect(8)->toBePowerOf(2);
expect(7)->not->toBePowerOf(2);
```

### DIFF
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -1177,4 +1177,25 @@ final class Expectation
 
         return $this;
     }
+
+    /**
+     * Asserts that the value is power of $expected.
+     *
+     * @return self<TValue>
+     */
+    public function toBePowerOf(int|float $expected, string $message = ''): self
+    {
+        $logarithme = log(
+            (float) $this->value,
+            (float) $expected
+        );
+
+        Assert::assertTrue(
+            $expected === 1
+            || floor($logarithme) === $logarithme,
+            $message
+        );
+
+        return $this;
+    }
 }
--- a/tests/.snapshots/success.txt
+++ b/tests/.snapshots/success.txt
@@ -543,6 +543,12 @@
   ✓ failures with custom message
   ✓ not failures
 
+   PASS  Tests\Features\Expect\toBePowerOf
+  ✓ Check whether one number is a power of another → pass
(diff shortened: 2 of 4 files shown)


LucasLaurens:
Could be improve by something like 

```php
function isPower($x, $y): bool
{
    if ($x === 1) {
    	return $y === 1;	
    }
 
    $pow = 1;
    while ($pow < $y) {
    	$pow *= $x;	
    }
 
    return $pow === $y;
}
```

owenvoke:
Just a thought, I feel like these would be quite nice in a maths (or math) plugin for Pest possibly. I'm not sure how commonly they'd be used.

I can see that there could be quite a lot of mathematical expectations that people would want, but which maybe shouldn't be in core as it'd add additional code.

Don't mind either way though, so going to leave this for someone else.

LucasLaurens:
> Just a thought, I feel like these would be quite nice in a maths (or math) plugin for Pest possibly. I'm not sure how commonly they'd be used.
> 
> I can see that there could be quite a lot of mathematical expectations that people would want, but which maybe shouldn't be in core as it'd add additional code.
> 
> Don't mind either way though, so going to leave this for someone else.

I totally agree with you. I think having an official plugin created by the Pest team to handle all mathematical cases would be really interesting. 

If I've done it this way, it's also because I assumed that I was comparing two numbers and that the assertion would return a boolean. 

Whether in a plugin or in the main code, I did it because I came across the need on one of my projects and thought it would be interesting to share it with the community.

I really thank you for your feedback, which I find very relevant. In any case, it was a pleasure for me to participate in the evolution of Pest, so I thought I'd propose this feature to help the community in certain cases.

This with the possibility of adding more if this first expectation is accepted and appreciated enough to add more (i mean math expectations).

nunomaduro:
I don't think people will actually use this; maybe something for a plugin.

PR #1097, opened 2024-02-20
[2.x] Modify `Result::exitCode` logic to address warning handling with `--fail-on-warning`

### LINKED ISSUE #1074: [Bug]: Pest doesn't fail on warning
### What Happened

When a warning occurs in my tests, I expect that pest should return a non-zero exit code when running with the `--fail-on-warning` flag.

### How to Reproduce

Below is the smallest reproduction steps to show what I'm seeing:

image

This shows creating a new composer project, installing pest, adding a warning to the test, and pest returning 0 for the exit code.

### Sample Repository

_No response_

### Pest Version

2.32.2

### PHP Version

8.2.6

### Operation System

macOS, Linux

### Notes

I'm currently running into a problem where I expect pest to return an exit code of 1 when an error occurs in the test.

I can see there are some differences with how PHPUnit derives the exit code:
https://github.com/sebastianbergmann/phpunit/blob/main/src/TextUI/ShellExitCodeCalculator.php

and how pest does it:
https://github.com/pestphp/pest/blob/2.x/src/Result.php

Is there a reason for this, or could I look to bring the pest one more in line with the PHPUnit one?

---

This is the equivalent test run for phpunit, you can see here I do a non-zero exit code:

image

arifszn:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Pest now fails (non-zero exit code) when warnings are encountered with the `--fail-on-warning` flag. Previously, it passed (zero exit code) even with warnings present as it returned `SUCCESS_EXIT` before checking the warning count.

### Related:

Fix #1074.

### DIFF
--- a/src/Result.php
+++ b/src/Result.php
@@ -40,9 +40,20 @@ final class Result
      */
     public static function exitCode(Configuration $configuration, TestResult $result): int
     {
-        if ($result->wasSuccessfulIgnoringPhpunitWarnings()
-            && ! $result->hasTestTriggeredPhpunitWarningEvents()) {
-            return self::SUCCESS_EXIT;
+        if ($result->wasSuccessfulIgnoringPhpunitWarnings()) {
+            if ($configuration->failOnWarning()) {
+                $warnings = $result->numberOfTestsWithTestTriggeredPhpunitWarningEvents()
+                    + count($result->warnings())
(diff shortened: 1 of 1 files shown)


arifszn:
Ping @nunomaduro

rburgst:
also waiting for this fix!

PR #1119, opened 2024-03-21
Updated Functions::datasets() method by adding $scope parameter

DimionX:
### What:

- [ ] New Feature. Added the ability to use a custom scope for a datataset.

### Description:

Updated Functions::datasets() method signature by adding $scope optional parameter.

### DIFF
--- a/src/Functions.php
+++ b/src/Functions.php
@@ -69,9 +69,9 @@ if (! function_exists('dataset')) {
      *
      * @param  Closure|iterable<int|string, mixed>  $dataset
      */
-    function dataset(string $name, Closure|iterable $dataset): void
+    function dataset(string $name, Closure|iterable $dataset, ?string $scope = null): void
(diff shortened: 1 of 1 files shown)


fabio-ivona:
hi @DimionX could you elaborate more on the need behind this feature?

DimionX:
> hi @DimionX could you elaborate more on the need behind this feature?

Since Dataset Boot connects tests from only one test directory, the limits their use in other scopes. For example, for component or unit tests that are located in the "app" directory, it is impossible to use a dataset that is located in the test directory (tests).
Thus, you can manually specify which scope a particular dataset belongs to.

fabio-ivona:
@DimionX I see, would you mind to add some test for your changes?

nunomaduro:
Just place those "global" datasets on the "tests/Datasets" folder.

PR #1134, opened 2024-04-11
Fix nested datasets

### LINKED ISSUE #1121: [Bug]: Cannot use Datasets in subfolders
### What Happened

As part of my transition from Laravel 9 to Laravel 10, I am also upgrading from Pest 1.x to 2.x. During this process, I encountered an issue concerning the resolution of datasets located within subfolders.

I'm not the only one with this problem, please refer to [laracast question] for more details. (https://laracasts.com/discuss/channels/testing/pest-global-datasets-not-allowed-in-modular-project)

### How to Reproduce

Construct a dataset and place it within a subfolder inside the Datasets directory. Afterwards, execute the command `./vendor/bin/pest MyTest.php`, which unfortunately results in a failure.

However, this situation is peculiar as executing the test through PHPStorm does not present any issues, indicating that the problem only occurs when running the test from the terminal...

### Sample Repository

_No response_

### Pest Version

2.34.4

### PHP Version

8.2.16

### Operation System

macOS

### Notes

_No response_

fabio-ivona:
This PR will allow to have nested datasets inside `Datasets` folder

**note**

this solution seems to work in standard mode, but fails with --parallel flag set up, it seems that the nested dataset si found, but the tests are filtered out by Paratest

can't get why, would be great if someone could take a look at this, maybe I'm missing something :(

fixes #1121

### DIFF
--- a/src/Support/DatasetInfo.php
+++ b/src/Support/DatasetInfo.php
@@ -20,6 +20,11 @@ final class DatasetInfo
         return basename(dirname($file)) === self::DATASETS_DIR_NAME;
     }
 
+    public static function isNestedInsideADatasetsDirectory(string $file): bool
+    {
+        return str_contains($file, DIRECTORY_SEPARATOR.self::DATASETS_DIR_NAME.DIRECTORY_SEPARATOR);
+    }
+
     public static function isADatasetsFile(string $file): bool
     {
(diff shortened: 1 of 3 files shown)


nunomaduro:
Re-open this, if you ever finish it.

PR #1163, opened 2024-06-02
Fix Laravel 11 TestCase.php stub

Danger-Mkh:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

As Laravel 11 `tests/TestCase.php` it doesn't have `use CreatesApplication;` statement.

So in fresh Laravel install if you install Pest right after that, by default the `tests/TestCase.php` exists so it would recognized.

I manually removed the `tests/TestCase.php` so `pest:install` replaced the stubs, after running the `php artisan test` command i encouraged an error that shows  `use CreatesApplication;` is not found.

### DIFF
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -468,6 +468,18 @@ final class Expectation
     }
 
     /**
+     * Asserts that the value is a list.
+     *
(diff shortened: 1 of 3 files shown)


nunomaduro:
this is fixed.

PR #1170, opened 2024-06-08
[3.x] Arch Presets

nunomaduro:
> This pull request is a work in progress, and presets are currently being defined...

This pull request introduces the concept of "Arch Presets" to Pest 3; witch a simple way of executing a "group" of arch tests predefined by the Pest team. 

While the existing arch tests (https://pestphp.com/docs/arch-testing) are powerful, they still require you to granular define the rules you want to enforce. This is great for fine-grained control, but can be a bit verbose for simple use-cases.

```php
// before:
arch('globals')
    ->expect(['dd', 'dump'])
    ->not->toBeUsed();

// + 100 rules written by you
```

```php
// after:
arch()->preset()->base(); // == 100 rules...

arch()->preset()->laravel(); // == 120 rules...

arch()->preset()->security(); // == 50 rules
```


src/ArchPresets/Base.php
```diff
@@ -0,0 +1,76 @@
...
+            'var_dump',
+            'phpinfo',
+            'echo',
+            'print',
+            'print_r',
+            'var_export',
+            'xdebug_break',
+            'xdebug_call_class',
+            'xdebug_call_file',
+            'xdebug_call_(int',
```
giggsey:
```suggestion
            'xdebug_call_int',
```

giggsey: I can't see any proof that that function exists. Was this meant to be `xdebug_call_function` ?

giggsey: @nunomaduro Not sure if you saw this comment

Issue #837, opened 2023-06-15
The feature "Filter by dirty files" requires "git"


jasonmccallister:
I get this one randomly through out the day. I'm running my site with ddev and connected via `ddev ssh`. I can run the command multiple times successfully, but at least ~10 times a day - in the same terminal session - I get the following error:

```
The feature "Filter by dirty files" requires "git".
```

This happens when running `php artisan test --dirty` and `./vendor/bin/pest --dirty`.

**Details**

PHP: 8.2.3
Pest: 2.6.3

owenvoke:
Hmm, I'm assuming this is due to [this check](https://github.com/pestphp/pest/blob/c0af671ca214c1fb43133daaa0a667f57b11ef10/src/TestCaseFilters/GitDirtyTestCaseFilter.php#L49).

It only checks whether the command failed, not whether Git actually exists. So it's probably just that the command returned an invalid exit code, and might be hiding the underlying issue.

jasonmccallister:
Alright, I edited that file in vendor and added the following:

```
if (! $process->isSuccessful()) {
    dd($process->getErrorOutput());
    throw new MissingDependency('Filter by dirty files', 'git');
}
```

I finally hit that error and this is the output:

```
fatal: detected dubious ownership in repository at '/var/www/html'\n
To add an exception for this directory, call:\n
\n
\tgit config --global --add safe.directory /var/www/html\n
```

jasonmccallister:
Running `git config --global --add safe.directory /var/www/html` does resolve the problem, but it still occurs randomly. I'm assuming this has something to do with `ddev` and mutagen mounts?

devajmeireles:
Hey, @jasonmccallister . Thanks for your report.

I was not able to reproduce this, so I assume this issue is specific to your environment. To improve our organization, I will close this topic. Please feel free to open a new topic, whatever you need.

kerunaru:
@jasonmccallister Just four your information. This happens when Docker mounts the volume with your code; your user used to log in the container differs from the one who creates all the directory hierarchy and git complains about that.

PR #1187, opened 2024-07-09
Add dubious ownership error control

kerunaru:
Now the filter informs the user if git complains about the ownership of the current repository folder. Before this change, Pest kept saying that git was not installed if the process failed.

### What:

- [X] Bug Fix
- [ ] New Feature

### Description:

Using Docker on a GNU/Linux machine, Pest throws an error complaining about missing Git as a dependency although it is indeed installed. The real problem comes with the fact that the user used to mount the Docker volumes differs from the one assigned to the current user, so what is happening is that Git is complaining about dubious folder ownership.

This behaviour can be reproduced by performing a `chmod 100000:100000 .` (or whatever non-existing user and group ids) and trying to check the repository status with `git status`. Launching `pest --dirty` in this context will throw a `MissingDependency` exception which is not true.

This PR tries to check the exit code related to this Git error (`128`) and inform the user accordingly, saving lots of research about what is happening.

### Related:

https://github.com/pestphp/pest/issues/837

### DIFF
--- a/src/TestCaseFilters/GitDirtyTestCaseFilter.php
+++ b/src/TestCaseFilters/GitDirtyTestCaseFilter.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Pest\TestCaseFilters;
 
 use Pest\Contracts\TestCaseFilter;
+use Pest\Exceptions\InvalidArgumentException;
 use Pest\Exceptions\MissingDependency;
 use Pest\Exceptions\NoDirtyTestsFound;
 use Pest\Panic;
@@ -13,6 +14,8 @@ use Symfony\Component\Process\Process;
 
 final class GitDirtyTestCaseFilter implements TestCaseFilter
 {
+    private const EXIT_CODE_GIT_DUBIOUS_OWNERSHIP = 128;
+
     /**
      * @var string[]|null
      */
@@ -46,7 +49,11 @@ final class GitDirtyTestCaseFilter implements TestCaseFilter
         $process = new Process(['git', 'status', '--short', '--', '*.php']);
         $process->run();
 
-        if (! $process->isSuccessful()) {
+        if ($process->getExitCode() === self::EXIT_CODE_GIT_DUBIOUS_OWNERSHIP) {
+            throw new InvalidArgumentException('Dubiuous folder ownership');
+        }
+
+        if (!$process->isSuccessful()) {
             throw new MissingDependency('Filter by dirty files', 'git');
         }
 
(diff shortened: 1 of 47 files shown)


src/TestCaseFilters/GitDirtyTestCaseFilter.php
```diff
@@ -13,6 +14,8 @@
 
 final class GitDirtyTestCaseFilter implements TestCaseFilter
 {
+    private const EXIT_CODE_GIT_DUBIOUS_OWNERSHIP = 128;
```
owenvoke: Isn't exit code 128 a generic exit code for a fatal error, that could be for various reasons (e.g. if the directory is not a Git repository, or ? 🤔

Perhaps the best solution would be to name this `EXIT_CODE_FATAL_ERROR`, and then where we throw the exception, just output the `$process->getErrorOutput()` 🤷🏻

The error output for dubious ownership should contain:
```
fatal: detected dubious ownership
```

src/TestCaseFilters/GitDirtyTestCaseFilter.php
```diff
@@ -46,7 +49,11 @@ private function loadChangedFiles(): void
         $process = new Process(['git', 'status', '--short', '--', '*.php']);
         $process->run();
 
-        if (! $process->isSuccessful()) {
+        if (!$process->isSuccessful()) {
+            if ($process->getExitCode() === self::EXIT_CODE_GIT_DUBIOUS_OWNERSHIP) {
+                throw new InvalidArgumentException('Dubious folder ownership');
```
owenvoke: Perhaps a different exception type would make more sense here. 🤔 `InvalidArgumentException` seems a bit weird possibly.

src/TestCaseFilters/GitDirtyTestCaseFilter.php
kerunaru: I was considering that too... Let me change it.

src/TestCaseFilters/GitDirtyTestCaseFilter.php
kerunaru: Yeah... I use the generic one because `MissingDependency` inherits from it and I don't know if I could make a new one. I'll change that.

src/TestCaseFilters/GitDirtyTestCaseFilter.php
kerunaru: Done.

src/TestCaseFilters/GitDirtyTestCaseFilter.php
kerunaru: Done.

nunomaduro:
Can you rebase this PR, and re-submit it? thanks!

PR #1188, opened 2024-07-09
[2.x] Fix non-zero exitcodes when using CLI --fail-on-xxx config options

grEvenX:
### What:
- [x] Bug Fix

### Description:
This commit updates the Result.php class so that it returns the correct exit code when PHPunit is passed a configuration option like '--fail-on-warning'.

In addition, it adds a condition for '--fail-on-notice' which is supported based on the output from `pest --help`, though not listed in the online [Documentation](https://github.com/pestphp/docs/blob/master/cli-api-reference.md).

Screenshot of pest --help output

### DIFF
--- a/src/Result.php
+++ b/src/Result.php
@@ -40,13 +40,14 @@ final class Result
      */
     public static function exitCode(Configuration $configuration, TestResult $result): int
     {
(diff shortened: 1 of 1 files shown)


nunomaduro:
this should be already fixed.

PR #1205, opened 2024-08-24
Add Exception Handling for undefined method and  Status File Opening

ghost:
Adds a check to ensure the status file is successfully opened. If not, it throws a [`RuntimeException`]

**Code:**
```php
if (!$statusFile) {
    throw new \RuntimeException('Failed to open status file');
}
```

```php
 if (!method_exists($this, $name)) {
        throw new \BadMethodCallException("Method $name does not exist.");
    }
```


nunomaduro:
No plans for this.

PR #1212, opened 2024-08-30
feat: [3.x] add `uses` counterpart to `covers`

JonathanGawrych:
### What:

- [ ] Bug Fix
- [X] New Feature

### Description:

Right now, pest supports `covers` for classes, functions, and "nothing" to include sections of code to track coverage for. This adds the `uses` counterpart, which exclude sections of code. (see https://docs.phpunit.de/en/11.3/code-coverage.html#targeting-units-of-code). This would be especially useful for `--mutate` allowing you to disable coverage (and therefore mutations) for specific parts of code on a per-test basis. While adding this, I saw that phpunit added a `coversTrait` and `usesTrait` in version 11.2, so I also included that in this MR.

Locally I ran `composer lint`, `composer update:snapshots`, and `composer test`. Almost everything succeeded but I had trouble with the `tests/Visual/Success.php` test. It also failed on the 3.x branch, so I'm hoping it's not related to my changes.

Thanks!

### DIFF
--- a/src/PendingCalls/TestCall.php
+++ b/src/PendingCalls/TestCall.php
@@ -488,15 +488,18 @@ final class TestCall
     public function covers(string ...$classesOrFunctions): self
     {
         foreach ($classesOrFunctions as $classOrFunction) {
-            $isClass = class_exists($classOrFunction) || trait_exists($classOrFunction);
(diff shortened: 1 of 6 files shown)


nunomaduro:
Can you rebase this pull request?

Issue #1192, opened 2024-07-14
Call to undefined method TestCase::get()


SuheylZ:
While running the tests I'm getting this error

> ReflectionException: Call to undefined method PHPUnit\Framework\TestCase::get()
> at vendor/pestphp/pest-plugin-laravel/src/Http.php:189
> at tests/Feature/HealthTest.php:8

I tried many things but nothing works. My Unit tests run correctly with no issue. there is no get, post delete request. but my feature tests don't run at all. I tried searching for solution for days and now I'm posting my issue here. Can anyone help me identify the cause and how can I solve it?

Directory Structure

|---tests
|--- Unit
|----Features
|---------HealthTest.php
|----Pest.php
|----TestCase.php
|-phpunit.xml

**Pest.php**

```
<?php declare(strict_types=1);
/*
|--------------------------------------------------------------------------
| Test Case
| The closure you provide to your test functions is always bound to a specific PHPUnit test
| case class. By default, that class is "PHPUnit\Framework\TestCase". Of course, you may
| need to change it using the "uses()" function to bind a different classes or traits.
|
*/

use PHPUnit\Framework\TestCase;

uses(TestCase::class)->in('Feature');
uses(TestCase::class)->in('Unit');


/*
|--------------------------------------------------------------------------
| Expectations
|--------------------------------------------------------------------------
|
| When you're writing tests, you often need to check that values meet certain conditions. The
| "expect()" function gives you access to a set of "expectations" methods that you can use
| to assert different things. Of course, you may extend the Expectation API at any time.
|
*/

expect()->extend('toBeOne', function () {
    return $this->toBe(1);
});

/*
|--------------------------------------------------------------------------
| Functions
|--------------------------------------------------------------------------
|
| While Pest is very powerful out-of-the-box, you may have some testing code specific to your
| project that you don't want to repeat in every file. Here you can also expose helpers as
| global functions to help you to reduce the number of lines of code in your test files.
|
*/

function something()
{
    // ..
}

```

**HealthTest.php**

```
<?php declare(strict_types=1);

//uses(Tests\TestCase::class);

use function Pest\Laravel\{get};

test('service healthy', function () {
    $response = get('/');
    $response->assertStatus(503);
});


test('pings', function (){
    $res = get('/ping', ['Device-Id'=> '3984732984792420934231']);
    $res->assertStatus(200);
});
```

 **phpunit.xml**

```
<?xml version="1.0" encoding="UTF-8"?>
[...]

PR #1213, opened 2024-09-01
Update Init.php to change how Laravel is detected

gregmsanderson:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Running `vendor/bin/pest --init` currently checks for Laravel by looking for `laravel/laravel` however a new Laravel 11 install does not include that.

As a result `isLaravelInstalled()` incorrectly returns false

That means it does not add the Laravel stub and so defaults to using phpunit for the base case in TestCase.php. Which then causes an undefined error when running any Feature test, including the example one.

(I'm not sure if this is the only file that does this check however manually changing this file and then running `vendor/bin/pest --init` appears to fix it for me.)

### Related:

https://github.com/pestphp/pest/issues/1192

### DIFF
--- a/src/Plugins/Init.php
+++ b/src/Plugins/Init.php
@@ -119,6 +119,6 @@ final class Init implements HandlesArguments
      */
     private function isLaravelInstalled(): bool
     {
-        return InstalledVersions::isInstalled('laravel/laravel');
+        return InstalledVersions::isInstalled('laravel/laravel') || InstalledVersions::isInstalled('laravel/framework');
     }
 }


src/Plugins/Init.php
```diff
@@ -119,6 +119,6 @@ public function init(): void
      */
     private function isLaravelInstalled(): bool
     {
-        return InstalledVersions::isInstalled('laravel/laravel');
+        return InstalledVersions::isInstalled('laravel/laravel') || InstalledVersions::isInstalled('laravel/framework');
```
owenvoke, line 122: I think an additional check is required (along with this one), as this would trigger on things such Laravel packages which shouldn't extend this testcase.

gregmsanderson: Ah, that's a good point. I'm not sure what the check would be to _exclude_ those.

I guess the _other_ way to do this would be for the user to tell the init command that Laravel is installed e.g `vendor/bin/pest --init --laravel`. That approach would mean the `isLaravelInstalled` check stays the same and so couldn't break anything (like a package). It would be nicer to automate it though.

owenvoke: I guess we could check if `artisan` exists maybe. 🤷🏻 That would handle most cases.

This is how Tinkerwell does it:
https://github.com/beyondcode/tinkerwell/blob/main/src/Drivers/LaravelTinkerwellDriver.php#L7

gregmsanderson: That makes sense. Perhaps:

```php
private function isLaravelInstalled(): bool
{
   return file_exists($this->testSuite->rootPath.'/public/index.php') && file_exists($this->testSuite->rootPath.'/artisan');
}
```

... or get them all in there to cover all the bases ...

```php
private function isLaravelInstalled(): bool
{
    return InstalledVersions::isInstalled('laravel/framework') && file_exists($this->testSuite->rootPath.'/public/index.php') && file_exists($this->testSuite->rootPath.'/artisan');
}
```

... since it looks like the equivalent to Tinkerwell's path is `$this->testSuite->rootPath`.

Perhaps checking for `artisan` alone is enough? I can't imagine you'd have one without the other, or create a file called `artisan` if _not_ using Laravel.

henzeb: what if someone changed the name of that command file?

gregmsanderson: @henzeb If they did, yep, the new check would also fail to detect Laravel. Just like Tinkerwell would fail to.

Doing _that_ seems less likely (to me anyway) than someone changing the `name` in composer.json.

nunomaduro:
this was fixed on pest 3.

PR #1254, opened 2024-09-16
fix: update assignee URL for Jira

owenvoke:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

This extension is required for Jira, otherwise it results in a 404. 🤷🏻

### DIFF
--- a/src/Configuration/Project.php
+++ b/src/Configuration/Project.php
@@ -89,7 +89,7 @@ final class Project
     {
         $this->issues = "https://{$namespace}.atlassian.net/browse/{$project}-%s";
 
-        $this->assignees = "https://{$namespace}.atlassian.net/secure/ViewProfile?name=%s";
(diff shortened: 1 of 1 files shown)


src/Configuration/Project.php
```diff
@@ -89,7 +89,7 @@ public function jira(string $namespace, string $project): self
     {
         $this->issues = "https://{$namespace}.atlassian.net/browse/{$project}-%s";
 
-        $this->assignees = "https://{$namespace}.atlassian.net/secure/ViewProfile?name=%s";
+        $this->assignees = "https://{$namespace}.atlassian.net/secure/ViewProfile.jspa?name=%s";
```
owenvoke, line 92: I'm not sure if we want to have `?name=%s` here, or use `?email=%s` or `?username=` (name doesn't seem to work, at least from my usage).

PR #1267, opened 2024-09-19
feat: Add ability to extend `TestCall` to create custom chained helpers

JHWelch:
### What:

- [ ] Bug Fix
- [X] New Feature

### Description:

Adds the ability to `->extend(...)` `TestCall` in the same fashion that expectations are extendable.

```php
test()->extend('fooBar', function () {
    return $this->with([
        'foo',
        'bar',
    ]);
});
```

I had this idea after [discussing on Pinkary](https://pinkary.com/@JHWelch/questions/9d08842d-5f42-4e2d-9509-6117b868f054) and realize this was not currently a way Pest could be extended.

This opens up the framework for a whole new area of extensibility.

If this is something you want to add, I can PR the docs repo wherever makes sense.

### DIFF
--- a/src/PendingCalls/TestCall.php
+++ b/src/PendingCalls/TestCall.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Pest\PendingCalls;
 
 use Closure;
+use Pest\Concerns\Extendable;
 use Pest\Concerns\Testable;
 use Pest\Exceptions\InvalidArgumentException;
 use Pest\Exceptions\TestDescriptionMissing;
@@ -31,6 +32,9 @@ use PHPUnit\Framework\TestCase;
 final class TestCall // @phpstan-ignore-line
 {
     use Describable;
+    use Extendable {
+        extend as traitExtend;
+    }
 
     /**
      * The list of test case factory attributes.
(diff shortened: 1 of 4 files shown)


src/PendingCalls/TestCall.php
```diff
@@ -642,9 +646,25 @@ public function __get(string $name): self
...
+
         return $this->addChain(Backtrace::file(), Backtrace::line(), $name, $arguments);
     }
 
+    public function extend(string $name, Closure $extend): void
+    {
+        $this->traitExtend($name, $extend);
+
+        $this->description = "extend: $name";
+        $this->testCaseMethod->closure = fn (): null => null;
```
JHWelch, line 666: This is required because `extend` is being called non-statically. With this combination no errors are thrown and no output is included from a `test()->extend(...)` call.

I would need to add a different `StaticallyExtendable` or something like it to be able to use `TestCall::extend(...)` which is another possible API.

nunomaduro:
sorry, going to table this one for now.

JHWelch:
@nunomaduro Makes sense that you closed this getting ready for v4, but is there any appetite for this functionality?

I know there is a plugin I worked on where this would allow a much cleaner API. 

I can rework for the new version, but if you don't want this don't want to spend the time. I would create this as a plugin but obviously this is _enabling_ plugin functionality and can't be added another way (easily). 

Thanks as always, love Pest!

PR #1289, opened 2024-10-05
Add `toBeWithin` expectation method

ghost:
### What:
- [ ] Bug Fix
- [x] New Feature

### Description:

In this PR i have added a new `toBeWithin` expectation method to Pest. This method allows for easy assertion that a value falls within a specified range, inclusive of the boundaries.

### Changes

1. Added `toBeWithin` method to `Expectation` class
2. Added corresponding test suite for `toBeWithin`

### New Functionality

The `toBeWithin` method can be used as follows:
```php
expect(5)->toBeWithin(1, 10);
expect(3.14)->toBeWithin(3, 4);

```
### Test Coverage

A comprehensive test suite has been added, covering:
- Basic functionality with integers and floats
- Edge cases (PHP_INT_MAX, PHP_INT_MIN)
- Negative numbers
- Error cases (non-numeric values)
- Custom error messages

### DIFF
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -1159,4 +1159,21 @@ final class Expectation
 
         return $this;
     }
+
+    /**
+     * Asserts that the value is within a specified range.
+     *
+     * @return self<TValue>
+     */
+    public function toBeWithin(int|float $min, int|float $max, string $message = ''): self
+    {
+        if (! is_numeric($this->value)) {
+            InvalidExpectationValue::expected('numeric');
+        }
+
+        Assert::assertGreaterThanOrEqual($min, $this->value, $message);
+        Assert::assertLessThanOrEqual($max, $this->value, $message);
+
+        return $this;
+    }
 }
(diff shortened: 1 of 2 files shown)


faissaloux:
What's the difference between `toBeWithin` and `toBeBetween`?

ghost:
@faissaloux  Thanks for this question.

The main differences between these two methods are:

1. Parameter types:
   - `toBeBetween` accepts `int`, `float`, or `DateTimeInterface` for both parameters.
   - `toBeWithin` only accepts `int` or `float` for both parameters.

2. Value type checking:
   - `toBeWithin` explicitly checks if the value is numeric before performing the comparison.
   - `toBeBetween` doesn't perform this check, allowing for comparison of `DateTimeInterface` objects.

3. Semantic usage:
   - `toBeBetween` is more flexible and can be used with dates and times as well as numbers.
   - `toBeWithin` is specifically for numeric ranges.

In practice, for numeric values, these methods perform the same assertion. The main difference is in their intended use cases and the types of values they can handle. `toBeBetween` is more versatile, while `toBeWithin` is more specific to numeric ranges.

faissaloux:
I think we don't need another method for each type, since we have `toBeBetween`  that does the job.

nunomaduro:

PR #1290, opened 2024-10-06
feat: add toBeOneOf() expectation

dshafik:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

Adds a `toBeOneOf` expectation that allows you to ensure that a value matches **only** one of a set of assertions. This is complimentary to #1286.

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -32,6 +32,7 @@ use Pest\Matchers\Any;
 use Pest\Support\ExpectationPipeline;
 use Pest\Support\Reflection;
 use PHPUnit\Architecture\Elements\ObjectDescription;
+use PHPUnit\Framework\AssertionFailedError;
 use PHPUnit\Framework\ExpectationFailedException;
 use ReflectionEnum;
 use ReflectionMethod;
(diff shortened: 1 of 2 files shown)


src/Expectation.php
```diff
@@ -324,6 +325,41 @@ public function when(callable|bool $condition, callable $callback): self
...
+                $exceptions[] = $key;
+            }
+        }
+
+        if (count($matches) === 1) {
+            return $this;
+        }
+
+        if (count($matches) > 1) {
+            throw new ExpectationFailedException('Failed asserting value matches exactly one expectation (matches: '.implode(', ', $matches).').');
```
KorvinSzanto, line 357: In my opinion this should throw

src/Expectation.php
```diff
@@ -324,6 +325,41 @@ public function when(callable|bool $condition, callable $callback): self
...
+    /**
+     * Asserts that only one of the given tests pass with the given expectation target.
+     *
+     * @param  (\Closure(self<TValue>): (mixed|void))  ...$tests
+     * @return self<TValue>
+     */
+    public function toBeOneOf(Closure ...$tests): self
+    {
+        if ($tests === []) {
+            return $this;
```
KorvinSzanto, line 337: In my opinion this should throw

src/Expectation.php
KorvinSzanto, line 357: Rather than outputting the keys, I'd just output a count of passes and fails for a few reasons:

- Keys are limited to ints, or variable names due to using the splat operator rather than accepting an iterable, for example: `->toBeOneOf(thisIsTheOnlyWayToNameAnExpectation: fn($e) => ..., spaces_are_not_allowed: fn($e) => ...)`
- If you were accepting an iterable rather than using a splat, array keys aren't guaranteed to be unique
- There could be a ton of expectations passed to this function, causing a failure to output a huge error
- Just counting passes and failures reduces the memory overhead

You can also output the total count of assertions ran by checking the delta of `Assert::getCount()` before and after.

tests/Features/Expect/toBeOneOf.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+use PHPUnit\Framework\ExpectationFailedException;
+
+expect(true)->toBeTrue()->and(false)->toBeFalse();
```
KorvinSzanto, line 5:
```suggestion
```

nunomaduro:

Issue #1010, opened 2023-11-14
[Bug]: `DescribeCall` class is missing `with` method


paulshryock:
### What Happened

When I use `describe()->with()` in my tests, I get an error:

```plaintext
ArgumentCountError

Too few arguments to function PHPUnit\Runner\TestSuiteLoader::{closure}(),
0 passed in ~/Desktop/my-org/my-app/vendor/pestphp/pest/src/PendingCalls/DescribeCall.php
on line 56 and exactly 1 expected
```

### How to Reproduce

1. Install PHP 8.3.12 (`brew install php:8.3.12 && composer require php:8.3.12`)
2. Install Pest 3.2.5 (`composer require --dev pestphp/pest:3.2.5 -W`)
3. Add a test file with the code below (`touch ./tests/MyTest.php`)
4. Run tests (`./vendor/bin/pest`)

`tests/MyTest.php`:

```php
<?php

declare(strict_types=1);

function getRequestMethod(): ?string {
  return $_SERVER['REQUEST_METHOD'];
}

describe('when the http request method is set', function(string $method) {
  $server = $_SERVER;

  beforeEach(fn() => ($_SERVER['REQUEST_METHOD'] = $method));
  afterEach(fn() => ($_SERVER = $server));

  it('should get the request method', fn() =>
    expect(getRequestMethod())->toBe($method));
})->with([
  'CONNECT',
  'DELETE',
  'GET',
  'HEAD',
  'OPTIONS',
  'PATCH',
  'POST',
  'PUT',
  'TRACE',
]);

describe('when the http request method is not set', function() {
  $server = $_SERVER;

  beforeEach(function() {
    unset($_SERVER['REQUEST_METHOD']);
  });

  afterEach(fn() => ($_SERVER = $server));

  it('should get null', fn() => expect(getRequestMethod())->toBeNull());
});

```

Here is the result:

```
composer test -- --filter=My
> Composer\Config::disableProcessTimeout
> pest '--filter=My'

   ArgumentCountError

  Too few arguments to function PHPUnit\Runner\TestSuiteLoader::{closure}(), 0 passed in ~/Desktop/my-org/my-app/vendor/pestphp/pest/src/PendingCalls/DescribeCall.php on line 56 and exactly 1 expected

  at tests/Unit/MyTest.php:9
      5▕ function getRequestMethod(): ?string {
      6▕   return $_SERVER['REQUEST_METHOD'];
      7▕ }
      8▕
[...]

Issue #1138, opened 2024-04-17
[Bug]: When nesting describers and the describer parent doesn't have test it will skip the beforeEach


hjJunior:
### What Happened

Giving the following structure with nested blocks:

```
beforeEach(fn () => dump('root block called'));

describe('#block1', function () {
  beforeEach(fn() => dump('#block1'));

  describe('#block1->subBlock#1', function () {
    beforeEach(fn() => dump('#block1->subBlock#1'));

    it('testing', function () {});
  });
});
```

I would expect the beforeEach from `#block1` to be executed

### How to Reproduce

- Create nested blocks with describe, each one having a beforeEach
- Add any test only on the last child of describe

**Example** 

```php
describe('#block1', function () {
  beforeEach(fn() => dump('#block1'));

  describe('#block1->subBlock#1', function () {
    beforeEach(fn() => dump('#block1->subBlock#1'));

    it('testing', function () {});
  });
});
```

### Pest Version

2.34.7

### PHP Version

8.2.0

### Operation System

macOS

### Notes

The root `beforeEach` is executed even when there is no test in same level, you can reproduce it by running the first code provided on "What Happened" section, you will notice that `root block called` is printed on terminal

nunomaduro:
this is fixed on 3.x and 4.x

Issue #1023, opened 2023-11-27
[Bug]: beforeEach doesn't run for tests in nested describes


ncphillips:
### What Happened

# What I expect to happen

If I have nested describes, top-level `beforeEach` calls are run before every nested test.

# What happens

`beforeEach` calls are only run for tests described at the same level as the `beforeEach`, not nested ones.

### How to Reproduce

```php

describe("root", function () {

    beforeEach(function () {
        throw new \PHPUnit\Framework\ExpectationFailedException('root error');
    });

    it("root tests", function () {
      //
    });

    describe("child", function () {
        beforeEach(function () {
            throw new \PHPUnit\Framework\ExpectationFailedException('child error');
        });

        it('child test', function () {
            expect(0).toBe(0);
        });
    });
});
```

With the above exampel I expect both tests to fail because `root error` was thrown.

Howver, the "child test" fails because "child error" is thrown. The upper beforeEach isn't called at all when running "child test"

### Pest Version

2.25

### PHP Version

8.1.25

### Operation System

Linux

ubald:
Having the same issue. It is extremely hard to structure and setup tests properly without this.

ncphillips:
Yeah, I've basically given up on Pest because of this. IMO this is the main selling point of the describe-it syntax. Without having this feature, describe-it is no better than xUnit.

atmikev:
Can this get prioritized? Without it there is an absurd amount of copy and pasting that needs to happen when testing detailed scenarios. I don't really see the point of allowing nested describes if this isn't working.

nunomaduro:
this is fixed on 3.x and 4.x.

Issue #984, opened 2023-10-19
[Improvement]: Add support for beforeAll() in describe and nested describe blocks 


MeiKatz:
### What Happened

As visible in the title this is not a bug report, but a request for some features I think are useful in Pest.

On the one side it would be great if we could use `beforeAll()` not only outside of `describe()`, but also inside it. Actually there is no good reason why this is not supported yet.

On the other side it would be great if we could nest `describe` blocks. I know that this is possible right now, but with some problems: what is the most obvious is that only the most inner description is used in the log.

For example I like to do something like the following:

```php
describe( Account::class, function () {
  describe('#activate()', function () {
    describe('when account is inactivate', function () {
      it('updates the related value in the database', function () {
        // ... some assertions
      });
    });
  });
});
```

Currently the output is something like this:

```
PASS  Tests\Unit\AccountTest
✓ when account is inactivate → it updates the related value in the database
```

What I like to see is something like this:
```
PASS  Tests\Unit\AccountTest
Account#activate()
  when account is inactivate
    ✓ it updates the related value in the database
    ✓ it sends some email to the user
    ✓ ...
  when account is activate
    ✓ ...
```

But what is worse is that I only the values defined in the most inner `beforeEach()` are accessible inside the tests and values are not merged in some way. Actually I am not sure why it is handled this way in the first place.

Thanks a lot for your time and effort. Pest is a great testing library and it keeps getting better. But there are some issues that are really annoying right now.

### Pest Version

2.19.1

### PHP Version

8.2.11

### Operation System

Linux

Rydgel:
That would be nice to have!

nunomaduro:
i think this solved this problem: https://github.com/pestphp/pest/pull/1295.

paulshryock:
@nunomaduro, this is still an unresolved issue.

https://github.com/pestphp/pest/pull/1295#issuecomment-2430335720 that PR doesn't fix https://github.com/pestphp/pest/issues/984:

> No, this PR doesn't address those issues.

---

**My setup**:

- MacOS 15.5
- PHP 8.4.10
- Pest 3.8.2

## Here's a different example of the same issue

**Actual output format** (note how long the last few lines are):
```
PASS  Tests\php\Unit\Users\PageTest
✓ when admin scripts are enqueued on users.php → it should enqueue a stylesheet                                                                                                                          0.08s
✓ when users.php loads with no query params → it should not try to authorize the user
✓ when users.php loads with no query params → it should not add a group to any users
[...]

PR #1295, opened 2024-10-12
Support for nested describe blocks

jshayes:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Calls parent beforeEach and afterEach functions in nested describe blocks. Also updates the test names to handle multiple nested describe blocks.

Given the following test, here is the output before and after this change
```php
describe('outer', function () {
    beforeEach(function () {
        var_dump('outer');
    });

    describe('middle', function () {
        beforeEach(function () {
            var_dump('middle');
        });

        describe('inner', function () {
            beforeEach(function () {
                var_dump('inner');
            });

            test('test', function () {
                expect(1)->toBe(1);
            });
        });
    });
});
```

**Before**
```
string(5) "inner"

   PASS  Test
  ✓ inner → test → This test printed output: string(5) "inner"
```

**After**
```
string(5) "outer"
string(6) "middle"
string(5) "inner"

   PASS  Test
  ✓ outer → middle → inner → test
```

This change could cause breaking changes, so I'm not sure what you think of this. Since this project follows semver, I suppose this should go towards version 4.

### Related:

https://github.com/pestphp/pest/issues/1023
https://github.com/pestphp/pest/issues/1138

This change was based off https://github.com/pestphp/pest/pull/1294, so it includes those changes as well, since they were required for this change.

### DIFF
--- a/src/PendingCalls/AfterEachCall.php
+++ b/src/PendingCalls/AfterEachCall.php
@@ -54,7 +54,7 @@ final class AfterEachCall
         $proxies = $this->proxies;
 
         $afterEachTestCase = ChainableClosure::boundWhen(
-            fn (): bool => is_null($describing) || $this->__describing === $describing,
+            fn (): bool => $describing === [] || in_array(end($describing), $this->__describing, true),
             ChainableClosure::bound(fn () => $proxies->chain($this), $this->closure)->bindTo($this, self::class), // @phpstan-ignore-line
         )->bindTo($this, self::class);
 
--- a/src/PendingCalls/AfterEachCall.php
+++ b/src/PendingCalls/AfterEachCall.php
@@ -6,6 +6,7 @@ namespace Pest\PendingCalls;
 
 use Closure;
 use Pest\PendingCalls\Concerns\Describable;
+use Pest\Support\Arr;
 use Pest\Support\Backtrace;
 use Pest\Support\ChainableClosure;
 use Pest\Support\HigherOrderMessageCollection;
@@ -54,7 +55,7 @@ final class AfterEachCall
         $proxies = $this->proxies;
 
         $afterEachTestCase = ChainableClosure::boundWhen(
-            fn (): bool => $describing === [] || in_array(end($describing), $this->__describing, true),
+            fn (): bool => $describing === [] || in_array(Arr::last($describing), $this->__describing, true),
             ChainableClosure::bound(fn () => $proxies->chain($this), $this->closure)->bindTo($this, self::class), // @phpstan-ignore-line
         )->bindTo($this, self::class);
 
--- a/src/PendingCalls/Concerns/Describable.php
+++ b/src/PendingCalls/Concerns/Describable.php
@@ -12,14 +12,14 @@ trait Describable
     /**
      * Note: this is property is not used; however, it gets added automatically by rector php.
      *
-     * @var string[]
+     * @var array<int, string>
      */
     public array $__describing;
 
     /**
      * The describing of the test case.
      *
-     * @var string[]
+     * @var array<int, string>
      */
     public array $describing = [];
 }
--- a/src/PendingCalls/Concerns/Describable.php
+++ b/src/PendingCalls/Concerns/Describable.php
@@ -11,11 +11,15 @@ trait Describable
 {
     /**
      * Note: this is property is not used; however, it gets added automatically by rector php.
+     *
+     * @var string[]
      */
-    public string $__describing;
+    public array $__describing;
 
     /**
      * The describing of the test case.
+     *
+     * @var string[]
      */
-    public ?string $describing = null;
+    public array $describing = [];
 }
--- a/tests/.pest/snapshots/Visual/Todo/todo_in_parallel.snap
+++ b/tests/.pest/snapshots/Visual/Todo/todo_in_parallel.snap
@@ -15,7 +15,7 @@
   ↓ todo on describe → should not fail
   ↓ todo on describe → should run
 
-   TODO  Tests\Features\Todo - 7 todos
+   TODO  Tests\Features\Todo - 11 todos
   ↓ something todo later
   ↓ something todo later chained
   ↓ something todo later chained and with function body
@@ -24,6 +24,10 @@
   ↓ it may have an associated PR #1
   ↓ it may have an associated note
   // a note
+  ↓ todo on describe → todo block → nested inside todo block → it should not execute
+  ↓ todo on describe → todo block → it should not execute
+  ↓ todo on beforeEach → todo block → nested inside todo block → it should not execute
+  ↓ todo on beforeEach → todo block → it should not execute
 
    PASS  Tests\CustomTestCase\ChildTest
   ✓ override method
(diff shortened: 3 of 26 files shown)


nunomaduro:
@jshayes honestly, looks solid. do you mind to add a few more tests? specially around stuff like chaining methods after describe blocks, hooks, etc...

jshayes:
@nunomaduro I added some more test. Let me know if there are any cases you think I missed that you would like tested.

zormandi-byborg:
I would **LOVE** this. Having no support for nested `describe` and `beforeEach` blocks was the main blocker why we hesitated to switch to Pest. This would finally remove the last roadblock. ❤️

src/PendingCalls/AfterEachCall.php
```diff
@@ -54,7 +54,7 @@ public function __destruct()
         $proxies = $this->proxies;
 
         $afterEachTestCase = ChainableClosure::boundWhen(
-            fn (): bool => is_null($describing) || $this->__describing === $describing,
+            fn (): bool => $describing === [] || in_array(end($describing), $this->__describing, true),
```
nunomaduro: > `end() advances array's internal pointer to the last element, and returns its value.`

Because this function actually mutates the given value, i rather not use it.

src/PendingCalls/Concerns/Describable.php
```diff
@@ -11,11 +11,15 @@ trait Describable
 {
     /**
      * Note: this is property is not used; however, it gets added automatically by rector php.
+     *
+     * @var string[]
```
nunomaduro: We use the format `array<int, string>` instead.

tests/.pest/snapshots/Visual/Todo/todo_in_parallel.snap
```diff
@@ -24,6 +24,10 @@
   ↓ it may have an associated PR #1
   ↓ it may have an associated note
   // a note
+  ↓ todo on describe → todo block → nested inside todo block → it should not execute
```
nunomaduro, line 27: Can you add some tests too in the middle of the test file?

tests/.pest/snapshots/Visual/Todo/todos.snap
```diff
@@ -24,6 +24,10 @@
   ↓ it may have an associated PR #1
   ↓ it may have an associated note
   // a note
+  ↓ todo on describe → todo block → nested inside todo block → it should not execute
```
nunomaduro, line 27: Can you add some tests too in the middle of the test file?

tests/.pest/snapshots/Visual/Todo/todos_in_parallel.snap
```diff
@@ -24,6 +24,10 @@
   ↓ it may have an associated PR #1
   ↓ it may have an associated note
   // a note
+  ↓ todo on describe → todo block → nested inside todo block → it should not execute
```
nunomaduro, line 27: Can you add some tests too in the middle of the test file?

tests/.snapshots/success.txt
```diff
@@ -178,6 +183,10 @@
   ✓ it may be used with high order with dataset "informal"
   ✓ it may be used with high order even when bound with dataset "formal"
   ✓ it may be used with high order even when bound with dataset "informal"
+  ✓ with on nested describe → nested → before inner describe block with (1)
```
nunomaduro, line 186: Can you add some tests too in the middle of the test file?

tests/.snapshots/success.txt
```diff
@@ -188,6 +197,12 @@
   ✓ depends run test only once
   ✓ it asserts true is true
   ✓ depends works with the correct test name
+  ✓ describe block → first in describe
```
nunomaduro, line 204: Can you add some tests too in the middle of the test file?

tests/.snapshots/success.txt
```diff
@@ -1067,9 +1083,22 @@
   ✓ nested → it may have static note and runtime note
   // This is before each static note
   // This is describe static note
+  // This is before each describe static note
   // This is a static note within describe
   // This is before each runtime note
+  // This is before each describe runtime note
   // This is a runtime note within describe
+  ✓ nested → describe nested within describe → it may have a static note and runtime note
```
nunomaduro, line 1096: Can you add some tests too in the middle of the test file?

tests/.snapshots/success.txt
```diff
@@ -1255,6 +1301,12 @@
   - it skips when skip after assertion
   - it can use something in the test case as a condition → This test was skipped
   - it can user higher order callables and skip
+  - skip on describe → skipped tests → nested inside skipped block → it should not execute
```
nunomaduro, line 1309: Can you add some tests too in the middle of the test file?

tests/.snapshots/success.txt
```diff
@@ -1285,6 +1337,12 @@
   ↓ it may have an associated PR #1
   ↓ it may have an associated note
   // a note
+  ↓ todo on describe → todo block → nested inside todo block → it should not execute
```
nunomaduro, line 1347: Can you add some tests too in the middle of the test file?

nunomaduro (changes requested):

tests/.snapshots/success.txt
jshayes: There is a test after this one: `multiple notes`, let me know if you want other tests here as well

jshayes:
I added some more tests to the end of the test files, after the describe blocks. Is this what you had in mind?

jshayes:
Also, I noticed another issue with describe blocks that I have a separate PR for, but I think the issue is a bigger deal due do to nested describe blocks in this PR.

For instance, the following test
```php
describe('same-name', function () {
    beforeEach(function () {
        var_dump('before each');
    });
});

describe('same-name', function () {
    test('test', function () {
        expect(true)->toBeTrue();
    });
});
```

Will produce the following output
```
string(11) "before each"

   WARN  Tests\Features\BeforeEach
  ! same-name → test → This test printed output: string(11) "before each"                                               0.01s  
```

Since it is just checking the name of the describe block, it will match the before each from one block into another block. I do have a PR where I address this issue, however I figured I'd bring it up here because this could have some implications on nested describe blocks until the issue is resolved.

For instance, the following test
```php
describe('outer block 1', function () {
    describe('same-name', function () {
        beforeEach(function () {
            var_dump('before each');
        });
    });
});

describe('outer block 2', function () {
    describe('same-name', function () {
        test('test', function () {
            expect(true)->toBeTrue();
        });
    });
});

```

will produce the following output
```
string(11) "before each"

   WARN  Tests\Features\BeforeEach
  ! same-name → test → This test printed output: string(11) "before each"                                               0.01s  
```

This is the current behaviour, before my changes here, but I think this is a fairly common use case to have describe blocks with the same name when they are nested in different describe blocks.

Anyway, let me know if you are okay with this being a separate PR, or if you want me to add those changes here.

hafezdivandari:
Does this also fix #1010? And #984?

jshayes:
> Does this also fix #1010? And #984?

No, this PR doesn't address those issues. For #1010, the calling `with` works on a describe block, however, the value is not passed in as an argument to the describe callback. It is only passed in to the test callback

PR #1300, opened 2024-10-17
`toBeDeepOf()`

faissaloux:
Hello 👋🏻 

This PR introduces a new expectation: **toBeDeepOf()**, which makes sure that an array is deep of specific value.

### Examples
```php
$array = [1, 2, 3, 4];

expect($array)->toBeDeepOf(0); // true
expect($array)->toBeDeepOf(1); // false
```

```php
$array = [1, 2, 3 => [1, 2], 4];

expect($array)->toBeDeepOf(0); // false
expect($array)->toBeDeepOf(1); // true
```

### DIFF
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -27,6 +27,8 @@ use ReflectionNamedType;
 use Throwable;
 use Traversable;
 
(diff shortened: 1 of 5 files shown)


nunomaduro:
no plans for this.

PR #1304, opened 2024-10-22
Fix an issue with describe blocks with matching names

jshayes:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Fixes an issue described here: https://github.com/pestphp/pest/pull/1295#issuecomment-2409008790

The issue is caused by using the describe block name to determine if a `beforeEach`, or similar hook, should be applied. This PR wraps the description in an object so that the strict equal check will only pass when the hook is applied to the same describe block, not just a describe block with the same name.

### Related:

https://github.com/pestphp/pest/pull/1295#issuecomment-2409008790


src/PendingCalls/DescribeCall.php
```diff
@@ -25,22 +26,27 @@ final class DescribeCall
...
+    private readonly Description $description;
+
     /**
      * Creates a new Pending Call.
      */
     public function __construct(
         public readonly TestSuite $testSuite,
         public readonly string $filename,
-        public readonly string $description,
+        string $description,
```
nunomaduro: can you accept an `public readonly Description $description` here?

nunomaduro (changes requested):

JeffBeltran:
ran into something similar this this weekend after upgrading to v3 and started using nested describe blocks. would love to see this merged in

zillemarco:
I ran into exactly this today 😅 Is this planned to be picked up/merged any time soon @nunomaduro ? 🙂

PR #1315, opened 2024-11-17
feat: add support for anonymous tests

andaniel05:
### What:

- [ ] Bug Fix
- [X] New Feature

### Description:

Sometimes we start thinking about the test code rather than its title.

Given these situations, I think it makes sense to allow specifying the closure from the first argument and in that case, take as title the relative path of the test file and the line where the closure is defined. In those cases, it would be necessary to be able to filter this type of tests.

```php
<?php

test(function () {
    expect(true)->toBe(true);
});
```

Output result:

imagen

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -1044,6 +1044,29 @@ final class Expectation
     }
 
     /**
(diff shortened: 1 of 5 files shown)


nunomaduro:

PR #1319, opened 2024-11-20
Allow custom arch expectations

olivernybroe:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:
Allow adding custom arch expectations into pest.

A custom arch expectation needs to have typehinted the return type to the `ArchExpectation` for it to work with this implementation. 

An example could be the following added as a custom expectation
```php
expect()->extend('toHaveMethodsDocumentedMatching', function (string $pattern): ArchExpectation {
    return Targeted::make(
        $this, // @phpstan-ignore-line
        fn (ObjectDescription $object): bool => isset($object->reflectionClass) === false
            || (
                array_filter(
                Reflection::getMethodsFromReflectionClass($object->reflectionClass),
                fn(ReflectionMethod $method): bool => (enum_exists($object->name) === false || in_array($method->name, ['from', 'tryFrom', 'cases'], true) === false)
                    && realpath($method->getFileName() ?: '/') === realpath($object->path)
                    && !Str::isMatch($pattern, $method->getDocComment()),
                ) === []
            ),
        'to have methods with documentation / annotations matching pattern',
        FileLineFinder::where(fn (string $line): bool => str_contains($line, 'class'))
    );
});
```

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -330,7 +330,7 @@ final class Expectation
      * @param  array<int, mixed>  $parameters
      * @return Expectation<TValue>|HigherOrderExpectation<Expectation<TValue>, TValue>
      */
-    public function __call(string $method, array $parameters): Expectation|HigherOrderExpectation|PendingArchExpectation
+    public function __call(string $method, array $parameters): Expectation|HigherOrderExpectation|PendingArchExpectation|ArchExpectation
     {
         if (! self::hasMethod($method)) {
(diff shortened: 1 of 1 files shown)


Lukasss93:
I would love to have this feature in pest

fabio-ivona (approved):

Lukasss93:
Can you add this feature to the documentation?

Issue #1282, opened 2024-10-01
[Bug]: beforeAll from uses not triggered since 3.0.4


laylatichy:
### What Happened

hook using

uses()->beforeAll

not triggered since v3.0.4

### How to Reproduce

https://phpsandbox.io/n/aquamarine-boba-fett-lh7iw#tests/Pest.php

open terminal on the right and run vendor/bin/pest

![image](https://github.com/user-attachments/assets/55257c2c-be90-41da-8ecd-69019413e0d6)

exception is not thrown

meanwhile beforeEach runs fine

![image](https://github.com/user-attachments/assets/130236cc-9427-404b-aa88-1f443a25ae92)

### Sample Repository

https://phpsandbox.io/n/aquamarine-boba-fett-lh7iw#tests/Pest.php

### Pest Version

3.0.4

### PHP Version

8.3

### Operation System

Linux

FrancisMawn:
We noticed the same breaking change when we updated from v2 to v3.

Is this intended? Is there a new way of handling this?

laylatichy:
@FrancisMawn I was being annoyed a bit that I had to pin pest to 3.0.3 for it to work so I created pr with fix for it, hopefully it will get merged so we can enjoy again ;)

widoz:
If you don't want to stick with an older version, you can extend Pest and get back to phpunit approach by defining ` setUpBeforeClass`. I did that in a project, mainly, in your `Pest.php` file you can have something like 

```php
pest()->extends(MyTestCase::class)
    ->in('integration');
```

then your `MyTestCase`  would be something like 

```php

use PHPUnit\Framework\TestCase;

class MyTestCase extends TestCase
{
        public static function setUpBeforeClass(): void
        {
                 // Your setup

                 parent::setUpBeforeClass();
        }
}
```

When the case I do also add the teardown `tearDownAfterClass`  but just to have everything in the same place. 

Tested with v3.8.2

nunomaduro:
fixed on 4.x

PR #1322, opened 2024-11-24
fix: uses beforeAll not being called since 3.0.4

laylatichy:
### What:

- [x] Bug Fix

### Description:

fixes regresion from 3.0.4 that beforeAll hook from pest()/uses() is not called
moved TestCaseMethodFactory setUp method call back to testable constructor because calling it in setUp was already to late

### Related:

#1282


Ehsan-hulle24 (approved):

nunomaduro:
this is fixed on 4.x

PR #1326, opened 2024-12-12
feat: add toBeUlid() expectation

lalcebo:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

This PR introduces a new expectation: `toBeUlid()`, which makes sure that `$value` is a valid ULID.

### DIFF
--- a/tests/Features/Expect/toBeUlid.php
+++ b/tests/Features/Expect/toBeUlid.php
@@ -1,0 +1,24 @@
+<?php
+
+use Pest\Exceptions\InvalidExpectationValue;
+use PHPUnit\Framework\ExpectationFailedException;
+
+test('failures with wrong type', function () {
+    expect([])->toBeUlid();
+})->throws(InvalidExpectationValue::class, 'Invalid expectation value type. Expected [string].');
+
+test('pass', function () {
+    expect('01JEWKD14JZJC6GKVQG8CAF93G')->toBeUlid();
+});
+
+test('failures', function () {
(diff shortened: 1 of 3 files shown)


tests/Features/Expect/toBeUlid.php
```diff
@@ -0,0 +1,24 @@
...
+
+use Pest\Exceptions\InvalidExpectationValue;
+use PHPUnit\Framework\ExpectationFailedException;
+
+test('failures with wrong type', function () {
+    expect([])->toBeUlid();
+})->throws(InvalidExpectationValue::class, 'Invalid expectation value type. Expected [string].');
+
+test('pass', function () {
+    expect('01JEWKD14JZJC6GKVQG8CAF93G')->toBeUlid();
```
don-tre, line 11: Hi! I'm just walking around pest and I would like to say about case "00000000-0000-0000-0000-000000000001", We are using a lot this format a lot in our huge app.

lalcebo: @don-tre As far as I understand GUID has no difference with UUID, so you could still use `expect('value')->toBeUuid()` https://en.wikipedia.org/wiki/Universally_unique_identifier

But for better semantics you could add your own expectation in Pest.php

```php
// Pest.php
expect()
    ->extend('toBeGuid', fn (string $message = '') => expect($this->value)
        ->toBeUuid($message));

// On your tests
expect('value')->toBeGuid()
```

nunomaduro:

PR #1345, opened 2025-01-25
Add quiet-coverage option

carlos-granados:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

Now that the `--quiet-coverage` option has been added to Collision, this PR adds the same functionality to Pest. If this command line flag is set, then files where code coverage is 100% will not be shown. Notice that this is different from the `--compact` flag, which also hides these files with full coverage but which also affects other parts of the output

### Related:

https://github.com/nunomaduro/collision/pull/314


nunomaduro:
this is done.

PR #1378, opened 2025-03-28
[Feature] Skip tests based on current environment

ExeQue:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

Skips tests based on the output of `Environment::name()` which is either `local` or `ci` depending on whether the `--ci` option is used.

This skip is handy if a test case shouldn't be executed on CI (or locally) due to platform, memory and/or time restrictions.
This could be an integration test utilizing an external API which cannot be mocked properly that can't/shouldn't be called from CI.

It could be handled manually, but since the `Environment` class is marked as `@internal` code inspections locally would mark it as an issue.

Style inspiration taken from `skipOnOs` aliases.

Added features:
 - `skipOnLocal` skips test on `local`
 - `skipOnCi` skips test on `ci`
 - `onlyOnLocal` skips test if not on `local`
 - `onlyOnCi` skips test if not on `ci`

### DIFF
--- a/src/PendingCalls/TestCall.php
+++ b/src/PendingCalls/TestCall.php
@@ -361,7 +361,7 @@ final class TestCall // @phpstan-ignore-line
      */
     private function skipOnEnvironment(string $environment): self
     {
-        if(Environment::name() === $environment) {
+        if (Environment::name() === $environment) {
             return $this->skip(sprintf('This test is skipped on %s.', $environment));
         }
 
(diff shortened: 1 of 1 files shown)


ExeQue:
@nunomaduro Is this something that can get merged at some point soon?

I'm in need of this feature on a few projects and I'd rather not keep maintaining a fork just for this.

nunomaduro:
this is done on 4.x

PR #1391, opened 2025-04-20
[WIP] Feat: PHPUnit 12

MrPunyapal:


src/Concerns/Testable.php
```diff
@@ -101,27 +101,6 @@ trait Testable
      */
     private array $__snapshotChanges = [];
 
-    /**
-     * Creates a new Test Case instance.
-     */
-    public function __construct(string $name)
```
MrPunyapal, line 107: had to remove this as the parent constructor is now final: https://github.com/sebastianbergmann/phpunit/commit/3263f4c62d4af44777ff1c81d093ee88eafccdad

Issue #1384, opened 2025-04-07
[Bug]: Pest internally throws  `$__latestDescription must not be accessed before initialization` when data provider returns empty array, instead of the expected `InvalidDataProviderException`


maples3:
### What Happened

Here is a minimal reproduction test case:
```php
test('example', function () {
    expect(true)->toBeTrue();
})->with(function () {
    return [];
});
```

It _looks_ like this is supposed to be throwing an `InvalidDataProviderException` exception alerting me that my data provider appears incorrect (`src/Metadata/Api/DataProvider.php`):
![Image](https://github.com/user-attachments/assets/dfd5bfb9-6ca1-45d5-9cda-d2de1451fe5f)

However, it is actually throwing a PHP error attempting to access the `$__latestDescription` variable:
![Image](https://github.com/user-attachments/assets/ff495904-cf32-4ffa-9bd2-3a926dfcedec)

### How to Reproduce

1. Create a brand new Composer project
2. `./vendor/bin/pest --init`
3. Add the above minimal reproduction test case to the automatically-generated example unit test.

### Pest Version

3.8.1

### PHP Version

8.3.19

### Operation System

Linux

### Notes

As a side note, it would be helpful if there was a mechanism to indicate "I expect no datasets here".

I have a different open issue (https://github.com/pestphp/pest/issues/1369) regarding filtering of test, and my workaround is to return an empty dataset array when a CLI filter is provided.  This workaround no longer works due to the `InvalidDataProviderException`, and it took a significant amount of time to realize that the `$__latestDescription must not be accessed before initialization` error was generated by Pest due to something going wrong when handling the `InvalidDataProviderException`

nunomaduro:
This issue is now fixed on Pest 4.

PR #1397, opened 2025-04-26
fix: initialize $__latestDescription to prevent fatal error with empty datasets

jeancodogno:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

This fixes a fatal error caused when accessing the static property `$__latestDescription` before it was initialized. The error occurs in specific situations where a conditional dataset returns an empty array.

### Related:

#1384


nunomaduro:
whatever this is, this is fixed on pest 4.x

PR #1426, opened 2025-07-31
[3.x] Throw a fatal exception if the tests directory does not exist

joelbutcher:
Duplicate of #1424 but for 3.x


nunomaduro:
no plans todo this for 3.x

PR #1430, opened 2025-08-07
[3.x] fix: respect --colors=never option in parallel mode

### LINKED ISSUE #1258: [Bug]: `--parallel` ignores or changes other commandline arguments
### What Happened

There appears to be some inconsistencies when using `--parallel`: 

- the `--colors=never` option is being disregarded
- the `--coverage-text` is disregarded
  - Note: without an added filename, this should output to the standard output. This seems broken in the non-parallel setup
  - When passing a filename like `--coverage-text=output.txt`, it DOES seem to write the correct output format in this file

### How to Reproduce

This outputs WITH colors (incorrect) and WITH coverage text (correct):
```sh
 vendor/bin/pest --parallel --colors=never --coverage-text
```

This outputs WITHOUT colors (correct) and WITHOUT coverage text (incorrect):
```sh
 vendor/bin/pest --colors=never --coverage-text
```

See screenshot:

afbeelding

The code coverage report should always work and have the same output in both commands, being in stdout:
```txt
Code Coverage Report:      
  2024-09-16 18:47:59      
                           
 Summary:                  
  Classes:  0.00% (0/419)  
  Methods:  0.00% (0/1734) 
  Lines:    0.00% (0/12906)
```

### Sample Repository

_No response_

### Pest Version

2.35.1 (Same issue in 3.x)

### PHP Version

8.2

### Operation System

Linux

### Notes

_No response_

snellingio:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

See the 4.x PR #1431

Fixes issue where `--colors=never` option was being disregarded when using `--parallel` flag.

**Problem:**
When running tests with both `--parallel` and `--colors=never` options, ANSI color codes were still being output, inconsistent with non-parallel mode behavior.

**Root Cause:**
The issue occurred in three places:

1. **Worker processes** (`bin/worker.php`): Were hardcoded to use decorated output (`true`)
2. **Main parallel process** (`src/Plugins/Parallel.php`): Was not passing decoration setting to `CleanConsoleOutput`  
3. **Progress printer** (`CompactPrinter`): Was hardcoded to use decorated output

**Solution:**
- Modified worker processes to respect the `--colors` parameter from command line arguments
- Updated main parallel process to check `--colors` option and pass decoration setting to output
- Added `CompactPrinter::create()` method to accept decoration parameter and updated `ResultPrinter` to use it

**Testing:**
- Added test case to verify `--colors=never` works correctly in parallel mode
- Verified existing functionality still works
- Manual testing confirms ANSI codes are properly suppressed when `--colors=never` is used with `--parallel`

```bash
# This should show NO ANSI color codes
vendor/bin/pest --parallel --colors=never

# This should show ANSI color codes (default behavior)  
vendor/bin/pest --parallel --colors=always
```

### Related:

Closes #1258

### DIFF
--- a/bin/worker.php
+++ b/bin/worker.php
@@ -22,7 +22,8 @@ $bootPest = (static function (): void {
 
     $input = new ArgvInput;
 
-    $output = new ConsoleOutput(OutputInterface::VERBOSITY_NORMAL, true);
+    $isDecorated = $workerArgv->getParameterOption('--colors', 'always') !== 'never';
+    $output = new ConsoleOutput(OutputInterface::VERBOSITY_NORMAL, $isDecorated);
 
     Kernel::boot($testSuite, $input, $output);
 });
--- a/src/Plugins/Parallel.php
+++ b/src/Plugins/Parallel.php
@@ -16,6 +16,7 @@ use Pest\TestSuite;
 use Stringable;
 use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Console\Output\ConsoleOutput;
 
 use function Pest\version;
 
@@ -127,7 +128,10 @@ final class Parallel implements HandlesArguments
             $arguments
         );
 
-        $exitCode = $this->paratestCommand()->run(new ArgvInput($filteredArguments), new CleanConsoleOutput);
+        $input = new ArgvInput($filteredArguments);
+        $isDecorated = $input->getParameterOption('--colors', 'always') !== 'never';
+        $output = new CleanConsoleOutput(ConsoleOutput::VERBOSITY_NORMAL, $isDecorated);
+        $exitCode = $this->paratestCommand()->run($input, $output);
 
         return CallsAddsOutput::execute($exitCode);
     }
--- a/src/Plugins/Parallel/Support/CompactPrinter.php
+++ b/src/Plugins/Parallel/Support/CompactPrinter.php
@@ -64,10 +64,18 @@ final class CompactPrinter
      */
     public static function default(): self
     {
+        return self::create(true);
+    }
+
+    /**
+     * Creates a new instance of the Compact Printer with specific decoration setting.
+     */
+    public static function create(bool $decorated): self
+    {
         return new self(
             terminal(),
-            new ConsoleOutput(decorated: true),
-            new Style(new ConsoleOutput(decorated: true)),
+            new ConsoleOutput(decorated: $decorated),
+            new Style(new ConsoleOutput(decorated: $decorated)),
             terminal()->width() - 4,
         );
     }
--- a/tests/.snapshots/success.txt
+++ b/tests/.snapshots/success.txt
@@ -1,10 +1,10 @@
 
-   FAIL  Tests\Arch
+   PASS  Tests\Arch
   ✓ preset → php → ignoring ['Pest\Expectation', 'debug_backtrace', 'var_export', …]
   ✓ preset → strict → ignoring ['usleep']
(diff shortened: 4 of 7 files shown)


bin/worker.php
```diff
@@ -22,7 +22,8 @@
 
     $input = new ArgvInput;
 
-    $output = new ConsoleOutput(OutputInterface::VERBOSITY_NORMAL, true);
+    $isDecorated = $workerArgv->getParameterOption('--colors', 'always') !== 'never';
+    $output = new ConsoleOutput(OutputInterface::VERBOSITY_NORMAL, $isDecorated);
```
snellingio, line 26: Worker processes were hardcoded to use decorated output (`true`), causing `--colors=never` to be ignored in parallel mode. Now respects the `--colors` parameter from command line arguments.

src/Plugins/Parallel.php
```diff
@@ -127,7 +128,10 @@ private function runTestSuiteInParallel(array $arguments): int
             $arguments
         );
 
-        $exitCode = $this->paratestCommand()->run(new ArgvInput($filteredArguments), new CleanConsoleOutput);
+        $input = new ArgvInput($filteredArguments);
+        $isDecorated = $input->getParameterOption('--colors', 'always') !== 'never';
+        $output = new CleanConsoleOutput(ConsoleOutput::VERBOSITY_NORMAL, $isDecorated);
+        $exitCode = $this->paratestCommand()->run($input, $output);
```
snellingio, line 134: Main parallel process wasn't passing decoration setting to CleanConsoleOutput. Now checks `--colors` option and passes the decoration setting.

src/Plugins/Parallel/Support/CompactPrinter.php
```diff
@@ -63,11 +63,19 @@ public function __construct(
...
+     * Creates a new instance of the Compact Printer with specific decoration setting.
+     */
+    public static function create(bool $decorated): self
     {
         return new self(
             terminal(),
-            new ConsoleOutput(decorated: true),
-            new Style(new ConsoleOutput(decorated: true)),
+            new ConsoleOutput(decorated: $decorated),
+            new Style(new ConsoleOutput(decorated: $decorated)),
```
snellingio, line 78: Added `create()` method to accept decoration parameter. Previously, CompactPrinter was hardcoded to use decorated output, ignoring `--colors=never` in parallel mode.

tests/Arch.php
```diff
@@ -33,6 +33,7 @@
         'dd',
         'dump',
         'expect',
+        'ray',
```
snellingio, line 36: This actually stumps me a bit -- because everyone else's pipelines pass. 

Had to have Claude help me, and they pointed out that the issue seems to be that `ray` is already used here: https://github.com/pestphp/pest/blob/3.x/src/Expectation.php#L176

tests/.snapshots/success.txt
```diff
@@ -1698,4 +1699,4 @@
    WARN  Tests\Visual\Version
   - visual snapshot of help command output
 
-  Tests:    2 deprecated, 4 warnings, 5 incomplete, 2 notices, 38 todos, 33 skipped, 1144 passed (2736 assertions)
\ No newline at end of file
+  Tests:    2 deprecated, 4 warnings, 5 incomplete, 2 notices, 38 todos, 33 skipped, 1145 passed (2739 assertions)
```
snellingio, line 1702: Unsure if I'm supposed to update the snapshot or not.

snellingio:
Closing these out since there isn't much interest in fixing.

PR #1431, opened 2025-08-07
[4.x] fix: respect --colors=never option in parallel mode

### LINKED ISSUE #1258: [Bug]: `--parallel` ignores or changes other commandline arguments

snellingio:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

See the 3.x PR #1430 

Fixes issue where `--colors=never` option was being disregarded when using `--parallel` flag.

**Problem:**
When running tests with both `--parallel` and `--colors=never` options, ANSI color codes were still being output, inconsistent with non-parallel mode behavior.

**Root Cause:**
The issue occurred in three places:

1. **Worker processes** (`bin/worker.php`): Were hardcoded to use decorated output (`true`)
2. **Main parallel process** (`src/Plugins/Parallel.php`): Was not passing decoration setting to `CleanConsoleOutput`  
3. **Progress printer** (`CompactPrinter`): Was hardcoded to use decorated output

**Solution:**
- Modified worker processes to respect the `--colors` parameter from command line arguments
- Updated main parallel process to check `--colors` option and pass decoration setting to output
- Added `CompactPrinter::create()` method to accept decoration parameter and updated `ResultPrinter` to use it

**Testing:**
- Added test case to verify `--colors=never` works correctly in parallel mode
- Verified existing functionality still works
- Manual testing confirms ANSI codes are properly suppressed when `--colors=never` is used with `--parallel`

```bash
# This should show NO ANSI color codes
vendor/bin/pest --parallel --colors=never

# This should show ANSI color codes (default behavior)  
vendor/bin/pest --parallel --colors=always
```

### Related:

Closes #1258

### DIFF
--- a/bin/worker.php
+++ b/bin/worker.php
@@ -22,7 +22,8 @@ $bootPest = (static function (): void {
 
     $input = new ArgvInput;
 
-    $output = new ConsoleOutput(OutputInterface::VERBOSITY_NORMAL, true);
+    $isDecorated = $workerArgv->getParameterOption('--colors', 'always') !== 'never';
+    $output = new ConsoleOutput(OutputInterface::VERBOSITY_NORMAL, $isDecorated);
 
     Kernel::boot($testSuite, $input, $output);
 });
--- a/src/Plugins/Parallel.php
+++ b/src/Plugins/Parallel.php
@@ -16,6 +16,7 @@ use Pest\TestSuite;
 use Stringable;
 use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Console\Output\ConsoleOutput;
 
 use function Pest\version;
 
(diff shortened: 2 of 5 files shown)


snellingio:
Closing these out since there isn't much interest in fixing.

PR #1455, opened 2025-08-22
feat: add toBeCasedCorrectly arch test assertion

SimonBroekaert:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

This PR adds a new arch test assertion that compares the file path and namespace casing with each other to detect casing errors.

As most developers work on Mac OS or Windows, which by default have case insensitive file systems, these casing spelling errors might not be caught until deployed on linux environments. 

This feature helps solve this issue by comparing both namespace and file path with each other. 

### Important:

This PR depends on the following PR in the `pest-plugin-arch` plugin: https://github.com/pestphp/pest-plugin-arch/pull/27

### Example:

File path: `app/Providers/AppServiceProvider.php`

```php
<?php

namespace App\providers

class AppServiceProvider {}
```

The wrongly cased namespace part 'providers' will not cause a 'Class not found' exception on Mac OS, but will on linux.

With this feature, we can add the following test:

```php
arch()
    ->expect('App')
    ->toBeCasedCorrectly();
```

This will catch the miss-cased namespaces / file paths

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -18,6 +18,7 @@ use Pest\Arch\Expectations\ToOnlyUse;
 use Pest\Arch\Expectations\ToUse;
 use Pest\Arch\Expectations\ToUseNothing;
 use Pest\Arch\PendingArchExpectation;
+use Pest\Arch\Support\Composer;
 use Pest\Arch\Support\FileLineFinder;
 use Pest\Concerns\Extendable;
 use Pest\Concerns\Pipeable;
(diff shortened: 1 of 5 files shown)


nunomaduro:
i like it.. can you add a test please?

SimonBroekaert:
> i like it.. can you add a test please?

@nunomaduro , Thank you for considering this feature.

I've added some tests.

Keep in mind that they **will fail** until the https://github.com/pestphp/pest-plugin-arch/pull/27 PR has been merged. (See the important note in the description).

To get the namespaces working for the tests, I've also added a new commit to that PR so that the `Composer::userNamespaces()` can find the namespace of the Fixture files.

PR #1473, opened 2025-08-25
[4.x] Fixes `--processes` argument combined with sharding

### LINKED ISSUE #1454: [Bug]: Pest 4: Unknown option "--processes" when Using Sharding
### What Happened

```bash
./vendor/bin/pest --processes=12 --shard=1/4 --parallel
   Symfony\Component\Process\Exception\ProcessFailedException

  The command "'php' './vendor/bin/pest' '--processes=12' '--cache-directory' '/Users/adam/Projects/app/vendor/pestphp/pest/.temp' '--no-output' '--list-tests'" failed.

Exit Code: 2(Misuse of shell builtins)

Working directory: /Users/adam/Projects/app

Output:
================

   INFO  Unknown option "--processes".



Error Output:
================

  at vendor/symfony/process/Process.php:269
    265▕      */
    266▕     public function mustRun(?callable $callback = null, array $env = []): static
    267▕     {
    268▕         if (0 !== $this->run($callback, $env)) {
  ➜ 269▕             throw new ProcessFailedException($this);
    270▕         }
    271▕
    272▕         return $this;
    273▕     }

```

`/vendor/bin/pest --processes=12 --parallel`
works normally

### How to Reproduce

run `/vendor/bin/pest --processes=12 --shard=1/4 --parallel`

### Sample Repository

_No response_

### Pest Version

4.0.0

### PHP Version

8.3.16

### Operation System

macOS

### Notes

This is happening on my local and CI.

Our CI runs faster when we have 12 be the process but sucks that shards seem to break process

adamjsturge:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Fixes #1454

When you try to use sharding since it has to see what tests, process is an illegal argument for that. I tried adding it to the array that filters out parallel, but it doesn't work because of the `=` i think. So I made a simple filter to get it

### DIFF
--- a/src/Plugins/Shard.php
+++ b/src/Plugins/Shard.php
@@ -100,7 +100,33 @@ final class Shard implements AddsOutput, HandlesArguments
      */
     private function removeParallelArguments(array $arguments): array
     {
-        return array_filter($arguments, fn (string $argument): bool => ! in_array($argument, ['--parallel', '-p'], strict: true));
+        $filtered = [];
+        $skipNext = false;
+
+        foreach ($arguments as $argument) {
+            if ($skipNext) {
+                $skipNext = false;
+
+                continue;
+            }
+
+            if (in_array($argument, ['--parallel', '-p'], strict: true)) {
+                continue;
+            }
+
+            if (str_starts_with($argument, '--processes')) {
+                if (str_contains($argument, '=')) {
+                    continue;
+                }
+                $skipNext = true;
+
+                continue;
+            }
+
+            $filtered[] = $argument;
+        }
+
+        return $filtered;
     }
 
     /**
--- a/src/Plugins/Shard.php
+++ b/src/Plugins/Shard.php
@@ -85,7 +85,7 @@ final class Shard implements AddsOutput, HandlesArguments
     {
         $output = (new Process([
             'php',
-            ...$this->removeParallelArguments($arguments),
+            ...$this->removeParallelizationArguments($arguments),
             '--list-tests',
         ]))->mustRun()->getOutput();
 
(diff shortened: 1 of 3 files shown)


adamjsturge:
@nunomaduro are you able to review this? This has been a blocker for my organization for a little bit and really hoping we can get this fix out

src/Plugins/Shard.php
```diff
@@ -100,7 +100,33 @@ private function allTests(array $arguments): array
      */
     private function removeParallelArguments(array $arguments): array
     {
-        return array_filter($arguments, fn (string $argument): bool => ! in_array($argument, ['--parallel', '-p'], strict: true));
```
sonalidudhia, line 103: @adamjsturge Please create a new function to handle the removal of the processes argument.
Add your implementation to that new function.
Then chain the functions together to keep the code clean and maintainable.

For Example:
 
private function removeParallelArguments(array $arguments): array

    {

        return $this->removeProcessesArguments(

            $this->removeParallelFlags($arguments)

        );

    }

src/Plugins/Shard.php
```diff
@@ -100,7 +100,33 @@ private function allTests(array $arguments): array
...
+        $filtered = [];
+        $skipNext = false;
+
+        foreach ($arguments as $argument) {
```
sonalidudhia: @adamjsturge 
I’ve put together a slightly optimized version of the logic to make it more concise by using array_filter and avoiding manual array building.
This helps keep the code a bit cleaner and easier to maintain.


 




        return array_values(array_filter($arguments, function (string $argument) {

            if (str_starts_with($argument, '--processes') && str_contains($argument, '=')) {
                return false;
            }

            static $skipNext = false;
            if ($skipNext) {
                $skipNext = false;
                return false;
            }

            if ($argument === '--processes') {
                $skipNext = true;
                return false;
            }

            return true;
        }));

adamjsturge:
@sonalidudhia Thanks for your comments, hopefully you are happy with how I addressed them!

src/Plugins/Shard.php
nunomaduro: i dont think there is need for this.

sonalidudhia: @nunomaduro  You mean chaining the functions or creating separate function?

adamjsturge: @nunomaduro do you want me to revert to the original 2 commits? Or do you have any other changes you want for this?

optii:
Any news on when this will land ? We are waiting on this for our test pipelines. If there is anything I can do to help get this through, let me know.

odinns:
I just encountered this problem when I tried to combine sharding with parallel testing.

@adamjsturge Would this fix make it possible to run each shard in parallel mode?

adamjsturge:
> I just encountered this problem when I tried to combine sharding with parallel testing.
> 
> @adamjsturge Would this fix make it possible to run each shard in parallel mode?

Parallel mode and sharding should still work but when you add `processes` to sharding it breaks.

nunomaduro:
Can you rewrite this pull request in a way that you specify the arguments to be removed like:

```
$this->removeParallelArguments(['--parallel', '-p', '--processes']);
```

Or something similar? Resubmit the PR after.

Issue #1511, opened 2025-09-11
[Bug]: AwaitableWebpage::click times out when the next page is slow to load


mattford:
### What Happened

When using `->click()` or `->press()` within a browser test, if the element clicked causes a navigation and the next page takes >5s to load, the action fails (even though the element was successfully clicked)

This is possibly an issue with Playwright but I'm starting here as I'm not as familiar with that side of things.

In my example the test fails at the `click` line:

```
   FAILED  Tests\Browser\WelcomePageTest > clicking on button
  Execution context was destroyed, most likely because of a navigation A screenshot of the page has been saved to [Tests/Browser/Screenshots/clicking_on_button].

  at tests/Browser/WelcomePageTest.php:4
      1▕ <?php
      2▕ test('clicking on button', function (): void {
      3▕    visit('/')
  ➜   4▕      ->click('Click here!')
      5▕      ->assertSee('Hello, World!');
      6▕ });
```

However the screenshot produced shows the loaded "Hello, World!" page, so it appears that the test should be successful. 
I put a dump in the sendMessage method and saw that it was repeatedly sending the click command to Playwright, given the navigation already occurred these later calls likely didn't find the element to click on.

```
PHPUnit\Framework\ExpectationFailedException^ {#4035
  #message: "Timeout 1000ms exceeded."
  #code: 0
  #file: "/var/www/html/vendor/pestphp/pest-plugin-browser/src/Playwright/Client.php"
  #line: 101
  #serializableTrace: array:30 [
    0 => array:3 [
      "function" => "execute"
      "class" => "Pest\Browser\Playwright\Client"
      "type" => "->"
    ]
    1 => array:3 [
      "file" => "/var/www/html/vendor/pestphp/pest-plugin-browser/src/Playwright/Concerns/InteractsWithPlaywright.php"
      "line" => 105
      "function" => "iterator_to_array"
    ]
    2 => array:5 [
      "file" => "/var/www/html/vendor/pestphp/pest-plugin-browser/src/Playwright/Locator.php"
      "line" => 137
      "function" => "processVoidResponse"
      "class" => "Pest\Browser\Playwright\Locator"
      "type" => "->"
    ]
    3 => array:5 [
      "file" => "/var/www/html/vendor/pestphp/pest-plugin-browser/src/Api/Concerns/InteractsWithElements.php"
      "line" => 169
      "function" => "click"
      "class" => "Pest\Browser\Playwright\Locator"
      "type" => "->"
    ]
[...]

PR #1558, opened 2025-10-29
[4.x] Add --attempts cli flag and test method

ozziexsh:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

This feature is mostly needed due to the addition of the browser testing API. I would argue in normal testing scenarios you would want to design tests that aren't flakey, however with the browser a lot of stuff is out of your control. This PR adds a new `--attempts=X` cli flag as well as an `->attempts(X)` test method, which will try to run the tests up to X times, marking it as passed the first time it succeeds. If the tries exceeds X then the test is marked as failed.

Something that would be nice that I haven't added yet is some sort of output at the end of the test run which lets you know that a test was attempted a few times. I've created a class to keep track of the tests that were reattempted, but I haven't done anything with the tracked tests yet as I'm not sure the best way to add to the final test output

### Related:

- #1511

### DIFF
--- a/bin/pest
+++ b/bin/pest
@@ -117,6 +117,16 @@ use Symfony\Component\Console\Output\ConsoleOutput;
             $arguments[] = '--no-output';
             unset($_SERVER['COLLISION_PRINTER']);
         }
+
+        if (str_contains($value, '--attempts=')) {
+            unset($arguments[$key]);
+        } elseif ($value === '--attempts') {
+            unset($arguments[$key]);
+
+            if (isset($arguments[$key + 1])) {
+                unset($arguments[$key + 1]);
+            }
+        }
     }
 
     // Used when Pest is required using composer.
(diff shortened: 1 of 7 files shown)


ozziexsh:
Converting this to a draft to get feedback first. It's working with normal testing suites but as I'm testing it more with the Browser plugin it doesn't seem to work with it just yet

nunomaduro:
can you check how this method is being called on frameworks like vitest or jest?

ozziexsh:
vitest passes `retry` as an option to the test itself

https://vitest.dev/api/#test-api-reference

```ts
test('some test', {retry: 3}, () => {...});
```

jest uses `jest.retryTrimes(3)` as a method call at the top of the test file:

https://jestjs.io/docs/jest-object#jestretrytimesnumretries-options

```ts
jest.retryTimes(3);

test('some test', () => {...});
```

I went with `attempts` since `--retry` is already a cli flag used for something else.

Also I just tested it out with the browser plugin again in a more controlled environment and I was wrong, it's working fine with it. If this is something you're interested in adding would you be able to give me some pointers on printing to the final test output? It'd be nice to be warned that reattempts were used

nunomaduro:
Thank you for this pull request, i will be writing this feature on my own soon, thanks!

Issue #1573, opened 2025-11-19
[Bug]: --parallel causes "Undefined array key 'testdox-columns'" warning in worker.php


phuclh:
### What Happened

When running tests with the `--parallel` flag, Pest v4.1.3 throws an `"Undefined array key"` warning for `testdox-columns` in the worker process, causing the test run to fail with a `WorkerCrashedException`.

### How to Reproduce

1. Install Pest v4.1.3 in a Laravel 12 project
2. Run tests with parallel execution:
  `php artisan test --parallel`

  Expected Behavior:

  Tests should run in parallel without warnings or errors, similar to sequential execution.

  Actual Behavior:

  The test run fails with the following error:

```
  In WorkerCrashedException.php line 41:

    The test "PARATEST='1' TEST_TOKEN='10' UNIQUE_TEST_TOKEN='10_691e399881a5a'
  /path/to/tests/Feature/StaticCampaignTest.php" failed.

    Exit Code: 1(General error)

    Output:
    ================

       INFO  Warning: Undefined array key "testdox-columns" in
  /path/to/vendor/pestphp/pest/bin/worker.php on line 89.

  Root Cause:

  In vendor/pestphp/pest/bin/worker.php at line 89, the code attempts to access
  $getopt['testdox-columns'] without first checking if the key exists:

  $application = new ApplicationForWrapperWorker(
      $phpunitArgv,
      $getopt['progress-file'],
      $getopt['unexpected-output-file'],
      $getopt['test-result-file'],
      $getopt['result-cache-file'] ?? null,
      $getopt['teamcity-file'] ?? null,
      $getopt['testdox-file'] ?? null,
      isset($getopt['testdox-color']),
      (int) $getopt['testdox-columns'] ?? null,  // ❌ Line 89 - Bug here
  );
```

The issue is that the code tries to cast to (int) before the null coalescing operator checks if the key exists. When running in parallel mode, the testdox-columns option is not always passed to worker processes, causing PHP to emit an `"Undefined array key"` warning.

### Pest Version

4.1.3

### PHP Version

8.3.26

### Operation System

macOS

PR #1574, opened 2025-11-19
Fix testdox-columns handling in worker.php

phuclh:
### What:

  - [x] Bug Fix
  - [ ] New Feature

  ### Description:

  Fixes operator precedence bug in `bin/worker.php` line 89 that causes "Undefined array key
  'testdox-columns'" warning when running tests with `--parallel` flag.

  The type cast `(int)` has higher precedence than the null coalescing operator `??`, causing PHP
  to access the array key before checking if it exists.

  Changed from:
  ```php
  (int) $getopt['testdox-columns'] ?? null,

  To:
  isset($getopt['testdox-columns']) ? (int) $getopt['testdox-columns'] : null,
  ```

 This ensures the array key is only accessed after verifying it exists.

  Related:

  https://github.com/pestphp/pest/issues/1573

### DIFF
--- a/bin/worker.php
+++ b/bin/worker.php
@@ -86,7 +86,7 @@ $bootPest = (static function (): void {
         $getopt['teamcity-file'] ?? null,
         $getopt['testdox-file'] ?? null,
         isset($getopt['testdox-color']),
-        (int) ($getopt['testdox-columns'] ?? null),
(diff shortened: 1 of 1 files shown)


Se7en-RU:
It's already fixed in latest release 
https://github.com/pestphp/pest/releases/tag/v4.1.4

PR #1592, opened 2025-12-18
feat: add only function to mark each test in a file as only

Willem-Jaap:
## Add file-level `only()` function

Adds a new global `only()` function to mark all tests in a file to run exclusively, complementing the existing test-level `->only()` method.

### Usage

```php
<?php

only(); // Mark all tests in this file to run exclusively

test('first test', function () {
    expect(true)->toBeTrue();
});

test('second test', function () {
    expect(1 + 1)->toBe(2);
});
```

### Benefits

- Quickly run all test in a file without having to add the entire path to `pest` cmd
- Avoid chaining `->only()` to every test in a file

### Docs

Related docs pr: https://github.com/pestphp/docs/pull/345

### DIFF
--- a/src/Functions.php
+++ b/src/Functions.php
@@ -314,7 +314,7 @@ if (! function_exists('fixture')) {
 
         if ($fileRealPath === false) {
             throw new InvalidArgumentException(
-                'The fixture file [' . $file . '] does not exist.',
+                'The fixture file ['.$file.'] does not exist.',
             );
         }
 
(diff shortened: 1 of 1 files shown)


Willem-Jaap:
I've considered the following as an alternative / additional API:

```php
pest()->only();
```

Where the method is implemented on the configuration class instead of globally.

Open to suggestions!

nunomaduro:
Can you make a pull request that adds "pest()->only();"?

Willem-Jaap:
Yes, definitely:
pest: https://github.com/pestphp/pest/pull/1598 & docs: https://github.com/pestphp/docs/pull/346

PR #1601, opened 2026-01-06
[4.x] Shard filter optimization

oddvalue:
### What:

- [x] Bug Fix
- [x] New Feature

### Description:

When using shards on a sufficiently large testsuite the filter argument generated by the shard can overflow the command argument limit.

This PR introduces compression logic to reduce the filter regex length (we found up to 50% reduction in testing). 

There is also a check to test the length of the argument and throw an error if it is deemed too long.

### DIFF
--- a/src/Plugins/Shard.php
+++ b/src/Plugins/Shard.php
@@ -22,6 +22,13 @@ final class Shard implements AddsOutput, HandlesArguments
     private const string SHARD_OPTION = 'shard';
 
     /**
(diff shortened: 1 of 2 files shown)


nunomaduro:
> This PR introduces compression logic to reduce the filter regex length (we found up to 50% reduction in testing).

Can you pr without that part, and after pr it individually?

nunomaduro:
Once you update the PR just re-open it please.

PR #1624, opened 2026-02-02
fix: preserve Unicode characters in filenames for --filter matching

Vmadmax:
### What:                                                                                                                                                                                              
                                                                                                                                                                                                         
- [x] Bug Fix                                                                                                                                                                                          
- [ ] New Feature                                                                                                                                                                                      
                                                                                                                                                                                                         
### Description:                                                                                                                                                                                       
                                                                                                                                                                                                         
The `--filter` option didn't work with Unicode characters (ß, ä, ö, ü, etc.) in filenames.                                                                                                             
                                                                                                                                                                                                         
For example, `--filter=".*Straß.*"` failed to find `StraßenTest.php` because the regex in `TestCaseFactory.php` stripped all Unicode characters from the filename (`StraßenTest` became `StraenTest`). 
                                                                                                                                                                                   
                                                                                                                                                                                                         
Changed the regex to use `\p{L}\p{N}` (Unicode letters and Numbers) with the `/u` modifier to preserve Unicode characters.                                                                                              
                                                                                                                                                                                                         
### Related:                                                                                                                                                                                           
--


Vmadmax:
Hey @nunomaduro,
would be great to get your eyes on this when you have a moment.

We're working on multiple larger projects tha use Unicode filenames. It's really annoying

Issue #1344, opened 2025-01-24
[Bug]: Unable to use named parameters


victorlap:
### What Happened

When using a dataProvider with named arguments, phpunit throws a deprecation warning and will not use the argument correctly

### How to Reproduce

```php
<?php

it('can use named arguments', function (
    string $a = 'foo',
    ?string $b = null,
): void {
    expect(true)->toBeTrue();
})->with([
    'positional arguments' => ['foo', 'bar'],
    'named argument' => ['b' => 'bar'],
]);
```

Running the above code produces a deprecation warning and error similar to the following:

```
   DEPR  Tests\Unit\Test
  ! it can use named arguments → Providing invalid named argument $b for method P\Tests\Unit\Test::it_can_use_named_arguments() is deprecated and will not be supported in PHPUnit 11.0. 0.02s
  ✓ it can use named arguments with dataset "positional arguments"                                                                                                                    0.18s

  Tests:    1 deprecated, 1 passed (2 assertions)
  Duration: 0.63s
```

This should work according to the phpunit functionality, however the generated test class code doesn't include the parameters. Phpunit uses reflection to get the parameters from the function and therefore this functionality is not working. 

For completeness sake, this is the generated code:
```
namespace P\Tests\Unit;

use Pest\Repositories\DatasetsRepository as __PestDatasets;

/**
 * @testdox /var/www/html/tests/Unit/Test.php
 */
#[\AllowDynamicProperties]
final class Test extends \Tests\TestCase implements \Pest\Contracts\HasPrintableTestCaseName
{
    use \Pest\Concerns\Testable, \Pest\Concerns\Expectable, \Pest\Arch\Concerns\Architectable;

    private static $__filename = '/var/www/html/tests/Unit/Test.php';

    /**
     * @test
     * @testdox it can use named arguments
     * @dataProvider __pest_evaluable_it_can_use_named_arguments_dataset
     */
    public function __pest_evaluable_it_can_use_named_arguments()
    {
        $test = \Pest\TestSuite::getInstance()->tests->get(self::$__filename)->getMethod($this->name())->getClosure($this);

        return $this->__runTest(
            $test,
            ...func_get_args(),
        );
    }

    public static function __pest_evaluable_it_can_use_named_arguments_dataset()
    {
        return __PestDatasets::get(self::$__filename, '__pest_evaluable_it_can_use_named_arguments');
    }
}
```

### Pest Version

v2.36.0

### PHP Version

8.4.2

### Operation System

Linux

[...]

Issue #1232, opened 2024-09-11
[Bug]: Dataset with named parameters fails in Pest 3 if parameters count mismatch


lukasleitsch:
### What Happened

We use shared datasets to provide data for multiple tests. The dataset uses keys to make the data readable. Not all tests require all parameters from the datasets. So, the dataset contains more items in the array than the parameters needed by the tests.

In Pest 2, this works without any issues. In Pest 3, this setup fails.

It would be great to have the same behavior in Pest 3 as in Pest 2. I want to avoid adding unnecessary parameters to tests that are not used in the test.

In the sample repository, I provide an example.

### How to Reproduce

- Clone the sample repository
- Run `composer install`
- Run `./vendor/bin/pest` 
- The tests fail with this error

```
   FAILED  Tests\Unit\ExampleTest > dataset with less parameters count with (1, 1, 1) #2                   Error
  Unknown named parameter $parameterC
  str_replace(): Argument #3 ($subject) must be of type array|string, null given

  at vendor/filp/whoops/src/Whoops/Exception/FrameCollection.php:43
     39▕      * @return FrameCollection
     40▕      */
     41▕     public function filter($callable)
     42▕     {
  ➜  43▕         $this->frames = array_values(array_filter($this->frames, $callable));
     44▕         return $this;
     45▕     }
     46▕
     47▕     /**
 ```

- Change the pest version in the `composer.json` to `^2.0`
- Reinstall the dependencies
- Run `./vendor/bin/pest`
- The tests pass

### Sample Repository

https://github.com/lukasleitsch/pest-issue

### Pest Version

3.0.4

### PHP Version

8.3.11

### Operation System

macOS

PR #1634, opened 2026-02-11
Dataset Named Parameters

dbpolito:
### What:

- [x] Bug Fix
- [x] New Feature

### Description:

- This adds support for named params without breaking current sequential logic.
- Also makes more consistent with PHPUnit Data Provider which does support named parameters.

### Related:

https://github.com/pestphp/pest/issues/1232
https://github.com/pestphp/pest/issues/1344


dbpolito:
hey @nunomaduro , would love to hear from you what you think about this one.

tests are failing but doesn't seem to be related to this PR

PR #1647, opened 2026-02-25
POC

vistik:
<!--
- Fill in the form below correctly. This will help the Pest team to understand the PR and also work on it.
-->

### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

## Time-based test sharding                                                                                                                              
                                                                      
  Adds a `--shard-timing` option that enables time-balanced sharding when combined with the existing `--shard` flag. Instead of splitting tests by count  (which can create unbalanced CI runners when some test classes are much slower than others), this distributes test classes across shards to minimize the overall wall-clock time of a parallel run.                                                                                                               
                                                                                                                                                           
  ### How it works                                                                                                                                         

  Pass a path to a timing file alongside `--shard`:

  ```bash
  ./vendor/bin/pest --shard=1/4 --shard-timing=junit.xml
  ```
                                                                                                                                                           
  The plugin resolves timing data from two sources, in order of preference:
                                                                                                                                                           
  1. JUnit XML — parses <testsuite class="..."> time attributes (falls back to <testcase> elements if no suite-level class attributes are present)         
  2. PHPUnit result cache — reads test-results or .phpunit.result.cache from the configured or default cache directory
                                                                                                                                                           
  Tests are then distributed using the LPT (Longest Processing Time) bin-packing algorithm: classes are sorted slowest-first, and each is assigned to the shard with the lowest accumulated time so far. This greedy approach minimizes the maximum shard duration with no extra configuration.
                                                                                                                                                           
  When no timing file is available (first run, file missing), it falls back to the existing even-count split. Unknown test classes default to 1 second estimated runtime.

  #### Output

  After the suite completes, a summary line is printed:                                                                                                    
   
    Shard:    2 of 4 — 12 files ran, out of 47 (time-balanced).                                                                                            
                             
  ### The catch
  The catch is that you have to commit your phpunit.xml test report to the repo, for the plugin to read the timing data 😄

### DIFF
--- a/bin/pest
+++ b/bin/pest
@@ -112,6 +112,17 @@ use Symfony\Component\Console\Output\ConsoleOutput;
             }
         }
 
+        if (str_contains($value, '--shard-timing=')) {
+            $_SERVER['PEST_SHARD_TIMING'] = substr($value, strlen('--shard-timing='));
+            unset($arguments[$key]);
+        } elseif ($value === '--shard-timing') {
+            if (isset($arguments[$key + 1])) {
+                $_SERVER['PEST_SHARD_TIMING'] = $arguments[$key + 1];
+                unset($arguments[$key + 1]);
+            }
+            unset($arguments[$key]);
+        }
+
         if (str_contains($value, '--teamcity')) {
             unset($arguments[$key]);
             $arguments[] = '--no-output';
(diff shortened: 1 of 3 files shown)


nunomaduro:
i don't understand what this does; when you have the time, re-submit this with a pr description, etc.

PR #1651, opened 2026-03-04
fix: handle `--compact` flag in non-parallel mode

### LINKED ISSUE #1632: [Bug]: Flag --compact does not work
### What Happened

When I start the tests with --compact, it does not change anything in the output. Only with --parallel the compact output is been shown. We do have a phpunit.xml.dist file, but when I delete that file, the output is still the same. So I assume that we do not have any config issue with that.


```
# ./vendor/bin/pest --compact

   PASS  App\Tests\Backend\Domain\Comparison\Model\TariffLabelsValueTest
  ✓ iteration                                                                                                                                                                                                                                                                                                                                                               0.01s  
  ✓ merging

   PASS  Tests\Shared\Translation\LiteralTranslationMapTest
  ✓ LiteralTranslatableMap → constructor → it creates instance with all three languages                                                                                                                                                                                                                                                                                     0.03s  
  ✓ LiteralTranslatableMap → constructor → it allows empty strings
  ✓ LiteralTranslatableMap → trans → it returns German translation for de_DE locale                                                                                                                                                                                                                                                                                         0.01s  
  ✓ LiteralTranslatableMap → trans → it returns English translation for en_US locale
  ✓ LiteralTranslatableMap → trans → it returns Polish translation for pl_PL locale
  ✓ LiteralTranslatableMap → trans → it returns German translation for de locale
  ✓ LiteralTranslatableMap → trans → it returns English translation for en locale
  ✓ LiteralTranslatableMap → trans → it returns Polish translation for pl locale
  ✓ LiteralTranslatableMap → trans → it does not call translator
  ✓ LiteralTranslatableMap → trans → it handles special characters in German
  ...
  ...

``` 

### How to Reproduce

Run the command `./vendor/bin/pest --compact` in your php project with pest tests

### Sample Repository

_No response_

### Pest Version

4.3.2

### PHP Version

8.5.0

### Operation System

Linux

### Notes

_No response_

isaackaara:
## Problem

The `--compact` CLI flag was listed in the help output but only functioned when running tests in parallel mode (`--parallel --compact`). In non-parallel mode, the flag was passed through to PHPUnit, which does not recognize it, resulting in no effect on the output format.

## Root Cause

The compact output in parallel mode is handled by `Pest\Plugins\Parallel\Support\CompactPrinter`, but there was no plugin to intercept the `--compact` argument and enable compact output via `DefaultPrinter::compact(true)` in standard (non-parallel) mode.

The programmatic API (`pest()->printer()->compact()` in `Pest.php`) worked correctly because it directly calls `DefaultPrinter::compact(true)`, but the CLI flag path was missing.

## Fix

Added a `Compact` plugin that:
1. Checks for the `--compact` argument
2. Enables compact output via `DefaultPrinter::compact(true)`
3. Removes the flag from arguments so PHPUnit does not receive an unknown option

Registered the plugin in `composer.json` under `extra.pest.plugins`.

## How to Reproduce

```bash
./vendor/bin/pest --compact
# Output was unchanged from default format
```

## After Fix

```bash
./vendor/bin/pest --compact
# Output now uses compact format (single-character symbols per test)
```

Fixes #1632


nunomaduro:
there is a lot of changes here that are not related.

Issue #1462, opened 2025-08-22
[Bug]: Pest 4.0.2: missing implementations in StateGenerator.php


thomas-0816:
### What Happened

time vendor/bin/sail php vendor/bin/pest --stop-on-defect --stop-on-error --stop-on-failure --stop-on-risky --passthru-php="-derror_log=/tmp/php.log" --parallel

  ...............................................................................................................................................................

  Tests:    159 passed (841 assertions)
  Duration: 2.40s
  Parallel: 16 processes

real    0m3,742s
user    0m0,250s
sys     0m0,189s

Processing finishes successfully, but only 159 tests are executed instead of 2.7k.

So the workers are mostly crashing, due to:

```
tail -f /tmp/php.log
[22-Aug-2025 14:59:43 UTC] PHP Warning:  Undefined array key "testdox-columns" in /var/www/html/vendor/pestphp/pest/bin/worker.php on line 89
[22-Aug-2025 14:59:43 UTC] PHP Warning:  Undefined array key "testdox-columns" in /var/www/html/vendor/pestphp/pest/bin/worker.php on line 89
[22-Aug-2025 14:59:43 UTC] PHP Warning:  Undefined array key "testdox-columns" in /var/www/html/vendor/pestphp/pest/bin/worker.php on line 89
[22-Aug-2025 14:59:43 UTC] PHP Warning:  Undefined array key "testdox-columns" in /var/www/html/vendor/pestphp/pest/bin/worker.php on line 89
[22-Aug-2025 14:59:43 UTC] PHP Warning:  Undefined array key "testdox-columns" in /var/www/html/vendor/pestphp/pest/bin/worker.php on line 89
[22-Aug-2025 14:59:43 UTC] PHP Warning:  Undefined array key "testdox-columns" in /var/www/html/vendor/pestphp/pest/bin/worker.php on line 89
[22-Aug-2025 14:59:43 UTC] PHP Warning:  Undefined array key "testdox-columns" in /var/www/html/vendor/pestphp/pest/bin/worker.php on line 89
[22-Aug-2025 14:59:43 UTC] PHP Warning:  Undefined array key "testdox-columns" in /var/www/html/vendor/pestphp/pest/bin/worker.php on line 89
[22-Aug-2025 14:59:43 UTC] PHP Warning:  Undefined array key "testdox-columns" in /var/www/html/vendor/pestphp/pest/bin/worker.php on line 89
[...]

Issue #1185, opened 2024-06-29
[Bug]: invalid dataset not showing as FAILED when running with --parallel


mtbossa:
### What Happened

When I have a test that uses a dataset `->with('my dataset')`, but this dataset doesn't exist, and I run `sail test --parallel`, no `FAILED` message is shown in the output. However, if I just run `sail test`, then it'll show as FAILED, so I know something is wrong.

### How to Reproduce

Install a fresh Laravel App, add Pest, add a test that uses a non-existent dataset, run `./vendor/bin/pest --parallel`.

### Sample Repository

https://github.com/mtbossa/pest-parallel-bug-laravel

### Pest Version

2.34.8

### PHP Version

8.3.8

### Operation System

Linux

### Notes

![Screenshot_20240629_082424](https://github.com/pestphp/pest/assets/45720067/39e67402-e11d-4f6b-866e-6c0793c52510)
![Screenshot_20240629_082401](https://github.com/pestphp/pest/assets/45720067/2265aacd-3b6b-4902-b5e6-6cc63e37b753)

FrittenKeeZ:
I'm experiencing the same error - when the fault is introduced it seems like the tests are just skipped.

This is the output for the suite before introducing the error.
```
Tests:    1190 passed (5990 assertions)
Duration: 3.83s
Parallel: 11 processes
```
This is the output for the suite after introducing the error.
```
Tests:    1062 passed (5862 assertions)
Duration: 3.88s
Parallel: 11 processes
```
It looks like 128 tests have disappeared, but no failures at all.
Running the tests without `--parallel` will instantly crash with the error being outputted.

Edit: I'm using Pest 3.2

mtbossa:
> I'm experiencing the same error - when the fault is introduced it seems like the tests are just skipped.
> 
> This is the output for the suite before introducing the error.
> 
> ```
> Tests:    1190 passed (5990 assertions)
> Duration: 3.83s
> Parallel: 11 processes
> ```
> 
> This is the output for the suite after introducing the error.
> 
> ```
> Tests:    1062 passed (5862 assertions)
> Duration: 3.88s
> Parallel: 11 processes
> ```
> 
> It looks like 128 tests have disappeared, but no failures at all. Running the tests without `--parallel` will instantly crash with the error being outputted.
> 
> Edit: I'm using Pest 3.2

[...]

Issue #1463, opened 2025-08-22
[Bug]: Pest 4.0.2 stop-on-failure continues on failure instead of stop


thomas-0816:
### What Happened

time vendor/bin/sail php vendor/bin/pest --stop-on-defect --stop-on-error --stop-on-failure --stop-on-risky

```
   FAIL  stocks\tests\Feature\Repository\WareRepositoryTest

   PASS  Tests\Unit\FactoryTest
  ✓ models can be instantiated                                                                                                                       1.31s  

   PASS  Tests\Unit\MigrationTest
  ✓ database tables success created                                                                                                                  0.01s  
...

  FAILED  stocks\tests\Feature\Repository\WareRepositoryTest >                                                                                                                                                        
  The data provider P\stocks\tests\Feature\Repository\WareRepositoryTest::__pest_evaluable_testWareRepositoryGetAllWarehousesAsKeyValue_dataset specified for P\stocks\tests\Feature\Repository\WareRepositoryTest::__pest_evaluable_testWareRepositoryGetAllWarehousesAsKeyValue is invalid
A dataset with the name `warehouses` does not exist. You can create it using `dataset('warehouses', ['a', 'b']);`.
```

Test execution should stop on first failure, not after all tests are executed.

### How to Reproduce

php vendor/bin/pest --stop-on-failure

### Pest Version

4.0.2

### PHP Version

8.4.11

### Operation System

Linux

PR #1655, opened 2026-03-25
fix: nested dataset discovery and parallel invalid-dataset reporting

### LINKED ISSUE #1121: [Bug]: Cannot use Datasets in subfolders

stsepelin:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

This PR follows up on #1427 and fixes two related problems in Pest's parallel execution flow.

First, it fixes nested dataset discovery for datasets stored in nested `Datasets/` directories, addressing #1121.

Second, it fixes parallel reporting when a dataset is invalid or missing, so these cases are surfaced correctly as failures rather than being lost in the summary, addressing #1185.

The PR also adds regression coverage for both scenarios and keeps the success snapshot coverage in place with a lower memory limit for the visual snapshot test.

### Related:

- Fixes #1121
- Refs #1185
- Follow-up to #1427

### DIFF
--- a/tests/Visual/Success.php
+++ b/tests/Visual/Success.php
@@ -12,9 +12,9 @@ test('visual snapshot of test suite on success', function () {
 
     $output = function () use ($testsPath) {
         $process = (new Process(
-            ['php', 'bin/pest'],
+            ['php', '-d', 'memory_limit=512M', 'bin/pest', '--exclude-group=integration'],
             dirname($testsPath),
-            ['EXCLUDE' => 'integration', '--exclude-group' => 'integration', 'REBUILD_SNAPSHOTS' => false, 'PARATEST' => 0, 'COLLISION_PRINTER' => 'DefaultPrinter', 'COLLISION_IGNORE_DURATION' => 'true'],
+            ['EXCLUDE' => 'integration', 'REBUILD_SNAPSHOTS' => false, 'PARATEST' => 0, 'COLLISION_PRINTER' => 'DefaultPrinter', 'COLLISION_IGNORE_DURATION' => 'true'],
         ));
 
         $process->run();
--- a/tests/Visual/Success.php
+++ b/tests/Visual/Success.php
@@ -12,9 +12,9 @@ test('visual snapshot of test suite on success', function () {
 
     $output = function () use ($testsPath) {
         $process = (new Process(
-            ['php', '-d', 'memory_limit=512M', 'bin/pest', '--exclude-group=integration'],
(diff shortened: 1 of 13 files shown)


tests/Visual/Success.php
```diff
@@ -12,7 +12,7 @@
 
     $output = function () use ($testsPath) {
         $process = (new Process(
-            ['php', 'bin/pest'],
+            ['php', '-d', 'memory_limit=256M', 'bin/pest'],
```
stsepelin, line 15: I'm not quite sure about this change. After the snapshots.txt update, the default PHP config (128M memory limit)  started failing with memory limit warnings.

thomas-0816:
Thank You!
I guess the changes in src/Support/StateGenerator.php should fix https://github.com/pestphp/pest/issues/1462 and https://github.com/pestphp/pest/issues/1463 .

@stsepelin Should these methods from https://github.com/sebastianbergmann/phpunit/blob/main/src/Runner/TestResult/TestResult.php also be handled in src/Support/StateGenerator.php ?
- public function testRunnerTriggeredDeprecationEvents() // added in phpunit 12
- public function testRunnerTriggeredNoticeEvents() // added in phpunit 12
- public function testRunnerTriggeredWarningEvents() // added in phpunit 12
- public function testSuiteSkippedEvents()
- public function errors()

nunomaduro:
@thomas-0816 all of them.

stsepelin:
Thanks for the merge! 

Opened a follow-up PR to handle the remaining `PHPUnit\TestRunner\TestResult\TestResult` methods you mentioned
(`testRunnerTriggered{Deprecation,Notice,Warning}Events`, `testSuiteSkippedEvents`, `errors`): #1683.

That also covers the "missing implementations in StateGenerator.php"
part of #1462.

PR #1667, opened 2026-04-12
Upgrade nunomaduro/collision to version 11.0.0

nunomaduro:


nunomaduro:
Closing this PR as `nunomaduro/collision` v11.0.0 has not been published to Packagist yet — the highest available version is v8.x. This means the dependency constraint `^11.0.0` is unresolvable, causing all CI jobs to fail at `composer update`.

Once collision v11.0.0 is actually released, this can be reopened and the upgrade can be properly validated against the 36+ files that import from collision.

PR #1677, opened 2026-04-16
[5.x] Make time-based sharding namespace-agnostic and forward --test-directory

norbybaru:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Time-based sharding (#1671) silently fails for projects whose tests do not live under the `Tests\` namespace.

**Root cause 1 — hardcoded namespace in regex (`Shard.php:196`):**

The `allTests()` method parses `--list-tests` output using a regex that only captures FQCNs starting with `Tests\`:

```php
preg_match_all('/ - (?:P\\\\)?(Tests\\\\[^:]+)::/', $output, $matches);
```

For any other namespace (`App\Tests\`, `Acme\Suite\`, etc.), `allTests()` returns `[]`, causing:
- `--update-shards` writes an empty `shards.json`
- Time-balanced partitioning silently falls back to naive `array_chunk`
- Feature appears to work but provides no benefit

**Root cause 2 — `--test-directory` not forwarded to subprocess:**

`bin/pest` strips `--test-directory` from argv before the plugin pipeline runs (lines 42-50). `Shard::allTests()` spawns a subprocess with those stripped arguments, so the subprocess always
defaults to the `tests/` directory — even when users configure a custom test path.

### Solution

**1. Namespace-agnostic regex**

Replaced `Tests\\`-hardcoded regex with a pattern matching any valid PHP FQCN:

```php
preg_match_all('/ - (?:P\\\\)?([A-Za-z_]\w*(?:\\\\[A-Za-z_]\w*)*)::/', $output, $matches);
```

Matches sequences of valid PHP identifiers joined by `\`. Stricter than a catch-all `[^:]+` — rejects whitespace, punctuation, and malformed lines.

**2. Forward `--test-directory` to the subprocess**

Read the already-resolved test path from the `TestSuite` singleton and inject it into the subprocess command:

```php
$command = $this->buildListTestsCommand($arguments, TestSuite::getInstance()->testPath);
```


nunomaduro:
Can you fix conflicts?

norbybaru:
> Can you fix conflicts?

@nunomaduro conflicts are solved

Issue #969, opened 2023-09-22
test dependency doesn't work in different file


aaykacc:
I can set test dependencies within the same file when using PEST, but it doesn't work when the tests are in different files.

devajmeireles:
Hey, @aaykacc ! Thanks for your report.

As mentioned by you, dependencies only work for tests on the same file. We have no plans to improve this by adding the ability to use test dependencies in another file. PR is welcome to suggest this to Pest.

According to our organization, I'll close this issue. Feel free to open a new issue whenever necessary.

PR #1678, opened 2026-04-16
[4.x] Add dependsExternal (depend on tests in external test files)

Ajility97:
### What

- [ ] Bug Fix
- [x] New Feature

### Description

The existing `depends()` method does not support dependencies across test files. In larger test suites, it can be useful to reuse values returned by tests defined in other test cases.

This PR introduces a chainable `dependsExternal()` method, which allows a test to depend on tests defined in another test file. It accepts the target test file as the first argument, followed by one or more dependency test names as variadic arguments.

Example usage:

```php
test('depends on tests in an external file', function (string $first, string $second) {
    expect($first)->toBe('first');
    expect($second)->toBe('second');
})->dependsExternal('Tests\\Features\\Depends', 'first', 'second');
```

### Related:
https://github.com/pestphp/pest/issues/969
https://github.com/pestphp/pest/pull/1059
https://github.com/pestphp/pest/issues/1054

### DIFF
--- a/overrides/Runner/TestSuiteLoader.php
+++ b/overrides/Runner/TestSuiteLoader.php
@@ -179,7 +179,6 @@ final class TestSuiteLoader
         if (! class_exists($suiteClassName, false)) {
             return $this->exceptionFor($suiteClassName, $suiteClassFile);
         }
-
         // @codeCoverageIgnoreEnd
 
         if ($class->isSubclassOf(TestCase::class) && ! $class->isAbstract()) {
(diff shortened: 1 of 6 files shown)


nunomaduro:
Sorry, seems to me that, if i were to accept something like this, it would have to be on the original `depends`.

Ajility97:
@nunomaduro, just to confirm: would you be open to this feature if I reworked it to be called like below?

```
test('depends on tests in external files', function (string $first, string $second, string $third) {
    // ...
})->depends(['Tests\\Features\\Something', 'first', 'second'], ['Tests\\Features\\SomethingElse', 'third']);
```

nunomaduro:
yeah, if it not alot of new code.

Issue #126, opened 2020-07-15
Missing API the naming generation for tests without names


olivernybroe:
# Problem
As the pest plugin for IntelliJ is getting more and more love, we want to add support for tests without names.

However this raises the problem of what to show in IntelliJ.

We therefore need to know how to generate the name and for that logic to be the same in pest and in the plugin.

# Solutions
There are multiple ways to make sure this logic is the same.

1. Having an API in Pest to generate the name.
This is a nice way for pest, as how we generate can change without any further impact. The problem is then that the plugin will be rather slow as it now has to run php for getting a tests name.

2. Having a formalised name generation in text.
For this we write some text that defines how the naming generation works and then we implement that in both PHP and the IntelliJ plugin. When the generation has to change, the official definition is then updated and then we have to update both implementations.
This is a rather cumbersome process, but okay as the generation might not change that often.

3. Having a standard defined as regex.
Basically we create a regex which can generate the name of the test. This would mean one implementation as it is just a regex. However the problem here is that regex might not be able to do all the things we need to do.

Any thought or other solutions?

nunomaduro:
@olivernybroe What is this state of this?

olivernybroe:
No work have been done on it. Still blocking in regards to the pest phpstorm plugin.

I think we should formalise it in text, that's at least what I did with other pest tests.
This is the steps I documented for name generation for tests with name

1. Take the path of the PEST file from the cwd.
2. Make the first folder's first letter uppercase.
3. Remove file extension.
4. Make directory separators to namespace separators.
5. Add P as a namespace before the generated namespace.

@octoper Did we ever add that to the website?

octoper:
@olivernybroe No I don't think that we had this on the site.

nunomaduro:
@olivernybroe I may need a call so you can explain me what you need on this, sorry.

olivernybroe:
@nunomaduro sure we can make a call. But to quickly explain, basically I need steps like the ones I posted earlier, just for when a test has no name.

nunomaduro:
@olivernybroe This is still an issue? If we don't plan to address it soon, can you close it please?

olivernybroe:
Actually I think @lukeraymonddowning wanted to address it soon 😂 

However this is still something I need as it is blocking me for adding support for higher order tests in phpstorm

nunomaduro:
Closing for lack of activity.

PR #1693, opened 2026-05-06
Add the ability to extend `TestCall`

JHWelch:
### What:

- [ ] Bug Fix
- [X] New Feature

### Description:

This is another go at the change in #126. I know you closed that getting ready for V4.

This feels like the missing piece of enabling making really powerful Pest extensions that can just be chained off of existing tests.

Adds the ability to `->extend(...)` `TestCall` in the same fashion that expectations are extendable.

```php
test()->extend('fooBar', function () {
    return $this->with([
        'foo',
        'bar',
    ]);
});
```

If this is something you want to add, I can PR the docs repo wherever makes sense.

If you don't think you'll ever have an appetite for this let me know and I won't PR again.

### DIFF
--- a/src/PendingCalls/TestCall.php
+++ b/src/PendingCalls/TestCall.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Pest\PendingCalls;
 
 use Closure;
+use Pest\Concerns\Extendable;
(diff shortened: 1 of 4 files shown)


nunomaduro:
No plans for this atm, sorry!

PR #1708, opened 2026-05-16
[4.x] Add `--only-changed` option for coverage 

JonPurvis:
### What:

- [x] New Feature

### Description:

This PR adds a new `--only-changed` option to Pest's existing coverage reporting.

When enabled, coverage is calculated only for files changed in the current Git branch, rather than the entire application.

Example:

```
./vendor/bin/pest --coverage --only-changed
```

This makes coverage feedback more relevant during development, especially in larger or legacy applications where overall coverage may already be low.

For example, if a developer adds a new Laravel command and fully tests it, they would see coverage for the code introduced in their branch instead of the overall application coverage percentage.

This also makes it easier for teams to enforce coverage standards for new code in CI without being affected by existing coverage debt.


nunomaduro:
No plans for this atm, sorry!

PR #1686, opened 2026-04-23
Feat: coverage in shards

MrPunyapal:
### What:

- [x] Bug Fix
- [x] New Feature

### Description

This PR adds first-class code coverage support for sharded Pest runs.

Previously, running `--coverage` together with `--shard=X/Y` collected coverage independently for each shard, but Pest did not provide a built-in way to combine those results into a single coverage report.

With this PR, each shard stores its coverage data during execution, and a final merge step generates the same consolidated coverage report users expect from a non-sharded run.

### What's Changed

- `pest --shard=X/Y --coverage` now stores shard coverage data in `.pest/coverage/` instead of generating a coverage report for each shard.
- Added `pest --shards-coverage` to merge all shard coverage into a single report.
- The merged report supports the existing coverage options:
  - `--min`
  - `--exactly`
  - `--only-covered`
- Added `--clean` to remove temporary shard coverage files after a successful merge.
- During sharded runs, report-generating coverage flags (such as `--coverage-html` and `--coverage-clover`) are ignored, and Pest prints a reminder showing how to generate the final merged report.
- Updated the help output and added test coverage for the new workflow.

### How To Use Locally

Run each shard with coverage enabled:

```bash
pest --shard=1/2 --coverage
pest --shard=2/2 --coverage
```

Generate the merged coverage report:

```bash
pest --shards-coverage
```

Additional examples:

```bash
pest --shard=1/3 --coverage
pest --shard=2/3 --coverage
pest --shard=3/3 --coverage

pest --shards-coverage --min=80
pest --shards-coverage --exactly=91.5
pest --shards-coverage --only-covered
pest --shards-coverage --clean
```

### Notes

- Temporary shard coverage files are stored in `.pest/coverage/{shard}.cov`.
- `--clean` removes the temporary coverage files after the merged report has been generated.
- Coverage report options (such as `--coverage-html`, `--coverage-clover`, etc.) are ignored during sharded runs because reports can only be generated after all shard coverage has been merged.

### Example CI Workflow

Run each shard in parallel and upload the generated coverage directory as an artifact:

```yaml
# shard jobs
- run: php bin/pest --shard=${{ matrix.shard }}/3 --coverage

- uses: actions/upload-artifact@v4
  with:
    name: coverage-shard-${{ matrix.shard }}
    path: .pest/coverage/
```

Merge coverage in a follow-up job:

```yaml
- uses: actions/download-artifact@v4
  with:
    pattern: coverage-shard-*
    merge-multiple: true
[...]

PR #1709, opened 2026-05-16
[4.x] Fix Checks

JonPurvis:
### What:

- [x] Bug Fix

### Description:

After pest-plugin-browser was raised to `^4.3.1` ([e766825f](https://github.com/pestphp/pest/commit/e766825f5b49e11561ce87f62155d6ae2615e319)), PRs from branches not named `4.x` fail at composer update because the root package is inferred as `dev-<branch>`, which does not satisfy pest `^4.4.5`. Setting the root version unblocks CI for all PRs targeting `4.x`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,6 @@
 {
     "name": "pestphp/pest",
     "description": "The elegant PHP Testing Framework.",
-    "version": "4.7.0",
     "keywords": [
         "php",
         "framework",
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,7 @@
 {
     "name": "pestphp/pest",
     "description": "The elegant PHP Testing Framework.",
+    "version": "4.7.0",
     "keywords": [
         "php",
         "framework",
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -54,7 +54,7 @@ jobs:
 
     - name: Install Dependencies
       env:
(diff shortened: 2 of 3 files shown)


JonPurvis:
Hey @nunomaduro 👋 

Does this change make sense to you? I noticed checks failing on my earlier PR and I believe this is the fix needed.

composer.json
```diff
@@ -1,6 +1,7 @@
 {
     "name": "pestphp/pest",
     "description": "The elegant PHP Testing Framework.",
+    "version": "4.7.0",
```
MrPunyapal: With this fix, we need to update the version here with each update, no?

JonPurvis: Yeah, of course, open to other suggestions on how to fix this. But we should have running CI checks before merging PRs

MrPunyapal: I already fixed it in #1686

JonPurvis: Thought of a better way of doing this actually

JonPurvis: Ha, you're PR does what I was thinking 😆

JonPurvis: @MrPunyapal Given this PR is smaller and fixes this one thing, I think it makes sense to fast-track getting this branch merged and then you can update yours. 

What do you think?

MrPunyapal: Agree

MrPunyapal (approved):

JonPurvis:
This will also need an approval by @nunomaduro

PR #1712, opened 2026-05-18
fix: add tests without explicit group to 'default' group

### LINKED ISSUE #1602: [Bug]: --group=default unsupported in Pest
### What Happened

In PHPUnit, you can run ungrouped tests by specifying to run the default group

In theory, you should be able to do --exclude-group but this doesn't work with multiple groups in parallel https://github.com/pestphp/pest/issues/1437

PHPUnit style tests run through Pest still work fine and are grouped default

### How to Reproduce

Add a test with Pest formatting
Run with --group=default

### Sample Repository

_No response_

### Pest Version

4.1.0

### PHP Version

8.3.29

### Operation System

Linux

### Notes

This, or the parallel exclude group, is a blocker to being able to transition to pest style tests

YoussefMansour9:
Fixes #1602 - Running pest --group=default now correctly runs Pest tests that don't have an explicit group, matching PHPUnit's behavior.

## What happened
When running `./vendor/bin/pest --group=default`, zero tests were executed, even though there were Pest tests without explicit groups.

## Root cause
PHPUnit treats tests without an explicit @group annotation as being in the "default" group. However, Pest tests created with `test()` or `it()` didn't have any group assigned, so --group=default found nothing.

## Solution
Add tests without an explicit group to the "default" group in TestRepository, matching PHPUnit's behavior.

### DIFF
--- a/src/Repositories/TestRepository.php
+++ b/src/Repositories/TestRepository.php
@@ -209,6 +209,18 @@ final class TestRepository
                     $method->groups = [...$groups, ...$method->groups];
                 }
 
+                if ($groups === []) {
+                    foreach ($testCase->methods as $method) {
+                        if ($method->groups === []) {
+                            $method->groups[] = 'default';
+                            $method->attributes[] = new Attribute(
+                                Group::class,
+                                ['default'],
+                            );
+                        }
+                    }
+                }
+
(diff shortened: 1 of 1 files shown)


nunomaduro:
No plans for this, sorry.

Issue #1711, opened 2026-05-18
[Bug]: --shard silently drops test files with non-Tests\ namespaces (e.g. modular app structures)


Orrison:
### What Happened

The `--shard` option silently ignores test classes whose fully-qualified name does not start with `Tests\`. This affects any project with test files outside the standard `tests/` directory — such as modular Laravel applications with test directories like `app-modules/*/tests/`.

### Root Cause

In `src/Plugins/Shard.php`, the `allTests()` method parses `--list-tests` output with this regex:

```php
preg_match_all('/ - (?:P\\\\)?(Tests\\\\[^:]+)::/', $output, $matches);
```

This hardcodes that the class identifier after the `P\` prefix must start with `Tests\`. However, when tests live outside the `tests/` directory, Pest generates path-based identifiers like:

```
 - P\Appmodules\ai\tests\Tenant\Actions\SomeTest::__pest_evaluable_...
```

After stripping `P\`, the remaining identifier starts with `Appmodules\`, not `Tests\`, so the regex never matches.

### Impact

In my project:
- **2540** total test cases in the Tenant suite
- **189** matched by the regex (tests in `tests/Tenant/`)
- **2351** silently dropped (tests in `app-modules/*/tests/Tenant/`)

The shard reports "16 files ran, out of 16" because it only sees 16 test files total. The remaining ~90% of the test suite is invisible to sharding — they either all run unsharded in every shard (wasting CI time) or are filtered out entirely by the `--filter` argument.

There is **no warning or error** indicating that tests were missed.

### Expected Behavior

The regex should match **any** test class identifier from `--list-tests` output, not just those namespaced under `Tests\`. A fix would be:

```php
preg_match_all('/ - (?:P\\\\)?([^:]+)::/', $output, $matches);
```

### How to Reproduce

1. Create a project with tests outside `tests/` (e.g. in `app-modules/module-name/tests/`)
2. Configure `phpunit.xml` testsuite to include those directories
3. Run `pest --testsuite=YourSuite --shard=1/2`
4. Observe the shard only reports tests from `tests/` — app-module tests are missing

### Sample Repository

https://github.com/canyongbs/advisingapp/pull/2549

### Pest Version

v4.7.0

### PHP Version

8.4

### Operation System

macOS

### Notes

- Pest v4.7.0
- PHP 8.4
- Laravel modular application with ~40 modules in `app-modules/`

mozex:
[...]

PR #1713, opened 2026-05-18
Bugfix/sharding module fix

Orrison:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Resolve an issue with sharding tests in a repo that has tests in non-standard tests directories (such as a repo using modular approaches, more details in the related issue)

In my tests I actually test the sharding functionality to make sure it is all working so it ends up actually generating a `tests/.pest/shards.json` file. Which is why during the `afterAll` of the test I `unlink` it.
This is fine now, but will be problematic and weird if Pest itself ever wants to shard it's own tests.
So I just wanted to ask if this was acceptable for now? If not, my thought was I could also add to this PR a new option to override the path `shards.json` is written to.
As maybe someone at some point would want that and it would allow us to point these tests to generate the file somewhere we can `.gitignore` it.

### Related:

https://github.com/pestphp/pest/issues/1711

### DIFF
--- a/src/Plugins/Shard.php
+++ b/src/Plugins/Shard.php
@@ -94,8 +94,8 @@ final class Shard implements AddsOutput, HandlesArguments, Terminable
         if (Parallel::isWorker() && Parallel::getGlobal('UPDATE_SHARDS') === true) {
             self::$updateShards = true;
 
-            Event\Facade::instance()->registerSubscriber(new EnsureShardTimingStarted);
-            Event\Facade::instance()->registerSubscriber(new EnsureShardTimingFinished);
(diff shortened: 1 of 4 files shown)


Orrison:
Hey @nunomaduro! Sorry to bother but I just want to bump this is this issue makes it impossible for modular based projects to use sharding and there is now multiple different PRs/Issues with different people trying to implement roughly this same fix.

Would it be possible to have this taken a look at sometime soon? Thank you so much!

adriaanzon:
Looks like https://github.com/pestphp/pest/pull/1677 fixed this in Pest 5

PR #1716, opened 2026-05-19
feat: add ignorePaths method to exclude specified paths from arch ana…

sonalidudhia:
## Summary

This PR introduces a minimal `ignorePaths()` API for architecture expectations to reduce noise during architecture analysis.

In larger projects, architecture rules may unintentionally analyze generated, infrastructure, or third-party files (for example migrations, vendor code, cache, or generated artifacts), which can create false positives unrelated to application boundaries.

Example:

```php
arch()
    ->expect('App')
    ->ignorePaths([
        'vendor',
        'database/migrations',
    ])
    ->not->toUse(['dd', 'dump']);
```

## Why

Architecture tests are typically intended to validate application boundaries and conventions.

In practice, projects often contain directories that are:

* generated
* infrastructure-related
* historical (e.g. migrations)
* third-party

These may not be desirable targets for architecture expectations and can introduce noise into otherwise intentional rules.

This PR adds a small, framework-agnostic API to selectively exclude paths from architecture analysis.

## Design

The implementation intentionally keeps the API minimal:

* single fluent API: `ignorePaths(array $paths)`
* framework agnostic
* chainable with existing expectations
* cross-platform path normalization
* no additional configuration
* no framework-specific helpers

## Tests

Added coverage for:

* single ignored path
* multiple ignored paths
* nested directory matching
* partial path matching
* Windows path separators
* non-ignored files remaining analysable
* chaining with existing expectations

### DIFF
--- a/src/Expectation.php
+++ b/src/Expectation.php
@@ -435,6 +435,39 @@ final class Expectation
     }
 
     /**
+     * Excludes files whose paths contain any of the given path fragments from arch analysis.
+     *
+     * @param  list<string>  $paths
+     */
+    public function ignorePaths(array $paths): PendingArchExpectation
+    {
+        $normalizedPaths = array_values(array_filter(array_map(
+            static fn (string $path): string => trim(str_replace(['/', '\\'], DIRECTORY_SEPARATOR, $path), DIRECTORY_SEPARATOR),
+            $paths,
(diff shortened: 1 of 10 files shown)


nunomaduro:
no plans for this atm

PR #1741, opened 2026-06-26
fix: strip `--processes` from `--list-tests` when sharding in parallel

### LINKED ISSUE #1731: [Bug]: --shard + --parallel exits 1 when shards.json is missing/incomplete (count-based fallback)
### What happened?

Running `--shard=N/M` together with `--parallel` exits non-zero even though all tests pass — **but only when a `tests/.pest/shards.json` timings file is absent or doesn't cover the sharded classes** (i.e. the count-based fallback path). With a complete, fresh `shards.json` (time-balanced path), the same command exits 0.

> **Correction to the original report:** I first attributed this to PHPUnit 12's `failOnPhpunitWarning` default flipping the exit code on a "No tests found" worker warning. After tracing it properly from a clean vendor checkout, that was wrong for this case — setting `failOnPhpunitWarning="false"` has **no effect**. The real cause is the `--list-tests` sub-invocation described below, which is the same root cause as #1454.

### Reproduction

Same suite, ~176 feature test classes, `--parallel`:

| Condition | Exit |
|---|---|
| `pest tests/Feature --shard=1/2 --parallel`, **no** `shards.json` | **1** ❌ |
| `pest tests/Feature --shard=1/2 --parallel`, **fresh** `shards.json` | 0 ✓ |
| `pest tests/Feature --parallel` (no `--shard`) | 0 ✓ |

When it fails, the error is:

```
Symfony\Component\Process\Exception\ProcessFailedException

The command "'php' './vendor/bin/pest' 'tests/Feature' '--processes=2'
'--cache-directory' '.../vendor/pestphp/pest/.temp' '--no-output' '--list-tests'" failed.

Exit Code: 2(Misuse of shell builtins)

Output:
   INFO  Unknown option "--processes". Most similar options are --covers, --colors, ...
```

### Root cause

On the **count-based fallback** path (no usable `shards.json`), Pest's sharding enumerates the suite by spawning a child `pest ... --list-tests` process to build the `--filter` regex. Under `--parallel`, that child invocation is passed `--processes=N`, but `--list-tests` does not accept `--processes`, so it exits 2 and Pest's `Process::mustRun()` throws `ProcessFailedException` → the whole command exits non-zero.

This is the **same underlying bug as #1454** (`Unknown option "--processes"` when sharding). This issue adds the detail that it is **gated by the presence of a complete `shards.json`**: the time-balanced path doesn't hit the broken `--list-tests` enumeration, so committing a current `tests/.pest/shards.json` is a reliable workaround — but any newly added class missing from that file silently drops you back onto the failing count-based path.

Likely fix: strip `--processes` / `--parallel` from the argv passed to the internal `--list-tests` sub-invocation (cf. the `--parallel`-stripping already done in `Shard.php::handleArguments`).

### Versions

- Pest: 4.7.2
- PHPUnit: 12.5.28
- ParaTest: 7.20.0
- PHP: 8.5.4

lazerg:
### What:

- [x] Bug Fix

### Description:

When `--shard` is used with `--parallel` and there's no usable `tests/.pest/shards.json`, Pest builds the shard's `--filter` by spawning an internal `--list-tests` run. `Shard::removeParallelArguments()` only stripped `--parallel`/`-p` from that sub-process, so it still received `--processes`, which `--list-tests` rejects — `mustRun()` then threw `ProcessFailedException` and the whole run exited non-zero even though every test passed. This strips `--processes` from that sub-invocation too, alongside the existing `--parallel` removal.

### Related:

Closes #1731 (same root cause as #1454).

### DIFF
--- a/src/Plugins/Shard.php
+++ b/src/Plugins/Shard.php
@@ -204,7 +204,34 @@ final class Shard implements AddsOutput, HandlesArguments, Terminable
      */
     private function removeParallelArguments(array $arguments): array
     {
-        return array_filter($arguments, fn (string $argument): bool => ! in_array($argument, ['--parallel', '-p'], strict: true));
+        $filtered = [];
+        $skipNext = false;
+
+        foreach ($arguments as $argument) {
+            if ($skipNext) {
+                $skipNext = false;
+
+                continue;
+            }
+
+            if (in_array($argument, ['--parallel', '-p'], strict: true)) {
+                continue;
+            }
+
+            if ($argument === '--processes') {
+                $skipNext = true;
+
+                continue;
+            }
+
+            if (str_starts_with($argument, '--processes=')) {
+                continue;
+            }
+
+            $filtered[] = $argument;
+        }
+
+        return $filtered;
     }
 
     /**
--- a/tests/Visual/Parallel.php
+++ b/tests/Visual/Parallel.php
@@ -41,6 +41,12 @@ test('a parallel test can extend another test with same name', function () use (
     expect($run('tests/Fixtures/Inheritance'))->toContain('Tests:    1 skipped, 1 passed (1 assertions)');
 })->skipOnWindows();
 
+test('a count-based shard runs in parallel when no shards file exists', function () use ($run) {
+    expect($run('tests/Fixtures/Inheritance', '--shard=1/1'))
+        ->toContain('Tests:    1 skipped, 1 passed (1 assertions)')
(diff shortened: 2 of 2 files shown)


lazerg:
Hey @nunomaduro, just a gentle bump here. It has been a few weeks since I opened this one. It fixes a real crash when `--shard` and `--parallel` are combined. No rush, just did not want it to get lost. I also have #1742 and #1743 open with the same kind of fix, so feel free to look at all three together if that is easier.

PR #1743, opened 2026-06-26
fix: preserve the global afterAll hook until class teardown

### LINKED ISSUE #1694: [Bug]: AfterAll() hook not being called
### What Happened

I am trying to use the AfterAll() global hook to reset fixture data in a file.

What I expected was that the data would be reset.

What happens is nothing; the closure I set in Pest.php is never called.

If I set a breakpoint on the closure  it is never reached.

By contrast, the BeforeAll() (which I use to initialise the fixture data), BeforeEach()  and  AfterEach() global hooks work as expected and pause on breakpoints when set.

### How to Reproduce

In Pest.php
`
pest()->in('Feature')
    ->extend(TestCase::class)
    ->beforeAll(function () {
        TestCase::beforeAll();
    })
    ->afterAll(function () {
        TestCase::afterAll();
    })
;
`

In TestCase.php
`
abstract class TestCase extends BaseTestCase {
    public static function afterAll(): void
    {
        // clear fixture - set a breakpoint here and it will not be hit
    }

    public static function beforeAll(): void
    {
        // initialise fixture - set a breakpoint here and it will be hit
    }
}
`

### Sample Repository

_No response_

### Pest Version

pest: 4.7 &  pest-plugin-browser: 4.3.1

### PHP Version

8.5.3

### Operation System

Linux

### Notes

Looking in Testable::tearDownAfterClass() (I assuming this is the right place as the comment states that "This method is called after the last test of this Test Case is run."), $afterAll is an instance of Support/NullClosure.php; I kind of think it should be the closure I set in Pest.php.

lazerg:
### What:

- [x] Bug Fix

### Description:

A global `afterAll()` hook (registered via `uses()`/`pest()`) never runs, while `beforeAll`, `beforeEach` and `afterEach` work. `beforeAll` is consumed in `setUpBeforeClass` before any test runs, but the last test's `tearDown` flushes the static `$__afterAll` before `tearDownAfterClass` reads it, so the global afterAll silently no-ops. This captures the hook in `setUpBeforeClass` while it's still set and uses that copy at teardown.

### Related:

Closes #1694

### DIFF
--- a/src/Concerns/Testable.php
+++ b/src/Concerns/Testable.php
@@ -118,6 +118,11 @@ trait Testable
     private static ?Closure $__afterAll = null;
 
     /**
+     * The test's after all closure, preserved for the class teardown.
+     */
+    private static ?Closure $__afterAllForClass = null;
+
+    /**
      * The list of snapshot changes, if any.
      */
     private array $__snapshotChanges = [];
@@ -214,6 +219,8 @@ trait Testable
             $beforeAll = ChainableClosure::boundStatically(self::$__beforeAll, $beforeAll);
         }
 
+        self::$__afterAllForClass = self::$__afterAll;
+
         try {
             call_user_func(Closure::bind($beforeAll, null, self::class));
         } catch (Throwable $e) {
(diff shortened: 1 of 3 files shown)


lazerg:
Hi @nunomaduro, just checking back on this one. It's been about three weeks since I flagged it. The global afterAll hook does not run right now. Let me know if you need anything from me to move it forward.

PR #1752, opened 2026-07-12
Fix project root resolution when vendor is a symlinked directory (git worktree support)

ceilidhboy:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

When running Pest from a git worktree where `vendor/` is a symlink to another worktree (e.g. to avoid the overhead of duplicating dependencies), the project root path was incorrectly resolved to the parent of the resolved vendor directory rather than the worktree root.

The root cause is that PHP's `__DIR__` and `__FILE__` resolve through symlinks. Since `bin/pest` lives inside the `vendor/` directory, if `vendor/` is a symlink then `__DIR__` follows it, causing `dirname(__DIR__, 4)` to point at the wrong worktree's root. The same issue affected `bin/worker.php` for parallel test execution.

The fix checks whether the current working directory has a `vendor` symlink that resolves to the same location as the autoloader's vendor parent. If so, the CWD is used as the project root.

### Related:

For detailed background on the worktree + symlinked vendor use case, see the [wiki page on my fork](https://github.com/ceilidhboy/pest/wiki).

### DIFF
--- a/bin/pest
+++ b/bin/pest
@@ -142,21 +142,7 @@ use Symfony\Component\Console\Output\ConsoleOutput;
         $autoloadPath = $localPath;
     }
 
-    // Get $rootPath based on $autoloadPath
-    $rootPath = dirname($autoloadPath, 2);
-
-    // Worktree / symlinked vendor support
-    // If the CWD has a vendor symlink pointing to the same
-    // resolved vendor directory, use CWD as root so worktree
-    // tests find their own Pest.php, tests/, and phpunit.xml.
-    $cwd = getcwd();
-    if ($cwd !== false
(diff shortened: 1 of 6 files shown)


ceilidhboy:
I'm closing this pull request because although symlinking vendor and node packages directories is a great idea to save massive duplication packages which can easily consume up to 1GB of disk space per worktree, and also speeds up creation of new work trees, there were additional subtle issues with creating soft links. I fixed that by using hard links, and one upshot of that is that my solution no longer requires any changes to Pest, and therefore I'm closing this pull request as it's no longer required. I can just use the standard Pest.

Issue #1483, opened 2025-08-27
[Bug]: Using `--parallel` causes overall process to fail with no reporting as to why even though all tests pass


ConorEdwardsCP:
### What Happened

Our team noticed after upgrading to Pest 4 that our tests running in parallel are showing all tests as passing, but having the overall command fail. However, if we run the exact same test suite sequentially, we get the same number of passing tests and the overall command passes.

## Sequential

```
./vendor/bin/pest --exclude-group=fails-in-parallel  --testsuite=Feature --compact

  ................................................................................................................................................................
  ............................................................................

  Tests:    236 passed (364 assertions)
  Duration: 139.51s

echo $?
0
```

## Parallel

```
./vendor/bin/pest --exclude-group=fails-in-parallel  --testsuite=Feature --parallel

  ................................................................................................................................................................
  ............................................................................

  Tests:    236 passed (364 assertions)
  Duration: 30.20s
  Parallel: 12 processes

echo $?
2
```

### How to Reproduce

1. Run test suite without `--parallel` flag and see it passes with the same number of tests and assertions
2. Run test suite with `--parallel` flag and see it returns an error code of 2, even though it shows the same number of tests and assertions passing successfully

### Pest Version

4.0.3

### PHP Version

8.4.11

### Operation System

macOS

### Notes

I'm not certain if it's related, but I've also noticed that when we run tests in parallel now and there is a test failure, it is swallowing the error and immediately killing the tests without reporting anything about what caused the failure. There was a previous issue #1451 that hinted at their problem could be related to the master process not properly handling the exit codes of the worker threads, which feels like it could be a reason for both of these issues.

ConorEdwardsCP:
For the record, my only indication it's related to Pest 4 is that we didn't have this issue at all when running it on Pest 3. I also am not super familiar with what I can do to help debug or print out any errors related to the process, so if there's anything I can do to help give information let me know!

thomas-0816:
see https://github.com/pestphp/pest/issues/1462

chrispmcgee:
I am running into this same problem now, and it is preventing us from upgrading to Pest 4. Are there any plans to address this?

**October 2025 update:** Still broken in v4.1.2
**December 2025 update:** Still broken in v4.1.6
**January 2026 update:** Still broken in v4.3.1

chrispmcgee:
> see [#1462](https://github.com/pestphp/pest/issues/1462)

Is there something actionable in this link that can be used to fix or work around this issue?

[...]

PR #1757, opened 2026-07-16
fix: parallel runs with group filters no longer exit 1 despite green tests

User7142:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

`--parallel` runs that use `--group`/`--exclude-group` can exit 1 with a fully green summary and no error output.

Cause: a wrapper worker's serialized `TestResult` only carries the `numberOfTests` of the **last** file it executed (PHPUnit's `Collector` overwrites the counter on every `executionStarted`). If a worker's final file has all tests excluded by the group filter, it reports `hasTests() === false` despite having run tests. When every worker ends on such a file — a timing lottery, hence flaky — the merged result in `WrapperRunner::complete()` claims the suite was empty, and since PHPUnit 12 enables `failOnEmptyTestSuite` implicitly for explicit selections, the run exits 1. The "No tests found" runner warnings are already filtered from the output, so nothing is printed.

Deterministic repro (fails before this fix, passes after — added as a test):

```bash
php bin/pest --parallel --processes=1 --group=filtered-group tests/.tests/ParallelGroupFilteredLastFile
# "Tests: 1 passed (1 assertions)" — exit code 1
```

Fix: while merging worker results, treat a worker with `numberOfTestsRun() > 0` as having tests. A companion test ensures genuinely empty selections still exit non-zero.

### Related:

- #1483

### DIFF
--- a/src/Plugins/Parallel/Paratest/WrapperRunner.php
+++ b/src/Plugins/Parallel/Paratest/WrapperRunner.php
@@ -323,8 +323,18 @@ final class WrapperRunner implements RunnerInterface
             /** @var list<AfterLastTestMethodFailed> $failedEvents */
             $failedEvents = array_merge_recursive($testResultSum->testFailedEvents(), $testResult->testFailedEvents());
 
+            // A worker's `hasTests()` only reflects the *last* file it executed:
+            // PHPUnit's Collector overwrites `numberOfTests` on every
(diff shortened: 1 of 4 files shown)


CovertError:
it would be amazing if this is gets merged if it fixes this annoying issue @nunomaduro

PR #1770, opened 2026-07-25
fix: apply negated toUse to each target in a multi-element expect

### LINKED ISSUE #1751: Negated toUse with a multi-element expect([...]) is silently inert — the arch rule can never fail
### Summary

`expect(['App\A', 'App\B'])->not->toUse('App\Forbidden')` never fires, no matter how egregiously one of the targets uses the forbidden dependency. The rule silently stops binding: the suite stays green and reviewers believe the boundary is policed.

Because arch rules exist to *enforce* boundaries, a rule that quietly becomes inert is worse than no rule at all. We hit this in a real codebase — several negated layer rules had been passing for months while the violation they were meant to catch was live.

### Root cause

`OppositeExpectation::toUse()` decomposes the *dependencies* into one `SingleArchExpectation` per dependency (each with an `opposite` callback), but the *targets* of the original `expect([...])` stay bundled in a single blueprint:

```php
// src/Expectations/OppositeExpectation.php
public function toUse(array|string $targets): ArchExpectation
{
    /** @var Expectation<array<int, string>|string> $original */
    $original = $this->original;

    return GroupArchExpectation::fromExpectations($original, array_map(fn (string $target): SingleArchExpectation => ToUse::make($original, $target)->opposite(
        fn () => $this->throwExpectationFailedException('toUse', $target),
    ), is_string($targets) ? [$targets] : $targets));
}
```

The positive `Blueprint::expectToUse()` fails as soon as **any** target does not use the dependency. `SingleArchExpectation::ensureLazyExpectationIsVerified()` then reads that failure as "the positive expectation failed, so the negation passes."

With two or more targets in the `expect([...])`, some target virtually always skips the dependency — so the positive form always throws, and the negated rule therefore always passes, including when another target blatantly violates it.

By contrast, `Targeted`-based matchers (`toHaveSuffix`, `toUseTrait`, …) have dedicated per-object opposite predicates and bind correctly over multi-element expects.

### Minimal repro

```php
// src/A/UsesDb.php
namespace App\A; use Illuminate\Support\Facades\DB;
final class UsesDb { public function x(): void { DB::transaction(fn () => null); } }

// src/B/Clean.php
namespace App\B;
final class Clean {}

// tests/ArchTest.php
arch('single target — correctly FAILS')
    ->expect('App\A')->not->toUse('Illuminate\Support\Facades\DB');

arch('multi target — silently PASSES despite the same violation')
    ->expect(['App\A', 'App\B'])->not->toUse('Illuminate\Support\Facades\DB');
```

The first rule fails as expected. The second passes.

### Expected

Either:

1. decompose the expect targets the same way the dependencies already are — one inverted blueprint per (target, dependency) pair; or
2. reject multi-element expects on negated layer matchers loudly with `InvalidExpectation`, which is exactly how `not->toUseNothing()` and `not->toOnlyUse()` already behave.

Silently passing is the one outcome that shouldn't be on the table.

### Workaround

One `arch()` rule per target:

```php
foreach (['App\A', 'App\B'] as $ns) {
    arch("{$ns} does not use DB")->expect($ns)->not->toUse('Illuminate\Support\Facades\DB');
}
```

### Versions

- `pestphp/pest` v4.6.3
- `pestphp/pest-plugin-arch` v4.0.2
- PHP 8.5

(Filed against `pestphp/pest` because the affected file lives here and issues are disabled on `pest-plugin-arch`.)

lazerg:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

A negated `toUse` on a multi-element `expect([...])` never failed. The positive check fails as soon as one target doesn't use the dependency, and the negation reads that failure as a pass, so a real violation by another target slips through. This decomposes the expect targets the way the dependencies already are, so each one is checked on its own.

### Related:

Fixes #1751

### DIFF
--- a/src/Expectations/OppositeExpectation.php
+++ b/src/Expectations/OppositeExpectation.php
@@ -12,7 +12,6 @@ use Pest\Arch\Expectations\ToBeUsedInNothing;
 use Pest\Arch\Expectations\ToUse;
 use Pest\Arch\GroupArchExpectation;
 use Pest\Arch\PendingArchExpectation;
-use Pest\Arch\SingleArchExpectation;
 use Pest\Arch\Support\FileLineFinder;
 use Pest\Exceptions\InvalidExpectation;
 use Pest\Exceptions\MissingDependency;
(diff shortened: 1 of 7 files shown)


lazerg:
@nunomaduro could you take a look at this when you get a chance.

lazerg:

PR #1772, opened 2026-07-28
Support Laravel 12 by allowing Symfony Process 7

dbpolito:
Laravel 12 requires Symfony Process `^7.2.0`, while Pest 5 currently requires Symfony Process `^8.1.0`.

This updates Pest's constraint to `^7.4.8 || ^8.1.0`, restoring Laravel 12 compatibility while retaining Symfony Process 8 support. Symfony 7.4.8 is the minimum supported version because it is required by ParaTest.

Related browser plugin PR: https://github.com/pestphp/pest-plugin-browser/pull/239

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "pestphp/pest-plugin-mutate": "^5.0.0",
         "pestphp/pest-plugin-profanity": "^5.0.0",
         "phpunit/phpunit": "^13.2.4",
-        "symfony/process": "^8.1.0"
+        "symfony/process": "^7.4.8 || ^8.1.0"
     },
     "conflict": {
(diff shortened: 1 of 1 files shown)


nunomaduro:
Sorry, no plans of supporting Symfony 7

dbpolito:
@nunomaduro 

Thanks for taking a look. I completely respect the decision, but would you be open to reconsidering? Pest isn’t using any Symfony Process 8-specific APIs, and ParaTest already supports `^7.4.8 || ^8.1.0`, so this wouldn’t hold Pest back. It would simply add compatibility with Symfony 7.4 LTS and allow Laravel 12 users to install Pest 5.

PR #1776, opened 2026-07-28
[Rector] Skip RemoveDefaultValueFromAssignedPropertyRector for now

samsonasik:
### What:

- [x] QA

### Description:

`RemoveDefaultValueFromAssignedPropertyRector` is new rule and seems still buggy, the fix is on the way:

- https://github.com/rectorphp/rector-src/pull/8214

Here skip for now.

### Related:

To fix CI failure https://github.com/pestphp/pest/actions/runs/30383066819/job/90355447958#step:10:23


samsonasik:
Ready to merge 👍

samsonasik:
The fix seems already in latest release so closing it.

PR #1787, opened 2026-07-29
feat: allow configuring the TIA directory

dbpolito:
Adds support for configuring where TIA stores its files:

```php
pest()->tia()->directory('.pest/tia');
```

Relative paths are resolved from the project root, while absolute paths are used directly.

The existing `~/.pest/tia/<project-key>` location remains the default when no custom directory is configured.


LudovicLemarinel:
Hi,

I second this : this would make much easier to share the tia directory through the CI, instead of having an extra step to do so !

PR #1806, opened 2026-07-31
fix: skip user hooks when replaying cached skipped, incomplete, and failed tests

### LINKED ISSUE #1785: TIA cached-replay run fires before/afterEach hooks for skipped tests outside the Laravel test bootstrap → "A facade root has not been set."
## Bug description

On a **fresh** `--tia` graph, skipped tests are correctly skipped and everything passes. On the **next** run — once TIA replays results from its cache instead of re-executing — every skipped test in the file fails with:

```
RuntimeException: A facade root has not been set.
  at vendor/laravel/framework/src/Illuminate/Support/Facades/Facade.php:361
```

The failure isn't inside the skipped test's own body (confirmed: one of our failing tests has an empty `{ }` body and still fails identically). It's inside a shared `afterEach()`/`beforeEach()` hook that calls a Laravel facade (e.g. `Storage::disk(...)`). The test body itself is correctly skipped, but the hook still runs for it during replay — and does so without the app/test bootstrap active, so any facade call in that hook throws.

## Reproduction

```php
afterEach(function () {
    Storage::disk('public')->deleteDirectory('some-dir');
});

test('a real test', function () {
    // passes fine
});

test('a skipped test', function () {
    //
})->skip('anything');
```

```
vendor/bin/pest --tia          # fresh graph — passes, skip honored
vendor/bin/pest --tia          # replay from cache — fails
```

Also reproducible with `beforeEach(fn () => ...)->skip('reason')` chained onto a shared `beforeEach` (skipping every test in the file that way) — same crash, same root cause.

Workaround: delete `~/.pest/tia/<project>` before every run (forces a fresh graph, but defeats the point of TIA).

## Environment

- pestphp/pest: 5.0.1 (`--version` reports `5.0.0-beta.1`, which may itself be a separate versioning-display bug worth flagging)
- laravel/framework: 13.23.0
- PHP: 8.4.23 (NTS, clang 15.0.0, macOS arm64)
- Reproduced both with and without `--parallel`

oddvalue:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

When TIA replays a cached **Pass/Risky** result, `Testable::setUp()` sets `$__replay` and returns early, and `tearDown()` checks `$__replay` and short-circuits too — neither `parent::setUp()` (which, in Laravel suites, boots the application) nor the user's `beforeEach`/`afterEach` chains run. That symmetry is the established semantic of replay: a replayed test executes no user hooks.

The cached **Skipped**, **Incomplete**, and **Failure** branches broke that symmetry: they throw (`markTestSkipped()` / `markTestIncomplete()` / `AssertionFailedError`) from inside `setUp()` **before** `parent::setUp()` has run, while leaving `$__replay` at `ReplayType::None`. PHPUnit's `runBare()` still invokes the after-test hooks when `setUp()` throws, and because `$__replay` was `None`, Pest's `tearDown()` did not short-circuit — the full `afterEach` chain and `parent::tearDown()` ran against a test case that was never set up (no booted app, no `Facade::$app`, empty container).

In a Laravel suite, any `afterEach` that touches the framework (e.g. `DB::flushQueryLog()`) then throws `A facade root has not been set`, and PHPUnit converts the cached skip into an **error** (`TestCase::runBare()` explicitly promotes a `SkippedWithMessageException` to an error when tearDown throws). This also made TIA results oscillate between runs: the errored result evicted the cached skip, the next run executed the test green and re-cached the skip, and the run after that errored again.

The fix marks the replay state before the throwing branches dispatch, so `tearDown()` short-circuits for every replayed result exactly as it does on the Pass/Risky path. `__runTest()` is unreachable after a `setUp()` throw, so the hoisted `__beginReplay()` has no effect on the Pass/Risky behaviour, and the replayed skip/incomplete statuses are re-recorded unchanged, keeping the cache stable.

The regression test seeds a TIA graph with cached pass/skipped/incomplete results for a fixture whose `afterEach` throws (and logs to a spy file), then replays it in a subprocess: all three results replay with their cached statuses, exit code 0, and no hook executes. Without the fix, the cached skip becomes a `RuntimeException` error from the `afterEach` hook.

### Related:

- Fixes #1785

### DIFF
--- a/src/Concerns/Testable.php
+++ b/src/Concerns/Testable.php
@@ -286,8 +286,10 @@ trait Testable
         if ($replay !== ReplayType::None) {
             assert($status !== null);
 
+            $this->__beginReplay($replay, $tia);
+
             match ($replay) {
-                ReplayType::Pass, ReplayType::Risky => $this->__beginReplay($replay, $tia),
+                ReplayType::Pass, ReplayType::Risky => null,
                 ReplayType::Skipped => $this->markTestSkipped($status->message()),
                 ReplayType::Incomplete => $this->markTestIncomplete($status->message()),
                 ReplayType::Failure => throw new AssertionFailedError($status->message() ?: 'Cached failure'),
--- a/tests/.snapshots/success.txt
+++ b/tests/.snapshots/success.txt
@@ -1565,6 +1565,9 @@
   ✓ it allows performing no expectations without being risky
   ✓ a "describe" group of tests → it allows performing no expectations without being risky
 
+   PASS  Tests\Features\TiaReplay
+  ✓ replaying cached skipped and incomplete results does not run user hooks
+
    PASS  Tests\Features\Ticket
   ✓ it may be associated with an ticket #1, #2
   ✓ nested → it may be associated with an ticket #1, #4, #5, #6, #3
@@ -2186,4 +2189,4 @@
   ✓ pass with dataset with ('my-datas-set-value')
   ✓ within describe → pass with dataset with ('my-datas-set-value')
 
-  Tests:    1 deprecated, 4 warnings, 5 incomplete, 2 notices, 40 todos, 35 skipped, 1542 passed (3375 assertions)
+  Tests:    1 deprecated, 4 warnings, 5 incomplete, 2 notices, 40 todos, 35 skipped, 1543 passed (3385 assertions)
--- a/tests/.tests/TiaReplayHooks.php
+++ b/tests/.tests/TiaReplayHooks.php
@@ -1,0 +1,32 @@
+<?php
+
+function tiaReplayHooksLog(string $hook): void
+{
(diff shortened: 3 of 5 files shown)


einar-hansen:
Just want to +1 this — we ran into the exact same thing yesterday and spent half a day chasing it before finding this PR. 😅

Our setup: ~11k test Laravel suite where tearDown() resolves tenancy from the container, and 14 tests that skip at runtime when ffmpeg isn't installed. Replays kept flip-flopping — run it once, 14 errors ("A facade root has not been set"), run it again, all green, run it again, 14 errors. No code changes in between. Drove us a bit crazy until we instrumented the graph and saw the cached skips getting overwritten by the teardown errors, exactly like this PR describes.

So: confirmed independently, fix works great on a big real-world suite. Would love to see this merged 🙏

nunomaduro:
fixed

Issue #1802, opened 2026-07-31
[Bug]: --tia unusable in monorepos


RomainM27:
### What Happened

When a Pest project (folder holding `composer.json`) is not the root of its git repository `--tia` refuses to run and panics with `TiaRequiresRepositoryRoot`.

### How to Reproduce

 Repository layout:
  ```
  /repo/                       ← git root
  ├── .git/
  ├── apps/
  │   └── laravel-app/          ← Pest project
  │       ├── composer.json
  │       ├── tests/
  │       └── vendor/
  └── libs/
  ```
  From `apps/laravel-app/`:
  ```
  $ vendor/bin/pest --tia
  ERROR: Tia mode requires the git repository root.
  This project sits in a subdirectory of a larger repo apps/laravel-app.
  Give the project its own git repository to use Tia.
  ```
  Triggered by `src/Plugins/Tia.php` line 634 (v5.0.2):
  ```php
  $subdirectoryPrefix = $this->gitSubdirectoryPrefix($projectRoot);
  if ($subdirectoryPrefix !== null) {
      Panic::with(new TiaRequiresRepositoryRoot($subdirectoryPrefix));
  }
  ```

### Pest Version

5.0.2

### PHP Version

8.4.17

### Operation System

macOS

RomainM27:
duplicate of https://github.com/pestphp/pest/issues/1783

PR #1809, opened 2026-07-31
feat(tia): support projects in git repository subdirectories (monorepos)

### LINKED ISSUE #1783: TIA: change detection misses all changes when the Pest project is a subdirectory of the git repository (monorepo)
### What happened

In a monorepo where the Pest project lives in a subdirectory of the git repository (e.g. `backend/`), TIA never detects any PHP source change. In the worst case it **replays the previous results as passing without executing the tests affected by the change**.

```
repo/                <- git root
├── frontend/
└── backend/         <- Laravel app, composer.json, pest runs here
    ├── app/
    └── tests/
```

Reproduction:

1. Record a baseline: `PEST_ALLOW_XDEBUG=1 XDEBUG_MODE=coverage vendor/bin/pest --tia` (from `backend/`, all green, graph written).
2. Make a semantic change to `backend/app/Services/Billing/PartialInvoiceService.php` (e.g. add a `private const`).
3. Run `vendor/bin/pest --tia` again.

**Expected:** the tests covering that file are re-executed (in our suite: 40 tests across 4 files).
**Actual:** `Tests: 1669 passed (…, 1669 replayed)` in ~3s — the change maps to zero tests and everything is replayed as unchanged.

With no coverage driver loaded the failure is the safe-but-slow variant instead: `WARN Detected PHP source changes but no coverage driver is available` followed by a full-suite run on every invocation.

### Root cause

`src/Plugins/Tia/ChangedFiles.php` collects changed paths via:

- `workingTreeChanges()` → `git status --porcelain -z --untracked-files=all` (cwd = project root)
- `diffSinceSha()` → `git diff --name-only <sha>..HEAD`

Both git commands print paths **relative to the repository root** (`backend/app/Services/...`), regardless of cwd. Downstream these are consumed as **project-relative** paths:

- `currentHash()` prepends the project root → `…/backend/backend/app/...` → `is_file()` false
- the dependency-graph keys are project-relative (`app/Services/...`) → the changed path never matches an edge → affected set is empty

Verified identical logic in `v4.7.5` (installed) and `v5.0.0` (GitHub source). Projects whose composer root equals the git root never notice, since both path notions coincide.

### Proof it is exactly this

Creating a throwaway git repository *inside* `backend/` (making project root == git root) and repeating the same experiment yields correct behavior immediately: `Tests: 1669 passed (…, 40 affected, 1629 replayed)` in ~100s. Removing the inner repo restores the broken behavior.

### Suggested fix

Ask git for the prefix once (`git rev-parse --show-prefix`, empty at repo root) and strip it from every path returned by `workingTreeChanges()` / `diffSinceSha()` before they are compared with project-relative graph keys; paths outside the prefix (sibling packages, e.g. `frontend/…`) can be discarded or kept for the structural-change checks.

### Environment

- Pest v4.7.5 (same code confirmed in v5.0.0)
- PHP 8.4.10, Laravel 12.64, PostgreSQL, macOS (darwin arm64)
- Layout: Laravel app in `backend/` subdirectory of the git repo, `tests/Pest.php` maps `Tests\TestCase` onto Feature + Unit, suite is 100% functional Pest style (1669 tests)

einar-hansen:
### What:

- [x] Bug Fix
- [x] New Feature

### Description:

Allow TIA to run when a Pest project lives inside a Git repository, such as `apps/api` in a monorepo. Translate repository paths into project paths so dependency selection, baseline detection, and graph storage use the correct project.

- `ChangedFiles` strips the project prefix from Git output, adds it to paths passed to Git, and ignores sibling projects. Null-delimited output preserves non-ASCII filenames, and `--no-renames` preserves deletions when files move outside the project.
- `GitRepository` finds the nearest governing `.git` entry. For worktrees and submodules, it follows the `gitdir` file and resolves relative or absolute paths. Worktrees read the shared config through `commondir`; submodules read their own config. This fixes silent baseline detection failures for both GitHub and GitLab.
- `Storage` gives sibling projects separate graph stores, and `Fingerprint` uses the governing repository for gitignore detection. Root-level projects retain their existing path translations and storage keys.

### Verified:

- The worktree and submodule regressions reproduce ten failures before the config fix. After the fix, `PEST_TIA=0 php bin/pest tests/Unit/Plugins/Tia/GitRepository.php tests/Unit/Plugins/Tia/Baselines.php --compact` passes all 28 cases, including real Git worktrees and submodules for both baseline providers.
- An actual Pest worktree resolves the main checkout's `.git/config` and detects `einar-hansen/pest` as its baseline repository.
- `XDEBUG_MODE=off PEST_TIA=0 composer test` passes for `e933e7f40704f72efcbc5b1b98d9fd93a15d2f56`: lint, static analysis, 100% type coverage, unit tests, parallel tests, and integration tests (239 passed, 1 skipped). Local PHP 8.5.8 used Xdebug so fixtures could toggle coverage through `XDEBUG_MODE`; PCOV was disabled for this run.
- `composer test:profanity` passes. Output snapshots include the current suite totals.

The GitHub baseline artifact name remains `pest-tia-baseline`, so multiple baselined projects in one repository still share that artifact name.

### Related:

- #1805 addresses the same path translation problem. This PR also covers non-ASCII filenames, renames leaving the project, baseline detection, per-project storage, and fingerprints.
- Fixes #1783; covers the scenario in #1802.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kolset:
Hey guys, this will be vey helpful! Also talked with several people waiting for this fix! 

Thanks for the great work🙏

PR #1812, opened 2026-08-01
Fix TIA for Livewire component views

### LINKED ISSUE #1808: [Bug]: TIA invalidates almost the entire suite for Livewire 4 single-file component views
## What Happened

Pest TIA invalidates almost the entire test suite when a Livewire 4 single-file component (SFC) view is changed.

A Livewire SFC source such as:

```text
resources/views/pages/orders/⚡index.blade.php
```

is compiled by Livewire to a worker-specific generated view such as:

```text
storage/framework/views/test_3/livewire/views/<hash>.blade.php
```

Pest's `BladeEdges` records `$view->getPath()`, so the dependency graph contains the generated view path, but not the original SFC source path.

When the original `⚡index.blade.php` is changed, `Graph` cannot find it in the recorded graph. It therefore reaches the watch-pattern fallback. The built-in Livewire rule maps `resources/views/pages/**/*.blade.php` to the entire test path, causing nearly every test file to be marked as affected.

In a real application with 447 configured test files, editing one SFC `⚡index.blade.php` selected 421 test files (94.2%). The generated Livewire view was linked to only one actual feature test in the graph.

Expected behavior: only tests that rendered or otherwise depended on that Livewire SFC should be affected.

Actual behavior: almost the whole suite is affected and its graph edges are refreshed.

This is reproducible with a fresh TIA graph. `--filtered` does not help because the affected-file set is already broad, and custom `watch()` rules are merged with the built-in rule rather than replacing it.

## How to Reproduce

1. Create a Laravel 13 application using Livewire 4 SFC components and Pest 5.
2. Enable PCOV or Xdebug coverage, as required by TIA.
3. Create two unrelated Livewire pages, for example:

   ```text
   resources/views/pages/orders/⚡index.blade.php
   resources/views/pages/users/⚡index.blade.php
   ```

4. Add separate feature tests for each page.
5. Record a fresh baseline:

   ```bash
   ./vendor/bin/pest --parallel --tia --fresh
   ```

6. Change only static Blade markup in `resources/views/pages/orders/⚡index.blade.php`.
7. Run:

   ```bash
   ./vendor/bin/pest --parallel --tia --filtered
   ```

8. Inspect the affected test files and `graph.json`.

The graph records paths under `storage/framework/views/test_N/livewire/views/<hash>.blade.php`, while the changed source path under `resources/views/pages/` is absent. The original source consequently reaches the Livewire watch fallback and invalidates the full test path.

## Sample Repository

Not available yet. The behavior can be reproduced with the steps above.

## Pest Version

5.0.1. The relevant implementation is unchanged in v5.0.2 and the current `5.x` branch.

## PHP Version

8.5.8

## Operation System

macOS

## Notes

Environment:

- Laravel 13
- Livewire 4.3.3
- Parallel TIA execution

Relevant code:

- `BladeEdges` links the rendered view's direct path: https://github.com/pestphp/pest/blob/5.x/src/Plugins/Tia/Edges/BladeEdges.php
- The Livewire watch default maps all SFC page views to the full test path: https://github.com/pestphp/pest/blob/5.x/src/Plugins/Tia/WatchDefaults/Livewire.php
- Livewire generates the cached view name from the original source path hash: https://github.com/livewire/livewire/blob/v4.3.3/src/Compiler/CacheManager.php

A possible fix would be to record both the generated Livewire view and its original SFC source path, or teach TIA to map an unknown changed SFC source to the generated `livewire/views/<hash>.blade.php` graph entry. A regression test should verify that changing one `⚡index.blade.php` affects only the test files linked to its generated view.

lucasromanojf:
## Summary

- map changed Livewire single-file and multi-file component sources to the generated view hashes already recorded by TIA
- preserve direct Blade edges and the existing broad watch-pattern fallback when no safe generated-view match exists
- cover the documented Livewire 4 component locations and formats, including optional emoji filenames, and update the visual test snapshots

## Root cause

Livewire renders single-file and multi-file components through generated views under `storage/framework/views/*/livewire/views/<hash>.blade.php`. TIA recorded those generated paths, but a Git change is reported using the original component path, so the graph could not connect the changed component to its tests.

The graph now derives the same platform-specific source-path hash without requiring Livewire as a dependency and reuses matching generated-view edges. If no matching edge exists, the existing watch-pattern fallback remains unchanged.

## Validation

- `composer test`
- fresh TIA baseline in an integration Laravel application: 1,259 tests passed with 2,820 assertions
- four real Livewire SFC changes across `resources/views/components/**` and `resources/views/pages/**`: exactly 7 related test files selected, with 167 tests and 340 assertions passing
- final integration application gate with `--parallel --tia`: 1,259 tests passed with 2,820 assertions

Fixes #1808

### DIFF
--- a/src/Plugins/Tia/Graph.php
+++ b/src/Plugins/Tia/Graph.php
@@ -100,14 +100,15 @@ final class Graph
 
         $this->applyTestFileChanges($nonMigrationPaths, $affectedSet);
 
-        $staticallyHandledBlade = $this->applyBladeStaticChanges($nonMigrationPaths, $affectedSet);
+        $handledBlade = $this->applyBladeStaticChanges($nonMigrationPaths, $affectedSet)
+            + $this->applyLivewireViewChanges($nonMigrationPaths, $affectedSet);
 
         $this->applyWatchPatternFallback(
             $nonMigrationPaths,
             $unparseableMigrations,
             $preciselyHandledPages,
             $sharedFilesResolved,
-            $staticallyHandledBlade,
+            $handledBlade,
             $affectedSet,
         );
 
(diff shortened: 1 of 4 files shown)


lucasromanojf:
Fechando: o mesmo fix já foi aplicado diretamente no branch 5.x no commit eb88513ba ("fix: livewire", por nuno maduro). O código em 5.x é idêntico ao deste PR (Graph.php e testes). Correção ainda sem tag/release — deve sair na próxima versão.

PR #1813, opened 2026-08-02
Fix TIA fingerprinting inside linked git worktrees

### LINKED ISSUE #1810: [Bug]: In a git worktree every TIA fingerprint input is null, so composer.lock / phpunit.xml changes never invalidate the graph
> **Correction (see comment below):** the original report claimed TIA never replays inside a worktree. That is wrong — it replays fine. The actual defect is the opposite and more serious: the structural fingerprint is entirely `null` in a worktree, so changes that should invalidate the graph never do.

## Summary

Inside a linked `git worktree`, every structural fingerprint input resolves to `null`. Because `null` compares equal to `null`, `Fingerprint::structuralMatches()` reports a match no matter what changed, and **a stale TIA graph is silently reused across `composer.lock`, `phpunit.xml`, `vite.config.*` and JS lockfile changes**.

Concretely: run `composer update`, then `--tia` keeps replaying cached results recorded against the *old* dependency tree.

## Cause

`Fingerprint::isTrackedByGit()` (`src/Plugins/Tia/Fingerprint.php:277`) decides whether a file is tracked using Symfony Finder:

```php
$finder = (new Finder)
    ->in($projectRoot)
    ->depth('== 0')
    ->name($relativePath)
    ->ignoreVCSIgnored(true);   // line 299

return $cache[$key] = $finder->hasResults();
```

In a linked worktree `.git` is a **file** containing `gitdir: …`, not a directory. `ignoreVCSIgnored(true)` does not resolve the real repository through that pointer and `hasResults()` returns `false` for plainly-tracked files. `trackedHash()` then returns `null` for every input, so `Fingerprint::compute()` produces:

```php
'structural' => [
    'schema'        => 18,
    'composer_lock' => null,
    'phpunit_xml'   => null,
    'vite_config'   => null,
    'package_lock'  => null,
    'js_config'     => null,
]
```

Line 291 already special-cases "no `.git` at all" and returns `true`; a worktree falls through it because `.git` exists as a file.

## Reproduction

```bash
git clone <repo> main-clone && cd main-clone
git worktree add ../wt -b tia-repro && cd ../wt
composer install
```

```bash
vendor/bin/pest tests/SomeDir --tia --fresh    # records
vendor/bin/pest tests/SomeDir --tia            # "29 replayed" — correct so far

printf '\n' >> composer.lock                   # dependency tree "changed"
vendor/bin/pest tests/SomeDir --tia            # still "29 replayed" — should have rebuilt
```

Verified on Pest 5.0.2 / PHP 8.4. In a normal clone the same edit correctly reports
`fresh graph (composer.lock … changed)`.

Direct check of the underlying call:

```php
require 'vendor/autoload.php';
use Symfony\Component\Finder\Finder;
$f = (new Finder)->in(getcwd())->depth('== 0')->name('composer.lock')->ignoreVCSIgnored(true);
var_dump($f->hasResults());            // false — but composer.lock is tracked
var_dump(is_file(getcwd().'/.git'));   // true — worktree pointer file
```

## Impact

1. **Stale replay after dependency changes** — the graph is not invalidated when it must be. Silent, and it produces results recorded against different code.
2. **Baselines are not portable between a normal clone and a worktree** in either direction: null fingerprints never match real ones, so a CI-recorded baseline is always discarded in a worktree (and vice versa).

## Suggested fix

Resolve the real git directory when `.git` is a file before consulting Finder, or replace the Finder check with `git ls-files --error-unmatch <path>` / `git check-ignore`, both of which handle worktrees natively. Failing that, treating an *unresolvable* tracking state as "hash it anyway" would fail safe, whereas `null` currently fails open.

ademola-emmanuel:
Fixes #1810.

**The problem**

`Fingerprint::isTrackedByGit()` answers "is this file gitignored?" with Symfony Finder's `ignoreVCSIgnored(true)`. Finder locates the repository root by walking upward until it finds a `.git` directory. In a linked worktree `.git` is a pointer file, so the walk misses the worktree root, and what happens next depends on the directory layout:

- If an ancestor directory contains a `.git` directory (worktrees kept inside the main repository under an ignored `.worktrees/` directory, or a dotfiles repository in `$HOME` that ignores `*`), Finder adopts that ancestor as the repository root and applies its ignore rules to the worktree's files. Every structural file then reports as ignored, `trackedHash()` returns null for all of them, and `--tia` rebuilds the graph on every run with no error, just a permanent "fresh graph" message.
- If no ancestor has one, the check happens to behave correctly, which is why the failure looks machine-dependent.

**The fix**

Ask git itself with `git check-ignore -q`, run from the project root, as the issue suggests. Git resolves the `gitdir:` pointer natively, so worktrees behave like regular clones. This also matches how TIA already answers the same question in `ChangedFiles::filterIgnored()`.

Exit code 0 keeps a file out of the fingerprint (ignored), 1 keeps it in, and anything else (git missing, not a repository) keeps the file, matching the existing no-`.git` early return.

**Verification**

- New test: a repository with a linked worktree under an ignored `.worktrees/` directory. On current 5.x it fails (worktree fingerprints are null); with this change the worktree produces the same structural fingerprint as the clone, which is the property replay depends on.
- The gitignored-untracked-lockfile exclusion documented on the method keeps working, also covered by a new test.
- Full unit suite (1,608 tests), PHPStan, 100% type coverage, rector, and pint all pass.

### DIFF
--- a/src/Plugins/Tia/Fingerprint.php
+++ b/src/Plugins/Tia/Fingerprint.php
@@ -5,7 +5,8 @@ declare(strict_types=1);
 namespace Pest\Plugins\Tia;
 
 use Pest\Plugins\Tia\Contracts\Lockfile;
-use Symfony\Component\Finder\Finder;
+use Symfony\Component\Process\Exception\ExceptionInterface as ProcessException;
+use Symfony\Component\Process\Process;
 
 /**
  * @internal
@@ -279,13 +280,16 @@ final readonly class Fingerprint
             return $cache[$key] = true;
         }
 
-        $finder = (new Finder)
-            ->in($projectRoot)
-            ->depth('== 0')
-            ->name($relativePath)
-            ->ignoreVCSIgnored(true);
+        $process = new Process(['git', 'check-ignore', '-q', '--', $relativePath], $projectRoot);
+        $process->setTimeout(5.0);
 
-        return $cache[$key] = $finder->hasResults();
+        try {
+            $process->run();
+        } catch (ProcessException) {
+            return $cache[$key] = true;
+        }
+
+        return $cache[$key] = $process->getExitCode() !== 0;
     }
 
     private static function contentHashOrNull(string $path): ?string
--- a/tests/Unit/Plugins/Tia/Fingerprint.php
+++ b/tests/Unit/Plugins/Tia/Fingerprint.php
@@ -1,0 +1,109 @@
+<?php
+
+declare(strict_types=1);
+
+use Pest\Plugins\Tia\Fingerprint;
+use Symfony\Component\Process\Process;
+
+function fingerprintGit(string $cwd, string ...$args): void
+{
+    $process = new Process(['git', ...$args], $cwd);
+    $process->setTimeout(10.0);
+    $process->mustRun();
+}
+
+function fingerprintRepository(): string
+{
+    $root = sys_get_temp_dir().DIRECTORY_SEPARATOR.'pest_fingerprint_'.uniqid();
+    mkdir($root, 0777, true);
+
+    fingerprintGit($root, 'init', '-q');
+    fingerprintGit($root, 'config', 'user.email', 'pest@example.com');
+    fingerprintGit($root, 'config', 'user.name', 'Pest');
+
+    return $root;
+}
+
+function fingerprintRemoveDirectory(string $path): void
+{
+    if (! is_dir($path)) {
+        return;
+    }
+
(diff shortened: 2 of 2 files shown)


ademola-emmanuel:
Rebased onto 5.x a while back, so this is currently conflict-free, and the suite, PHPStan, type coverage, and Pint are all green locally. CI is still waiting on first-time approval whenever someone has a moment.

Happy to rebase again or rework the approach if anything about it doesn't fit.

PR #1817, opened 2026-08-03
[5.x] feat: add support for customising baseline workflow name

owenvoke:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

This adds support for using another name other than `tia-baseline.yml`, for example we use `code-coverage.yml` to contain all our code coverage flows, and it would be nice to just add TIA to this, rather than needing to rename our existing workflow, or add a separate one just for this.


MrPunyapal:
Thanks! Custom baseline workflow support is already on 5.x (see BaselineSync::workflowFile() and WatchPatterns::setBaselineWorkflow()), so closing this.

PR #1818, opened 2026-08-03
[4.x] fix: resolve issue for named arguments

owenvoke:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

When running tests that have named arguments in a dataset, but only a single argument, the `$arguments[0]` throws an undefined array key exception before the test is run.

Screenshot 2026-08-03 at 11 35 05

This bug was introduced in [the 4.7.7 release](https://github.com/pestphp/pest/releases/tag/v4.7.7).


nunomaduro:
fixed.

PR #1822, opened 2026-08-03
[4.x] correct error handling for PestDatasetProviderError

maartenpaauw:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

When using named parameter, the argument array does not start by zero. Therefore, `array_key_first` must be used to check the first argument. Otherwise "Undefined array key 0" is seen as error message at test cases with datasets using named arguments.

### Related:

https://github.com/pestphp/pest/pull/1749

### DIFF
--- a/src/Factories/TestCaseMethodFactory.php
+++ b/src/Factories/TestCaseMethodFactory.php
@@ -235,8 +235,8 @@ final class TestCaseMethodFactory
             $attributesCode
                 public function $methodName(...\$arguments)
                 {
-                    if (count(\$arguments) === 1 && \$arguments[0] instanceof __PestDatasetProviderError) {
-                        throw \$arguments[0]->getPrevious() ?? \$arguments[0];
+                    if (count(\$arguments) === 1 && \$arguments[array_key_first(\$arguments)] instanceof __PestDatasetProviderError) {
+                        throw \$arguments[array_key_first(\$arguments)->getPrevious() ?? \$arguments[array_key_first(\$arguments)];
                     }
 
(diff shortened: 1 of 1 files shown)


maartenpaauw:
#1818 seems like to fix the same thing. I'm not sure if all array access calls should be replaced by `array_key_first` in my proposal.

nunomaduro:
i would feel more confortable if this is targetting 5.x + with a test on it

nunomaduro:
fixed.

src/Factories/TestCaseMethodFactory.php
```diff
@@ -235,8 +235,8 @@ public function buildForEvaluation(): string
             $attributesCode
                 public function $methodName(...\$arguments)
                 {
-                    if (count(\$arguments) === 1 && \$arguments[0] instanceof __PestDatasetProviderError) {
-                        throw \$arguments[0]->getPrevious() ?? \$arguments[0];
+                    if (count(\$arguments) === 1 && \$arguments[array_key_first(\$arguments)] instanceof __PestDatasetProviderError) {
+                        throw \$arguments[array_key_first(\$arguments)->getPrevious() ?? \$arguments[array_key_first(\$arguments)];
```
owenvoke, line 239: I'm pretty sure this introduced a new bug? It's missing a closing square bracket?

owenvoke: Ah, my bad. Didn't realise Nuno had closed this and made the change without the bracket issue.

PR #1832, opened 2026-08-07
fix: skip TIA when mutation testing consumes the coverage report

### LINKED ISSUE #1797: [Bug]: --tia + --mutate exits 1 without running mutation testing (a replayed run writes no coverage report)
### What Happened

`--mutate` combined with `--tia` exits 1 without running mutation testing at all — no mutation phase, no score, no error message. Nothing indicates why.

When the affected set is fully replayed from the Tia cache, no test process executes, so no `--coverage-php` report is written. `MutationTestRunner::run()` needs that report:

```php
if (! Coverage::isAvailable() || ! file_exists($reportPath = Coverage::getPath())) {
    Container::getInstance()->get(Printer::class)->reportError('No coverage report found, aborting mutation testing.');

    return 1;
}
```

The run ends at exit 1 and, under `--parallel`, without that message reaching the user.

### How to Reproduce

Run a `--mutate` command with `--tia` on a tree Tia has a usable cache for, so the affected set is replayed rather than executed:

```
$ ./vendor/bin/pest --tia --mutate --min=100

  ─ Experimental TIA mode enabled.

  Tests:    … (… replayed)

$ echo $?
1
```

No `Mutating application files...`, no mutant count, no score. Dropping `--tia` from the same command runs mutation normally and exits 0.

A replay is the normal, expected Tia outcome on an unchanged tree, so `--tia --mutate` fails on any run where nothing has changed.

### Expected

`--mutate` needs real test execution to produce coverage. One of:

1. `--mutate` opts out of replay (force execution of the affected set, keeping Tia's *selection* — mutating only affected files is still a win);
2. or Tia skips replay whenever a coverage report is required by another plugin;
3. or, at minimum, the abort is reported: currently the `reportError` above does not surface under `--parallel`, so the run looks like "tests passed" followed by a bare exit 1.

Worth noting for anyone hitting this: `--mutate` on Pest 5 also aborts with `Call to a member function getData() on array` once it does reach the mutation phase — pestphp/pest#1790, fixed by pestphp/pest-plugin-mutate#38. The two are independent; this report is with that patch applied, so the abort above is reached on its own.

### Environment

- pestphp/pest v5.0.2, pestphp/pest-plugin-mutate v5.0.0 (with pestphp/pest-plugin-mutate#38 applied)
- phpunit/phpunit 13.2.6, phpunit/php-code-coverage 14.2.3
- PHP 8.5.9, Xdebug coverage driver, Laravel 13

Martyrer:
### What:

- [x] Bug Fix

### Description:

`Mutate` runs before `Tia` in the plugin chain. It pops `--mutate` and appends `--coverage-php=<path>`. `Tia` reads raw argv, so it sees neither flag, never sets `piggybackCoverage`, and replays instead of running. The report is written but empty, so every mutant escapes.

```
argv: --tia --mutate     Mutate       --tia --coverage-php=...     Tia
                     ------------->                            ---------> raw argv has no --mutate
                     pops --mutate                                        piggyback OFF -> replay
                     adds --coverage-php                                  report written, EMPTY
```

`--tia --mutate` now skips TIA and runs the suite, so the report is real. Also fixes `--tia --mutate --fresh` purging the graph it then never rebuilt.

### Related:

Closes #1797

### DIFF
--- a/src/Plugins/Tia.php
+++ b/src/Plugins/Tia.php
@@ -129,6 +129,8 @@ final class Tia implements AddsOutput, HandlesArguments, HandlesOriginalArgument
         '--coverage-text', '--coverage-xml',
     ];
 
+    private const string MUTATE_OPTION = '--mutate';
+
     /** @var list<string> */
     private const array PARTIAL_SELECTION_FLAGS = [
         '--filter', '--exclude-filter', '--group', '--exclude-group',
@@ -504,6 +506,13 @@ final class Tia implements AddsOutput, HandlesArguments, HandlesOriginalArgument
 
             $this->forceRefetch = false;
             $this->filteredMode = false;
+
+            return $arguments;
+        }
+
+        if (! $isWorker && $enabled && $this->hasArgument(self::MUTATE_OPTION, $this->originalArguments)) {
+            $this->requestWorkerResults();
+            $this->emitMutationRecordSkipped();
 
             return $arguments;
         }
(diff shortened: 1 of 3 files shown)


nunomaduro:
fixed

PR #1834, opened 2026-08-08
fix(bin): resolve vendor directory dynamically to support custom Composer layouts

emyassine:
**Summary**

The `bin/pest` binary hardcodes `dirname(__DIR__, 4).'/vendor/autoload.php'` to locate the autoloader when installed as a Composer dependency. This breaks for any project that sets a custom `vendor-dir` in `composer.json`, and also silently produces a wrong `$rootPath` when Composer's bin proxy injects a path containing `..` segments that bare `dirname()` does not resolve.

**Fix**

Three-step resolution, in priority order:

1. **Prefer `$GLOBALS['_composer_autoload_path']`** — Composer itself sets this in every generated bin proxy via `BinaryInstaller`. It always points at the correct autoloader for the project invoking the binary, regardless of what the vendor directory is named. PsySH relies on the same contract.
2. **Try `dirname(__DIR__, 3).'/autoload.php'`** — since Composer always installs binaries at `{vendor-dir}/{vendor}/{package}/bin`, going up 3 levels always lands in the vendor dir, whatever its name.
3. **Walk up the tree** as a last resort for unusual install layouts.

`realpath()` is applied before `dirname()` so bin proxy paths containing `..` don't corrupt `$rootPath`. A clear stderr message is emitted if no autoloader is found, instead of a cryptic PHP fatal.

**Tested**

Verified against a project using a custom `vendor-dir` (`third_party/`). Both `--init` and running a full test suite work correctly end to end:

```
$ ./third_party/bin/pest --init
   INFO  Preparing tests directory.
  phpunit.xml ........ File already exists.
  tests/Pest.php ..... File already exists.
  ...

$ ./third_party/bin/pest --test-directory=software/onboarding/tests software/onboarding/tests
   PASS  Webkernel\Onboarding\Tests\Unit\OnboardingRuntimeTest
  ✓ complete when satisfies without store
  ✓ manual complete uses store
  ✓ optional steps do not block completion
  ✓ wizard locks future steps
  ✓ custom is onboarding done gate
  Tests:    5 passed (18 assertions)
  Duration: 0.09s
```

No regression on standard `vendor/` layouts — the `$GLOBALS['_composer_autoload_path']` path resolves correctly there too.

**Functional changes**
- No change for standard `vendor/` layouts
- Projects with a custom `vendor-dir` now work correctly
- `$rootPath` is always derived from a resolved, canonical path

### DIFF
--- a/bin/pest
+++ b/bin/pest
@@ -1,8 +1,6 @@
 #!/usr/bin/env php
 <?php
-
 declare(strict_types=1);
-
 use Pest\Contracts\Restarter;
 use Pest\Kernel;
 use Pest\Panic;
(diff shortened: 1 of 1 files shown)


emyassine:
Hey @nunomaduro 

I did a small fix for bin/pest — it breaks silently on any project using a custom vendor-dir in composer.json. The hardcoded dirname(__DIR__, 4) just doesn't hold up, and bare dirname() on an unresolved Composer bin proxy path corrupts $rootPath on top of it.

The fix uses $GLOBALS['_composer_autoload_path'] first — which is a Composer contract, PsySH does the same — then falls back gracefully. No behaviour change for standard layouts. Hope it lands! 🙌****

I did the same for filament, still pending review : https://github.com/filamentphp/filament/pull/20254 by @danharrin

nunomaduro:
there is a alot of changes on this pr that are not related.

emyassine:
@nunomaduro Composer is made in a way that the vendor directory can be renamed (for a reason or another). If this PR seems with a lot of "unrelated" changes ... okey. But is it possible po have pestphp get the vendor fir from composer just like the ComposerScripts of laravel does ? 

The only changes i made by the way are contained in one block in the bin/pest file 

I am sure you can do the same thing in a better way ! 
If I knew how to do it so that it can be accepted i would have

Thank you by advance again 

Yassine

PR #1840, opened 2026-08-10
fix: reject a wrong-cased Tia page directory on case-insensitive filesystems

marventhieme:
## The bug

`JsModuleGraph::PAGE_DIR_CANDIDATES` probes `resources/js/Pages` before `resources/js/pages`, and the probe is a bare `is_dir()`. `is_dir()` ignores casing on APFS and NTFS — and, less obviously, on a macOS bind mount seen from inside a Linux container, which is what Docker Desktop and ddev give you by default. So in a project whose real directory is the lowercase `resources/js/pages` (the Laravel 12 starter-kit layout), the capital-P candidate resolves and is accepted.

That breaks a `--tia` run in two ways.

**1. The run dies after the summary and the graph is never persisted.**

`dirHasPageFile()` iterates the wrong-cased path, which populates the container's dentry cache under `Pages`. `fingerprint()` then walks `dirname($pagesDir)` = `resources/js`, gets the real lowercase `pages` back from `readdir()`, and `opendir()` on it hits the case-aliased entry and fails:

```
Tests:    914 passed (3632 assertions)

   UnexpectedValueException
  RecursiveDirectoryIterator::__construct(/var/www/html/resources/js/pages): Failed to open directory: No such file or directory
  at vendor/pestphp/pest/src/Plugins/Tia/JsModuleGraph.php:277
```

Exit code 1, and — the part that actually costs the user — the graph never lands, so every subsequent run reports `fresh graph` and Tia never replays anything. The directory obviously exists, which sends you looking in the wrong place.

**2. `bin/pest-tia-vite-deps.mjs` silently under-selects.**

`discoverPagesDir()` runs the same capital-first probe with `existsSync`, so `pagesDir` comes back wrong-cased and every page path derives from it via `resolve(dir, entry.name)`. Deps reached through the `@` alias resolve lowercase and still match, but anything imported *relatively* from inside a page keeps the wrong casing and never joins with the paths reported for changed files. Those edges quietly disappear and Tia selects too few tests. Fixing only the crash would leave this in place, so both probes are fixed here.

`TIA_VITE_PAGES_DIR` is not a workaround for this: `firstExistingPagesDir()` never reads it.

## The fix

After `is_dir()` / `existsSync()` passes, require each segment of the candidate to appear in its parent directory with exactly the casing the candidate asks for (`scandir()` / `readdir()` are case-exact on every filesystem). The wrong-cased candidate is then skipped and the loop falls through to the real directory.

The `TIA_VITE_PAGES_DIR` override stays exempt — that casing was stated explicitly.

Behaviour on case-sensitive filesystems is unchanged: `is_dir()` was already exact there, and the new check agrees with it.

## Tests

`tests/Unit/Plugins/Tia/JsModuleGraph.php` (new) and one case added to `tests/Unit/Plugins/Tia/ViteDepsHelper.php` for the Node side. The `matchesDiskCasing` assertions are filesystem-independent, so they hold the contract on Linux CI too.

One caveat worth stating plainly: `it('resolves the pages directory with the casing it has on disk')` is the test that actually reproduces the bug, and it can only *fail* on a case-insensitive filesystem — on Linux CI it passes both before and after. On macOS, against `5.x` without this patch, it fails with:

```
-'/…/resources/js/pages'
+'/…/resources/js/Pages'
```

## Verification

- `vendor/bin/pint --test --parallel`, `rector --dry-run`, `phpstan` — clean; type coverage 100%.
- `pest --exclude-group=integration` — 1548 passed.
- `pest --group=tia` — 198 passed.
- Reproduced and fixed on a real project: a Laravel 13 + Inertia app with lowercase `resources/js/pages`, on macOS under ddev with `performance_mode: none`. Inside the container, `is_dir('…/resources/js/Pages')` is `true` while `scandir` reports `pages`, and the same check against a container-native path (`/tmp`) is `false` — so it is the mount, not PHP. Before the patch `firstExistingPagesDir()` returned `…/resources/js/Pages` and the suite exited 1 with the exception above; after it, it returns `…/resources/js/pages`.

### DIFF
--- a/bin/pest-tia-vite-deps.mjs
+++ b/bin/pest-tia-vite-deps.mjs
@@ -190,6 +190,24 @@ async function listPageFiles(pagesDir) {
   return out
 }
 
+// `existsSync()` ignores casing on APFS and NTFS, and on a macOS bind mount seen from
+// inside a Linux container, so the `resources/js/Pages` candidate resolves in a project
+// whose directory is really `resources/js/pages`. Every page path is derived from the
+// directory we accept here, so the wrong casing would silently detach the pages tree from
+// the paths the PHP side reports.
+export async function matchesDiskCasing(projectRoot, rel) {
+  let current = projectRoot
+
+  for (const segment of rel.split('/')) {
+    let entries
+    try { entries = await readdir(current) } catch { return false }
+    if (!entries.includes(segment)) return false
+    current = join(current, segment)
+  }
+
+  return true
+}
+
 async function discoverPagesDir() {
   const override = process.env.TIA_VITE_PAGES_DIR
   if (override && override.length > 0) {
@@ -199,6 +217,7 @@ async function discoverPagesDir() {
   for (const rel of PAGE_DIR_CANDIDATES) {
     const abs = resolve(PROJECT_ROOT, rel)
     if (!existsSync(abs)) continue
+    if (!(await matchesDiskCasing(PROJECT_ROOT, rel))) continue
     const files = await listPageFiles(abs)
     if (files.length > 0) return abs
   }
--- a/src/Plugins/Tia/JsModuleGraph.php
+++ b/src/Plugins/Tia/JsModuleGraph.php
@@ -85,12 +85,37 @@ final class JsModuleGraph
                 continue;
             }
 
+            if (! self::matchesDiskCasing($projectRoot, $rel)) {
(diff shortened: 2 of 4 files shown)


marventhieme:
Implemented hours later via https://github.com/pestphp/pest/releases/tag/v5.1.0.

PR #1841, opened 2026-08-10
feat(tia): add GitLab CI baseline support

calebdw:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

Hello!

This adds GitLab support to the tia engine, not just `gitlab.com` but also self-hosted instances (as long as the `glab` binary is authenticated) as many companies use the self-hosted option for their codebases.

Thanks!

### DIFF
--- a/src/Plugins/Tia/BaselineSync.php
+++ b/src/Plugins/Tia/BaselineSync.php
@@ -7,10 +7,10 @@ namespace Pest\Plugins\Tia;
 use Pest\Exceptions\BaselineFetchFailed;
 use Pest\Panic;
 use Pest\Plugins\Tia;
+use Pest\Plugins\Tia\Baselines\BaseRemote;
 use Pest\Plugins\Tia\Contracts\State;
 use Pest\Support\View;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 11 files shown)


nunomaduro:
hi friend, did you tested this on gitlab itself?

calebdw:
Not, yet---I started to but got pulled away on other fixes. I'll let you know once I confirm this works

calebdw:
@nunomaduro, I've tested this out in GitLab and it works! (even in a private self-hosted instance) You can test it yourself with this project: https://gitlab.com/calebdw/sql-entities

Note that this really depends on https://github.com/pestphp/pest/pull/1844 so that should be merged as well (I had to do some finagling in the `.gitlab-ci.yml` to get tia recording working which is definitely not something we want to have the general userbase doing

nunomaduro:
@MrPunyapal

MrPunyapal:
I'm flying let's see how early I can see it.

calebdw:
@MrPunyapal, tests should be passing now :crossed_fingers:

calebdw:
Rebased and conflicts resolved

calebdw:
Thanks! I'll put together some docs so that it can be shown on the website!

PR #1842, opened 2026-08-10
fix: fail missing snapshots in CI

calebdw:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Prevent snapshot assertions from silently creating missing snapshots during CI runs unless --update-snapshots is enabled.


nunomaduro:
thank you!

PR #1844, opened 2026-08-11
fix(tia): persist state on detached HEAD

calebdw:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

This allows TIA to persist its dependency graph and related state when Git HEAD is detached.

Detached HEAD is a normal checkout state for GitLab CI and for local workflows that use jj with a colocated Git repository. TIA previously suppressed graph writes, state deletion, structural rebuilds, and `--fresh` purges in this state. A run could therefore pass without producing the graph required by a baseline artifact or a later local run.

TIA now permits the same state operations for attached and detached checkouts. It resolves the writable baseline identity in this order:

1. Use the current Git branch when Git reports one.
2. Use the current branch reported by the GitLab or GitHub CI provider.
3. Use a stable `@workspace:<hash>` key for jj and other detached workspaces.

The default branch remains the read fallback. A local detached workspace can download a baseline recorded on the default branch, calculate changes from that baseline, and store its own SHA, tree, and test results under its workspace key. This prevents an unknown detached checkout from changing the published default-branch baseline.

The TIA CI abstraction now resolves both current and default branches, so each provider owns its environment-variable handling. Workspace baselines survive Git branch reclamation and do not participate in default-branch inference.

The tests cover sequential and parallel detached runs, structural rebuilds, `--fresh`, unreadable graph replacement, branch reclamation, default-baseline isolation, and detached GitLab and GitHub branch detection. The fixture process also removes inherited CI variables so local-behavior tests remain local when the Pest suite itself runs in CI.

Thanks!

### DIFF
--- a/src/Plugins/Snapshot.php
+++ b/src/Plugins/Snapshot.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Pest\Plugins;
 
 use Pest\Contracts\Plugins\HandlesArguments;
+use Pest\Support\Ci;
 use Pest\TestSuite;
 
 /**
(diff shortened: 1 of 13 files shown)


calebdw:
@nunomaduro, I came across this while testing the GitLab tia work---this is needed to support recording tia in gitlab ci

calebdw:
@MrPunyapal, can you please take a look at this? This is a blocker to be able to properly support Gitlab (#1841)

calebdw:
Nice, let me fix the tests

calebdw:
@MrPunyapal, can you run ci again please?

MrPunyapal:
> @MrPunyapal, can you run ci again please?

Refreshed my page multiple times already, but no luck yet.
image

MrPunyapal (approved):

calebdw:
I guess ci is down again lol

MrPunyapal (changes requested):
ci

MrPunyapal:
Can you try doing a new PR?

PR #1845, opened 2026-08-11
fix(tia): suppress PHPUnit XML coverage during recording

calebdw:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

When phpunit.xml has `<coverage>` and `<source>` sections, PHPUnit auto-initializes its code coverage driver which takes over xdebug's coverage APIs. TIA's recorder then gets empty data from `xdebug_get_code_coverage()`, resulting in zero recorded edges and no graph being saved.

### DIFF
--- a/src/Plugins/Tia.php
+++ b/src/Plugins/Tia.php
@@ -1238,6 +1238,10 @@ final class Tia implements AddsOutput, HandlesArguments, HandlesOriginalArgument
             return $arguments;
         }
 
+        if (! $this->piggybackCoverage && ! in_array('--no-coverage', $arguments, true)) {
+            $arguments[] = '--no-coverage';
+        }
+
         if (Parallel::isEnabled()) {
             $this->purgeWorkerPartials();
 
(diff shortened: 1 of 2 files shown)


calebdw:
@nunomaduro, recording tia in ci is broken for every project that has `<coverage>` its own coverage section in `phpunit.xml` unless they manually pass `--no-coverage` to the pest command which is counter intuitive

MrPunyapal (approved):

MrPunyapal (changes requested):
Two things before this can merge:

1. Rector fails on the heredoc indentation in tests/Features/Tia/CoveragePiggyback.php
2. The new test "xml-configured coverage does not prevent tia recording" fails in the integration suite, with exit code 1 instead of 0. Maybe it needs an env guard like the other TIA tests.

Once CI is green, we can merge.

calebdw:
@MrPunyapal, looks like it needs approval to run

MrPunyapal (approved):

MrPunyapal:
thanks

PR #1848, opened 2026-08-11
fix: don't treat array datasets as callables

### LINKED ISSUE #1847: [Bug]: 2 unkeyed (Model::class) datasets fail
### What Happened

When I run:

```php
use App\Models\Project;
use App\Models\User;

describe('audit logging', function () {
    it('created: event for observed model: ', function (string $modelClass) {
        /* expectations omitted, the test does not run with or without */
    })->with([
        Project::class,
        User::class,
    ]);
```

it should return: 
```sh
✓ audit logging → it created: event for observed model: with ('App\Models\Project')
✓ audit logging → it created: event for observed model: with ('App\Models\User')
```

### How to Reproduce

run a test with the dataset of only two model::class, you get the error below. 

```sh
  ⨯ audit logging → it created: event for observed model:                                                                0.41s
  ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   FAILED  Tests\Feature\AuditTest > `audit logging` → it created: event for observed model:                            Error
  Call to a member function connection() on null

  at vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php:2233
    2229▕      * @return \Illuminate\Database\Connection
    2230▕      */
    2231▕     public static function resolveConnection($connection = null)
    2232▕     {
  ➜ 2233▕         return static::$resolver->connection($connection);
    2234▕     }
    2235▕
    2236▕     /**
    2237▕      * Get the connection resolver instance.
```
If trim it to one, it succeeds, 
If you add more than to it succeeds. 
If you key any it will succeed, and;
If use arrays, viz `->with([[model::class], [model::class]])` it succeeds.

### Sample Repository

_No response_

### Pest Version

5.1

### PHP Version

8.5.9

### Operation System

macOS

### Notes

_No response_

lazerg:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

A dataset of exactly two class strings, such as `->with([Project::class, User::class])`, is also a valid PHP callable array when the first class has `__callStatic`, which every Eloquent model does. `DatasetsRepository::processDatasets()` runs its `is_callable()` check before the iterable handling, so it called `Project::User()` rather than using the two values, and the test died inside Eloquent. Arrays are now skipped by that check.

One or three values, keyed values, or nested arrays never matched the callable-array shape, which is why only this case failed.

### Related:

Fixes #1847

### DIFF
--- a/src/Repositories/DatasetsRepository.php
+++ b/src/Repositories/DatasetsRepository.php
@@ -141,7 +141,7 @@ final class DatasetsRepository
                 $datasets[$index] = self::getScopedDataset($data, $currentTestFile);
             }
 
-            if (is_callable($datasets[$index])) {
+            if (! is_array($datasets[$index]) && is_callable($datasets[$index])) {
                 $datasets[$index] = call_user_func($datasets[$index]);
             }
 
--- a/tests/.snapshots/success.txt
+++ b/tests/.snapshots/success.txt
@@ -1764,6 +1764,7 @@
   ✓ it shows the correct description for long texts with newlines
   ✓ it shows the correct description for arrays with many elements
   ✓ it shows the correct description of datasets with html
+  ✓ it does not treat a two element dataset of class names as a callable
 
    PASS  Tests\Unit\Expectations\OppositeExpectation
   ✓ it throw expectation failed exception with string argument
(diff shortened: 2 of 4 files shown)


MrPunyapal:
Already Fixed by @nunomaduro here https://github.com/pestphp/pest/commit/2c7cc1dc87ef5e5e85e7eafb411849bce3d95689

PR #1851, opened 2026-08-11
feat: add numeric expectations (toBePositive, toBeNegative, toBeZero,…

oluwatosinolamilekan:
### What:

- [x] New Feature

### Description:

Adds five numeric expectations, filling a gap alongside the existing
`toBeGreaterThan`, `toBeLessThan`, `toBeBetween`, `toBeInfinite`, and `toBeNan`:

- `toBePositive()` — value is greater than zero
- `toBeNegative()` — value is less than zero
- `toBeZero()` — value equals zero (`0`, `0.0`, `'0'`)
- `toBeOdd()` — value is an odd number
- `toBeEven()` — value is an even number

These read as plain English, matching Pest's design goal —
`expect($balance)->toBePositive()` over `expect($balance)->toBeGreaterThan(0)`.
Common uses include asserting prices/quantities are positive, a cleared
balance is zero, or batch/pagination sizes have the expected parity.

## Usage

```
expect(10)->toBePositive();
expect(-5)->toBeNegative();
expect(0)->toBeZero();
expect(3)->toBeOdd();
expect(4)->toBeEven();

```

### DIFF
--- a/src/Mixins/Expectation.php
+++ b/src/Mixins/Expectation.php
@@ -152,6 +152,68 @@ final class Expectation
     /**
      * @return self<TValue>
      */
+    public function toBePositive(string $message = ''): self
+    {
+        Assert::assertGreaterThan(0, $this->value, $message);
+
+        return $this;
+    }
+
+    /**
+     * @return self<TValue>
+     */
+    public function toBeNegative(string $message = ''): self
+    {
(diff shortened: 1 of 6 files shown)


nunomaduro:
not a fan of this.

PR #1861, opened 2026-08-12
fix(tia): fingerprint the configuration file in use

### LINKED ISSUE #1858: TIA: the fingerprint ignores the config actually in use (-c / --configuration)
`Fingerprint::compute()` hashes hardcoded paths:

```php
// src/Plugins/Tia/Fingerprint.php:36-37
'phpunit_xml'      => self::trackedHash($projectRoot, 'phpunit.xml'),
'phpunit_xml_dist' => self::trackedHash($projectRoot, 'phpunit.xml.dist'),
```

Nothing consults the `-c` / `--configuration` path in use. `configuration`, `argv`, `PHPUNIT_` and `getopt` each appear 0 times in that file.

Effect: `pest --tia` and `pest --tia -c alternate.xml` share one graph, although the two configs can declare different testsuites, different excluded groups, different bootstrap and different source roots. Switching config does not invalidate the graph.

Second gap, same helper: `trackedHash()` returns `null` for anything not tracked by git (`:255-262`, `isTrackedByGit()` at `:264`), so an untracked local `phpunit.xml` gets no fingerprint coverage at all.

Fix: hash the resolved active config path, and hash it whether or not it is tracked.

v5.1.0.

Related, same root cause — the fingerprint's input set being incomplete: #1856 (git HEAD not covered) and #1857 (`php_minor` holds the major version). Three independent gaps found in one sitting, which may argue for widening the input set rather than patching the fields.

lazerg:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

The TIA fingerprint hashed `phpunit.xml` and `phpunit.xml.dist` by name, so `-c` / `--configuration` never reached it. `pest --tia` and `pest --tia -c alternate.xml` shared one graph even though the two files can declare different testsuites, excluded groups, bootstrap and source roots, and results recorded under one config were replayed under the other.

`compute()` now takes the run's arguments and hashes whatever configuration the run actually resolves: the `-c` value when there is one (file or directory), otherwise `phpunit.xml` falling back to `phpunit.xml.dist`. The active config is hashed whether or not git tracks it, since an untracked local `phpunit.xml` still decides what the run does.

I checked this against a scaffolded project with a seeded graph. Before, `--tia -c alternate.xml` replayed all six cached results; after, it drops the graph and re-records, and plain `--tia`, `-c phpunit.xml` and `-c ./` all still replay.

Happy to add a `tests/Features/Tia` scenario for it if you want one.

### Related:

Fixes #1858.


lazerg:

PR #1867, opened 2026-08-14
fix(tia): exclude vendor from xdebug coverage while recording

### LINKED ISSUE #1865: TIA: the Xdebug recorder traces vendor/ and then discards it, where the pcov path restricts up front
## Summary

TIA's recorder starts Xdebug coverage with **no path filter**, traces every executed line including all of `vendor/`, and then discards everything out of scope in PHP. The stored graph holds zero vendor files, so all of that tracing is thrown away.

The pcov path in the same file does the opposite: it restricts collection to the in-scope list *before* materialising it.

On a real suite, excluding `vendor/` from instrumentation cut the recording pass from **1,633s to 657s** — the Xdebug tax dropped from **4.61× to 1.86×** relative to a plain run. The recorded graph was unchanged.

## The asymmetry

`src/Plugins/Tia/Recorder.php`:

```php
// pcov — restricted BEFORE collect (:136-149)
\pcov\stop();
$scope = $this->sourceScope();
$filesToCollectCoverageFor = [];
foreach (\pcov\waiting() as $file) {
    if (is_string($file) && $scope->contains($file)) {
        $filesToCollectCoverageFor[] = $file;
    }
}
$data = \pcov\collect(\pcov\inclusive, $filesToCollectCoverageFor);
```

```php
// xdebug — unfiltered, then discarded in PHP (:121, :152-160)
\xdebug_start_code_coverage();          // no filter, no arguments
...
$data = \xdebug_get_code_coverage();
\xdebug_stop_code_coverage(true);
$scope = $this->sourceScope();
foreach (array_keys($data) as $file) {
    if (! $scope->contains($file)) {     // discard vendor here, after tracing it
```

`xdebug_set_filter` does not appear anywhere in `src/`.

## Measured (pest 5.1.0, PHPUnit 13.3.0, PHP 8.4.23, Xdebug 3.4.0alpha2-dev, macOS)

Suite: 2,032 tests, 354s without coverage.

| run | wall | tax |
|---|---|---|
| `--tia` recording, unfiltered Xdebug | 1,633s | 4.61× |
| `--tia` recording, `vendor/` excluded | **657s** | **1.86×** |

The filter applied, in the project's Pest bootstrap:

```php
xdebug_set_filter(XDEBUG_FILTER_CODE_COVERAGE, XDEBUG_PATH_EXCLUDE, [__DIR__.'/../vendor/']);
```

Correctness was checked before speed, since a filter that narrows the graph would be worse than the tax:

- Graph shape unchanged — 945 files, 171 test files with edges, every top-level directory present with identical counts.
- Per-test-file edge sets identical for 167 of 168 files; the one difference was a transient that self-corrected on the next run and involved two *project* files, which the filter cannot explain.
- A filtered run of one edit went from 546s to 247s — from 1.54× slower than a plain full suite to 1.43× faster.

## Suggested fix

Call `xdebug_set_filter(XDEBUG_FILTER_CODE_COVERAGE, XDEBUG_PATH_EXCLUDE, [...])` — or the `XDEBUG_PATH_INCLUDE` equivalent derived from `sourceScope()` — before `xdebug_start_code_coverage()`, so Xdebug is asked not to instrument what the recorder is about to throw away.

An exclude list is the safer form: an include list built from a scope guess can silently narrow the graph, whereas excluding `vendor/` can only remove files `sourceScope()` already rejects.

## Note

Excluding `vendor/` is what was measured. Excluding it *and* everything else out of scope may do better; that was not tested. The 4.61× and 1.86× figures are from one machine and one suite, so treat the ratio as indicative rather than a benchmark.

lazerg:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

The recorder starts Xdebug coverage with no filter, so every executed line of `vendor/` is traced and then dropped by the scope check in `endTest()`. The pcov path in the same class already narrows the file list before it collects anything.

`Recorder::activate()` now calls `xdebug_set_filter()` with the directories `SourceScope` treats as noise (`vendor`, `node_modules`, the cache directories, `storage/framework`), so Xdebug never instruments them. Xdebug only honours the filter for files compiled after it is set, which is why the call sits in `activate()` and not next to `xdebug_start_code_coverage()` in `beginTest()`. Every prefix ends with a separator, otherwise `vendor` would also match a sibling such as `vendored`.

The scope check after `xdebug_get_code_coverage()` stays, since the filter is a prefilter and not a replacement for it. The recorded graph is unchanged: `tests/Features/Tia` passes in full with Xdebug 3.5.3 in coverage mode.

I added a unit file for `noisePaths()` and bumped both suite tallies by hand, since `composer update:snapshots` would have baked in the `Tests\Unit\Support\Backtrace` failure I get on a clean `5.x` here.

### Related:

Fixes #1865

### DIFF
--- a/src/Plugins/Tia/Recorder.php
+++ b/src/Plugins/Tia/Recorder.php
@@ -48,6 +48,14 @@ final class Recorder
     {
         $this->active = true;
         $this->captureCoverage = true;
+
+        if ($this->driverAvailable() && $this->driver === 'xdebug') {
+            \xdebug_set_filter(
+                XDEBUG_FILTER_CODE_COVERAGE,
+                XDEBUG_PATH_EXCLUDE,
+                SourceScope::noisePaths(TestSuite::getInstance()->rootPath),
+            );
+        }
     }
 
     public function activateLinkTracking(): void
--- a/src/Plugins/Tia/SourceScope.php
+++ b/src/Plugins/Tia/SourceScope.php
@@ -100,6 +100,21 @@ final class SourceScope
         return array_values(array_unique($out));
     }
 
+    /**
+     * @return list<string> Absolute directory prefixes, each ending in a separator, that hold no project source.
+     */
+    public static function noisePaths(string $projectRoot): array
+    {
+        $out = [];
+
+        foreach ([...self::TOP_LEVEL_NOISE, ...self::NESTED_NOISE] as $relative) {
+            $abs = $projectRoot.DIRECTORY_SEPARATOR.str_replace('/', DIRECTORY_SEPARATOR, $relative);
+            $out[] = self::normalise(@realpath($abs) ?: $abs).DIRECTORY_SEPARATOR;
+        }
+
+        return array_values(array_unique($out));
+    }
+
     public function contains(string $absoluteFile): bool
     {
         if (isset($this->containsCache[$absoluteFile])) {
--- a/tests/.snapshots/success.txt
+++ b/tests/.snapshots/success.txt
@@ -1937,6 +1937,10 @@
   ✓ it records assertions against the same per-dataset key
   ✓ it leaves a test without a dataset keyed by class and method
 
+   PASS  Tests\Unit\Plugins\Tia\SourceScope
+  ✓ noisePaths() → it lists the directories that hold no project source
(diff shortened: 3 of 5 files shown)


lazerg:

PR #1869, opened 2026-08-14
fix: datasets with a single named argument

### LINKED ISSUE #1868: [Bug]: Datasets with only one named parameter result in tests failing
### What Happened

Following the pest docs about named parameters in datasets, I had setup some tests with a dataset that uses an associative array with only one key. The tests were running correctly in pest version 2, 3 and 4 but are now failing to run after upgrading to pest 5. It seems that pest handles arrays with a count of 1 as an indexed array and then this line https://github.com/sonalidudhia/pest/blob/974546d49f3813456826ee6e58bc611812f9b52d/src/Factories/TestCaseMethodFactory.php#L238 tries to read array index 0 and fails.

### How to Reproduce

You can try to following simple example:
```php
it('has user data', function (array $data) {
    expect($data['name'])->toBeString();
    expect($data['email'])->toContain('@');
})->with([
    ['data' => ['name' => 'Taylor', 'email' => 'taylor@laravel.com']],
    ['data' => ['name' => 'Nuno', 'email' => 'enunomaduro@gmail.com']],
]);
```

Execution:
```
php artisan test tests/Feature/MiniTest.php                                                                                                  776ms

   FAIL  Tests\Feature\MiniTest
  ⨯ it has user data with (['Taylor', 'taylor@laravel.com'])                                                                           0.10s  
  ⨯ it has user data with (['Nuno', 'enunomaduro@gmail.com'])                                                                          0.03s  
  ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────  
   FAILED  Tests\Feature\MiniTest > it has user data with (['Taylor', 'taylor@laravel.com'])                                 ErrorException   
  Undefined array key 0

  ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────  
   FAILED  Tests\Feature\MiniTest > it has user data with (['Nuno', 'enunomaduro@gmail.com'])                                ErrorException   
  Undefined array key 0


  Tests:    2 failed (0 assertions)
  Duration: 0.19s

```

### Sample Repository

_No response_

### Pest Version

5.1.1

### PHP Version

8.4.20

### Operation System

Linux

### Notes

_No response_

lazerg:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

A dataset row written as an associative array with one key, like `['name' => 'Taylor']`, is spread into the generated test method as a named argument, so `$arguments` holds `name` and no `0`. The guard for dataset provider errors still reads `$arguments[0]` whenever the count is 1, so every row of such a test raises `Undefined array key 0` before the body runs. Rows with two or more named keys survive because the count check short-circuits first.

This is already fixed on 4.x in 3223d50, it just never made it over to 5.x. I took that commit as is, `array_key_first()` on the guard plus the test that came with it, so a later 4.x merge stays a no-op. The snapshot and the parallel tally are updated for the extra test.

### Related:

Fixes #1868

### DIFF
--- a/src/Factories/TestCaseMethodFactory.php
+++ b/src/Factories/TestCaseMethodFactory.php
@@ -187,8 +187,8 @@ final class TestCaseMethodFactory
             $attributesCode
                 public function $methodName(...\$arguments)
                 {
-                    if (count(\$arguments) === 1 && \$arguments[0] instanceof __PestDatasetProviderError) {
-                        throw \$arguments[0]->getPrevious() ?? \$arguments[0];
+                    if (count(\$arguments) === 1 && \$arguments[array_key_first(\$arguments)] instanceof __PestDatasetProviderError) {
+                        throw \$arguments[array_key_first(\$arguments)]->getPrevious() ?? \$arguments[array_key_first(\$arguments)];
                     }
 
                     return \$this->__runTest(
--- a/tests/.snapshots/success.txt
+++ b/tests/.snapshots/success.txt
@@ -275,6 +275,7 @@
   ✓ dataset items can mix named and sequential styles with ('Taylor', 'taylor@laravel.com')
   ✓ dataset items can mix named and sequential styles with ('James', 'james@laravel.com') #1
   ✓ dataset items can mix named and sequential styles with ('James', 'james@laravel.com') #2
+  ✓ named parameters work with a single argument with ('Taylor')
 
    PASS  Tests\Features\Depends
   ✓ first
@@ -2226,4 +2227,4 @@
   ✓ pass with dataset with ('my-datas-set-value')
   ✓ within describe → pass with dataset with ('my-datas-set-value')
 
-  Tests:    2 deprecated, 4 warnings, 5 incomplete, 2 notices, 40 todos, 35 skipped, 1573 passed (3414 assertions)
+  Tests:    2 deprecated, 4 warnings, 5 incomplete, 2 notices, 40 todos, 35 skipped, 1574 passed (3415 assertions)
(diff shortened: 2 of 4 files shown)


nunomaduro:
this is fixed.

frutality:
This is not in 5.1.2. Any issues with the fix?

lazerg:
You are right, it is still broken. I checked out `5.x` at HEAD, which is the same commit as `v5.1.2`, and it fails there:

```php
test('single named key', function (string $name): void {
    expect($name)->toBe('Taylor');
})->with([['name' => 'Taylor']]);
// Undefined array key 0
```

Two or more named keys still pass, because the `count($arguments) === 1` guard short-circuits.

The 4.x fix (3223d506, released in v4.7.8) never reached 5.x. `git merge-base --is-ancestor 3223d506 origin/5.x` returns false, and `array_key_first` does not appear anywhere in `src/` on 5.x. The generated guard in `src/Factories/TestCaseMethodFactory.php:190` still reads `$arguments[0]`.

I applied the same one-line change from 3223d506 on top of 5.x HEAD and the test passes.

frutality:
Ok I have to ask :) what is the way to push this to one of next releases? We can't try pest 5 because of this bug without writing a bit of workarounds in test cases

lazerg:
It is still broken. I checked the 5.x branch again today: `TestCaseMethodFactory.php` still generates `$arguments[0]`, and the repro from #1868 still fails on it.

My guess at what happened is #1769, which merged about twenty seconds before this PR was closed. That one fixes `repeat()` with named dataset arguments, which is a different code path in `Testable.php`.

GitHub would not let me reopen this PR, so I opened #1894 with the same one line fix rebased on current 5.x. Tests and types pass there. When it gets released is up to the maintainers, not me.

Until then the guard only misfires when a row has exactly one named key, so a row with two keys or a plain positional row works.

PR #1895, opened 2026-09-02
Command gh not found on Windows

sertxudev:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

When running the TIA engine on Windows, the `gh` command is not detected because `which gh` only works on Unix-based OS.

For Windows, we have to use `where gh`.

### DIFF
--- a/src/Plugins/Tia/BaselineSync.php
+++ b/src/Plugins/Tia/BaselineSync.php
@@ -583,7 +583,12 @@ final readonly class BaselineSync
 
     private function commandExists(string $cmd): bool
     {
-        $process = new Process(['which', $cmd]);
+        if (PHP_OS_FAMILY === 'Windows') {
+            $process = new Process(['where', $cmd]);
+        } else {
+            $process = new Process(['which', $cmd]);
+        }
+
         $process->run();
 
         return $process->isSuccessful();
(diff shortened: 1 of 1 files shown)


iammikek (approved):

src/Plugins/Tia/BaselineSync.php
```diff
@@ -583,7 +583,12 @@ private function classifyGhError(string $output): array
 
     private function commandExists(string $cmd): bool
     {
-        $process = new Process(['which', $cmd]);
+        if (PHP_OS_FAMILY === 'Windows') {
```
MrPunyapal: Is it possible to use (`new \Symfony\Component\Process\ExecutableFinder())->find('gh');` here? 👀

MrPunyapal (changes requested):

src/Plugins/Tia/BaselineSync.php
sertxudev: @MrPunyapal tomorrow morning I'll try it!

sertxudev:
Hi @MrPunyapal . I've updated the code so we now use the `find` method from `Symfony\Component\Process\ExecutableFinder` as proposed. This method returns the command path if it exists, or null if it doesn't. 

We're checking whether the command exists, so it makes sense to check that the output is not null.

MrPunyapal:
can you rebase this

MrPunyapal (approved):

sertxudev:
Hi @MrPunyapal, I've just rebased the PR.

FYI: You can enable "Always suggest updating pull request branches" in the GitHub repository settings:

imagen

This option shows a button on the PR timeline to update with merge or rebase, so it's a one-click action.

imagen

MrPunyapal:
Thanks

